PR #388, opened 2025-12-11
156 arch d 22 fix hardware capabilities name collision

PeterGRutherford:
>> - Changed lorawan_available bool to lorawan_hardware Option<LoRaWANHardware>
>> - Updated all usage sites to check .is_some()/.is_none() directly
>> - Unified hardware detection in discovery/hardware.rs (production version)
>> - Removed backward compatibility helper (pre-alpha code)


umwelt (changes requested):
Requesting changes from an architecture/security perspective: this refactor collapses LoRaWAN detection to a single device and drops the detailed capabilities (band/power/class) from the surfaced API, so the stack can’t handle multi-radio nodes or enforce region/power/security policy when selecting LoRaWAN. Please keep full capability data and support multiple devices.

lib-network/src/discovery/hardware.rs
```diff
@@ -5,12 +5,13 @@
 use anyhow::Result;
 use tracing::{info, debug};
 use std::collections::HashMap;
+use super::lorawan_hardware::LoRaWANHardware;
 
 /// Hardware capabilities detected on the system
 #[derive(Debug, Clone, Default)]
 pub struct HardwareCapabilities {
-    /// LoRaWAN radio hardware available
```
umwelt, line 12: Switching to a single `Option<LoRaWANHardware>` means we can only represent one radio; the previous detection recorded all found devices. Architecturally we need to support multiple LoRa radios (e.g., different bands/paths or redundant concentrators). Consider exposing `Vec<LoRaWANHardware>` (and populating `hardware_details` accordingly) so multi-radio nodes don’t lose capability info.

lib-network/src/discovery/hardware.rs
```diff
@@ -79,33 +80,35 @@ impl HardwareCapabilities {
...
-        return detect_macos_lorawan_hardware(hardware_details).await;
-    }
-    
-    #[cfg(not(any(target_os = "linux", target_os = "windows", target_os = "macos")))]
-    {
-        warn!("LoRaWAN hardware detection not implemented for this platform");
-        false
+    // Use the dedicated lorawan_hardware detection
+    match super::lorawan_hardware::detect_lorawan_hardware().await {
+        Ok(Some(hardware)) => {
```
umwelt: This detection path inserts a minimal HardwareDevice and then drops the richer fields from `LoRaWANHardware` (bands, max_tx_power, class support, payload limits). That means the rest of the stack can’t enforce region/power/data-rate policy or pick the right band, which is both a compliance and reliability risk. Can we surface the full capabilities (e.g., keep them on HardwareCapabilities, maybe as Vec<LoRaWANHardware>) and have consumers use them instead of just is_some?

PeterGRutherford: Yes I will update it, I just pushed my code that would have fixed it, but my mac did something that caused it to pull 52 different commits???? I'm fixing it but I've never had this happen to me.

PR #438, opened 2025-12-16
fix(identity): Properly load node identity with private key

umwelt:
## Summary
Fix for `create_or_load_node_identity()` silently overwriting identities.

The function was using `serde_json::from_str` which fails because `ZhtpIdentity` requires `from_serialized` with private key.

**Before**: File exists → parse fails → creates new identity → OVERWRITES
**After**: File exists → load with private key → if fails → FATAL error (no overwrite)

## Root Cause
`serde_json::from_str::<ZhtpIdentity>` doesn't work - the identity needs to be deserialized with `from_serialized()` which requires the private key to re-derive secrets.

## Test Plan
- [ ] Existing identity loads successfully
- [ ] Corrupted identity returns FATAL error (no overwrite)
- [ ] Missing keystore creates new identity

### DIFF
--- a/zhtp/src/runtime/mod.rs
+++ b/zhtp/src/runtime/mod.rs
@@ -2233,20 +2233,61 @@ pub async fn create_or_load_node_identity(
 
     let identity_file = keystore_path.join("node_identity.json");
 
-    // Try to load existing identity from keystore
-    if identity_file.exists() {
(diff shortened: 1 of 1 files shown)


umwelt:
Closing stale PR - the branch has diverged. Will create new PR with just the critical fix.

umwelt:
Superseded by #439

PR #441, opened 2025-12-17
Fix #168: Consolidate storage abstractions and resolve stack overflow

### LINKED ISSUE #168: [ARCH-D-3.5] Unify Storage Abstractions
**Phase:** 3 - P2-Medium
**Category:** Cross-Layer Integration Cleanup
**Files:** `lib-storage/src/lib.rs`, `storage.rs`

## Description
Two storage systems - `UnifiedStorageSystem` vs `DhtStorage` - pick one

## Key Tasks
- [ ] Make `UnifiedStorageSystem` canonical
- [ ] Refactor `DhtStorage` as internal component
- [ ] Export unified API
- [ ] Deprecate direct `DhtStorage` usage

## Acceptance Criteria
- Single storage API

## References
See ARCHITECTURAL_FIXES_IMPLEMENTATION_PLAN.md Ticket 3.5

PeterGRutherford:
- Unified DhtStorage as single source of truth across all packages
- Made ZkDHTIntegration methods fully async to eliminate nested block_on calls
- Updated all call sites in lib-blockchain, lib-network, and zhtp to use async/await
- Removed problematic tokio::runtime::Handle::current().block_on() pattern

Changes:
- lib-storage: DhtStorage is now the canonical DHT storage implementation
- lib-network: ZkDHTIntegration uses async initialization without block_on
- lib-blockchain: DhtBlockchainIndex updated to use async DHT operations
- zhtp: RuntimeOrchestrator properly awaits async DHT initialization


umwelt (changes requested):
Requesting changes: switching DHT integration to UnifiedStorageSystem is good, but it breaks the existing synchronous constructors and hides DhtStorage entirely. We need a clear migration path/compat layer (async init, error propagation, and explicit deprecation), plus guarded re-exports so downstreams don’t panic. Also ensure the new generic store/get operations are gated/validated (TTL ignored, no quotas) before merging.

lib-network/src/dht/mod.rs
```diff
@@ -31,13 +31,17 @@ use std::sync::Arc;
...
-    storage::DhtStorage,
     routing::KademliaRouter,
 };
 pub use lib_storage::types::dht_types::DhtNode;
 
+// Import UnifiedStorageSystem for lib-network's internal use
+use lib_storage::UnifiedStorageSystem;
+
 // Re-export our new transport layer
 pub use transport::{DhtTransport, PeerId, UdpDhtTransport, BleDhtTransport, MultiDhtTransport, PeerAddressResolver};
```
umwelt: ZkDHTIntegration::new() is now async and returns Result, but the API still exposes the same name. 

Callers compiled against the old sync ctor will now fail at build time and may not handle init errors. 

 adding a `new_sync()` wrapper that panics only in tests, or better, clearly deprecate the old signature and thread the async Result through all call sites with a migration note.

lib-storage/src/dht/storage.rs
```diff
@@ -63,8 +63,11 @@ async fn atomic_write_async(path: PathBuf, bytes: Vec<u8>) -> std::io::Result<()
 /// DHT storage manager with networking
 ///
```
umwelt, line 64: DhtStorage is now pub(crate) and not re-exported, which is fine, but we need a clear migration path for downstream crates.

 Right now they lose all direct access without a replacement. Please provide a thin compatibility layer in lib-network (or lib-storage) that exposes the necessary operations via UnifiedStorageSystem, and document the deprecation path so dependents don’t break silently

lib-storage/src/lib.rs
```diff
@@ -587,6 +590,44 @@ impl UnifiedStorageSystem {
         tracing::info!("Listed {} domain records from DHT storage", records.len());
         Ok(records)
```
umwelt, line 597: UnifiedStorageSystem now exposes generic store/get/remove/list functions that bypass any TTL/quota/ACL policy (ttl arg ignored). 

Before merging, please make sure  these are validated/gated (or clearly scoped to internal-only use) so callers don’t accidentally rely on unbounded storage writes/reads in production.

PR #495, opened 2025-12-23
Change binary name

TNTTheLagger:
Changed the binary name in the build.sh and build.ps1 from zhtp-orchestrator to zhtp since that is the name of the binary that gets built.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -15,12 +15,12 @@ cargo build --release --workspace
 if ($LASTEXITCODE -eq 0) {
     Write-Host "`n✅ Build complete!" -ForegroundColor Green
     Write-Host ""
-    Write-Host "Binary location: target\release\zhtp-orchestrator.exe" -ForegroundColor White
+    Write-Host "Binary location: target\release\zhtp.exe" -ForegroundColor White
     Write-Host ""
     Write-Host "To run a node:" -ForegroundColor Cyan
     Write-Host "  .\run-node.ps1" -ForegroundColor White
     Write-Host "  or" -ForegroundColor White
-    Write-Host "  .\target\release\zhtp-orchestrator.exe --config crates\zhtp\configs\test-node1.toml" -ForegroundColor White
+    Write-Host "  .\target\release\zhtp.exe --config crates\zhtp\configs\test-node1.toml" -ForegroundColor White
 } else {
     Write-Host "`n❌ Build failed!" -ForegroundColor Red
     exit 1


umwelt (changes requested):
Great catch on the PowerShell script! 👍 

However, to make this fully complete, we should also update the bash build script to be consistent:

**Still need updating:**
- **build.sh line 17**: Change `echo "Binary location: target/release/zhtp-orchestrator"` to `echo "Binary location: target/release/zhtp"`
- **build.sh line 22**: Change `echo "  ./target/release/zhtp-orchestrator --config..."` to `echo "  ./target/release/zhtp --config..."`
- **run-node.sh line 10**: Change `if [ ! -f "target/release/zhtp-orchestrator" ]; then` to `if [ ! -f "target/release/zhtp" ]; then`

Could you add these final fixes to keep all build scripts consistent across platforms (bash, PowerShell, and Cargo output)?

run-node.sh
```diff
@@ -13,4 +13,4 @@ if [ ! -f "target/release/zhtp-orchestrator" ]; then
 fi
```
umwelt, line 13: Line 10 above still has the old binary name in the file check:

```bash
if [ ! -f "target/release/zhtp-orchestrator" ]; then
```

Should be:

```bash
if [ ! -f "target/release/zhtp" ]; then
```

run-node.sh
```diff
@@ -13,4 +13,4 @@ if [ ! -f "target/release/zhtp-orchestrator" ]; then
...
 
 echo "▶️  Launching node..."
-./target/release/zhtp-orchestrator --config "$CONFIG_FILE"
+./target/release/zhtp --config "$CONFIG_FILE"
```
umwelt, line 16: ✅ Good fix here.

However, **line 10 above** still has the old binary name in the file check and needs to be updated:
```bash
if [ ! -f "target/release/zhtp" ]; then
```

Also, **build.sh** (lines 17 & 22) needs the same binary name updates.

umwelt (commented):
Also, please update build.sh to match these PowerShell changes

build.ps1
```diff
@@ -15,12 +15,12 @@ cargo build --release --workspace
 if ($LASTEXITCODE -eq 0) {
     Write-Host "`n✅ Build complete!" -ForegroundColor Green
     Write-Host ""
-    Write-Host "Binary location: target\release\zhtp-orchestrator.exe" -ForegroundColor White
+    Write-Host "Binary location: target\release\zhtp.exe" -ForegroundColor White
```
umwelt, line 18: ✅ Good fix here for PowerShell.

For consistency, **build.sh also needs updating** (lines 17 & 22):
```bash
echo "Binary location: target/release/zhtp"
echo "  ./target/release/zhtp --config zhtp/configs/test-node1.toml"
```

PR #641, opened 2026-01-06
Refactor: Move difficulty adjustment ownership to lib-consensus

PeterGRutherford:
- Create DifficultyManager in lib-consensus
- Move DifficultyConfig to lib-consensus
- Update BlockchainConsensusCoordinator to use DifficultyManager
- Update Blockchain to delegate difficulty adjustment to Coordinator
- Add governance parameters for difficulty adjustment
- Update documentation and architecture diagrams

### DIFF
--- a/lib-consensus/src/difficulty.rs
+++ b/lib-consensus/src/difficulty.rs
@@ -294,9 +294,13 @@ impl DifficultyManager {
         // Calculate actual timespan
         let actual_timespan = interval_end_time.saturating_sub(interval_start_time);
         if actual_timespan == 0 {
-            return Err(DifficultyError::CalculationError(
-                "Actual timespan cannot be zero".to_string(),
-            ));
+            // Edge case: identical timestamps between adjustment interval blocks
+            // Return None to indicate no adjustment needed (defensive behavior matching legacy fallback)
+            tracing::warn!(
+                "Difficulty adjustment skipped: actual_timespan is zero at height {}",
+                height
+            );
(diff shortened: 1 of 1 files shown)


umwelt:
### Code review

Found 1 issue:

1. Inconsistent error handling for zero actual_timespan between DifficultyManager and legacy fallback creates consensus risk

When the consensus coordinator path is used, `DifficultyManager::adjust_difficulty()` returns an error if `actual_timespan == 0`. However, when this error triggers the fallback to `calculate_difficulty_legacy()`, the legacy code silently returns the current difficulty unchanged for the same condition. This creates a window where different nodes could diverge on difficulty calculations:
- Node A: coordinator succeeds, returns error (actual_timespan == 0) 
- Node B: coordinator fails, fallback silently handles it → different result
- Result: potential chain fork

Both paths should handle zero timespan consistently—either both error out or both silently return current difficulty (but not mixed).

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/9350cf7350a953bf249d40878703cef8a265cad0/lib-blockchain/src/blockchain.rs#L104-L112

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/9350cf7350a953bf249d40878703cef8a265cad0/lib-consensus/src/difficulty.rs#L1184-L1192

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1099, opened 2026-02-05
fix(lib-client): Deterministic seed recovery and migration endpoint

umwelt:
## Summary

- Fix native `generate_keypair_from_seed` to use deterministic generation (was ignoring seed, generating random keys)
- Add migration endpoint for users with broken seed phrases

## Problem

Native implementation in `lib-client/src/crypto.rs` was calling `Self::generate_keypair()` instead of using the seed, making recovery impossible. Same seed produced different keys each time.

## Changes

- **lib-client/Cargo.toml**: Add `crystals-dilithium` to native deps
- **lib-client/src/crypto.rs**: Fix `generate_keypair_from_seed` to use `DilithiumKeypair::generate(Some(seed))`
- **zhtp/.../backup_recovery.rs**: Add `POST /api/v1/identity/migrate` endpoint
- **zhtp/.../mod.rs**: Wire up migration route

## Migration Endpoint

`POST /api/v1/identity/migrate`

```json
{
  "display_name": "username_to_claim",
  "new_did": "did:zhtp:...",
  "new_public_key": "hex_dilithium_pk",
  "device_id": "device-uuid"
}
```

Protections:
- Verifies display_name exists on an identity
- Blocks duplicate migrations (only one allowed per display_name)
- Blocks if new DID already registered
- Marks old identity with `migrated_to` metadata

## Test plan

- [x] `cargo check -p lib-client` passes
- [x] `cargo test -p lib-client test_dilithium5_deterministic_keygen` passes
- [x] `cargo check -p zhtp --lib` passes
- [ ] Deploy to dev and test migration flow end-to-end

### DIFF
--- a/zhtp/src/runtime/test_api_integration.rs
+++ b/zhtp/src/runtime/test_api_integration.rs
@@ -3,6 +3,17 @@ mod api_integration_tests {
     
     use crate::runtime::{RuntimeOrchestrator, Component, ApiComponent};
     use crate::config::NodeConfig;
+    use crate::api::handlers::identity::IdentityHandler;
+    use crate::api::middleware::{RateLimiter, CsrfProtection};
+    use crate::session_manager::SessionManager;
+    use crate::api::handlers::identity::login_handlers::AccountLockout;
+    use lib_identity::{IdentityManager, RecoveryPhraseManager, economics::EconomicModel as IdentityEconomicModel};
+    use lib_protocols::types::{ZhtpHeaders, ZhtpMethod, ZhtpRequest, ZHTP_VERSION, ZhtpStatus};
+    use lib_protocols::zhtp::ZhtpRequestHandler;
+    use lib_storage::{UnifiedStorageConfig, UnifiedStorageSystem};
+    use std::sync::Arc;
+    use tokio::sync::RwLock;
+    use base64::Engine as _;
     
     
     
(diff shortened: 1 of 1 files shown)


umwelt:
## Answers

### New seed source

**Currently:** `generate_identity()` uses `random_bytes(32)` for master_seed, but then generates keys with `Dilithium5::generate_keypair()` (random, NOT derived from seed).

**For migration flow**, app should:
1. Generate new random 32-byte seed via `random_bytes(32)`
2. Call `restore_identity_from_seed(new_seed, device_id)` to derive keys deterministically
3. Convert seed to mnemonic with `mnemonic_from_entropy()` to show user

This ensures the new seed phrase will work for future recovery.

### Transport for `/api/v1/identity/migrate`

**Problem:** Users can't authenticate with UHP because their old private keys don't match their seed phrase (that's the whole bug).

**Options:**

1. **Public ALPN (no UHP)** - Anyone could claim any display_name. Need additional verification:
   - Rate limiting per IP
   - Email/phone verification
   - Admin approval queue

2. **Authenticated ALPN (UHP required)** - User signs with old private key (if they still have it in app storage). More secure but requires app to have old keys.

**Recommendation:** If users still have the app with old private keys stored locally, use authenticated. If they've lost everything except seed phrase, need public with additional verification.

Which scenario are we targeting?

PR #1791, opened 2026-03-06
fix(consensus): bootstrap validator key seeding and consensus round improvements

umwelt:
## Summary
- Fix remote validators receiving 32-byte blake3 placeholder keys instead of real Dilithium2 keys from TOML config, causing vote rejection with "non-verifiable consensus key"
- Wire consensus step timeouts (propose/prevote/precommit) from TOML config
- Relax bootstrap TOFU signature verification (security maintained by validate_committed_block)
- Increase MAX_MESSAGE_SIZE from 16MB to 64MB for large block sync batches
- Reduce block page size from 200 to 50 for more reliable sync
- Reorder validator registration before registry seeding to prevent idempotency skip
- Fix partition detection threshold to reduce false positives

## Test plan
- [ ] All 4 nodes deploy without crash
- [ ] No "non-verifiable registered consensus key" warnings in logs
- [ ] Blocks commit at height 1179+
- [ ] G4 reconnects to full validator set

### DIFF
--- a/lib-protocols/src/wire/mod.rs
+++ b/lib-protocols/src/wire/mod.rs
@@ -32,8 +32,11 @@ use crate::types::{ZhtpRequest, ZhtpResponse, ZhtpStatus};
 /// Wire protocol version
 pub const WIRE_VERSION: u16 = 1;
 
-/// Maximum message size (64 MB — needed for large block sync batches)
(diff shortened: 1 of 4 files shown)


umwelt:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/1791#pullrequestreview-3903004526)

PR #1936, opened 2026-03-19
fix(storage): supplementary sled writes + V8 bonding-curve-registry persistence

umwelt:
## Summary

- **No-active-block-transaction halt** (hit twice in production at heights 3474 and 3604): `finish_block_processing` was called with no active sled transaction when the executor path was active. Added `begin/commit/rollback_supplementary_writes` to open an identity/wallet batch after the executor has already committed the block — without re-validating block height or touching `LATEST_HEIGHT`.

- **Oracle `cbe_usd_price: None` after every restart**: `bonding_curve_registry` was never persisted, so `get_cbe_curve_price_atomic()` always returned `None` after load. Bumped storage to V8 which serializes the registry. V7 files migrate transparently via a post-load backfill that re-registers the CBE genesis curve entry.

- **Late-finalization dropped**: commit quorum for round K arriving while engine was at round K+1 (same height) was silently ignored. `process_committed_block` now takes an explicit `round` parameter so the quorum check uses the correct round.

- Validate non-empty signatures on system transactions to prevent malformed sigs bypassing mempool intake.
- Expand block limits: payload 1MB→4MB, witness 512KB→2MB.

## Test plan

- [ ] Restart all 4 nodes and confirm `cbe_usd_price: Some(...)` in oracle finalization logs
- [ ] Submit an identity registration transaction and confirm no `No active block transaction` error at the next block commit
- [ ] Confirm `📂 Loaded blockchain storage v8` on restart after first save

### DIFF
--- a/zhtp/src/api/handlers/bonding_curve/api_v1.rs
+++ b/zhtp/src/api/handlers/bonding_curve/api_v1.rs
@@ -648,7 +648,10 @@ impl BondingCurveApiHandler {
 
     /// Get current supply band for CBE
     fn get_current_band(&self, supply: u128) -> u32 {
-        u32::try_from(band_for_supply(supply).index + 1).unwrap_or(u32::MAX)
+        band_for_supply(supply)
+            .ok()
+            .and_then(|b| u32::try_from(b.index + 1).ok())
+            .unwrap_or(u32::MAX)
     }
 
     /// Get requester public key from authenticated request


umwelt:
### Code review

Found 3 issues:

1. **Block size increase not matched by QUIC transport read cap** — `lib-blockchain/src/resources.rs` raises `max_payload_bytes` to 4 MB, but `lib-network/src/protocols/quic_mesh.rs` has three hard-coded `stream.read_to_end(1024 * 1024)` calls (1 MB cap). Any block using the new payload limit will fail to be received over QUIC, causing silent block-propagation failures and consensus stalls.

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/b03a050c360aa1e1ff54c1801e2e30ac1d958b0c/lib-network/src/protocols/quic_mesh.rs#L1003-L1006

2. **`commit_supplementary_writes` silently drops `token_contracts` writes** — `begin_supplementary_writes` allocates a full `PendingBatch` (including `token_contracts`), but `commit_supplementary_writes` only applies four trees (`identities`, `identity_metadata`, `identity_by_owner`, `accounts`). `process_wallet_transactions`, called within the supplementary write window (see `blockchain.rs:3194`), calls `put_token_contract` (see `blockchain.rs:5889`), which writes to `batch.token_contracts`. Those writes are silently dropped on every executor-path block.

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/b03a050c360aa1e1ff54c1801e2e30ac1d958b0c/lib-blockchain/src/storage/sled_store.rs#L1170-L1206

3. **`tx_to_info` casts `u128` amount to `u64` without bounds check** — `TokenTransferData.amount` and `TokenMintData.amount` are `u128`, but `tx_to_info` casts both with `d.amount as u64`. Any token amount exceeding `u64::MAX` (~1.84×10¹⁹) will silently truncate in API responses. CBE total supply (10¹⁹ atoms) fits just under this limit, but there is no compile-time guard and no saturating cast, making this a latent hazard for future token types.

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/b03a050c360aa1e1ff54c1801e2e30ac1d958b0c/zhtp/src/api/handlers/blockchain/mod.rs#L100-L108

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #2243, opened 2026-04-21
EPIC-ZDNS: Enable ZDNS and bind to public interface


umwelt:
Parent: #2242

## Task

1. Remove the `storage-integration` feature gate in `lib-network/src/zdns/mod.rs` (line 32-45)
2. Configure ZDNS to bind to `0.0.0.0:53` instead of localhost
3. Add ZDNS startup to the validator runtime (start alongside QUIC server)
4. Add config option: `[zdns] enabled = true, bind = "0.0.0.0:53"`
5. Verify UDP port 53 is open on all 3 validators

## Files
- `lib-network/src/zdns/mod.rs` — remove feature gate
- `lib-network/src/zdns/config.rs` — default bind to 0.0.0.0
- `zhtp/src/unified_server.rs` or `runtime/mod.rs` — start ZDNS server
- `zhtp/src/config/aggregation.rs` — parse `[zdns]` config section

## Acceptance
- `dig @77.42.37.161 test.sov` returns an A record

umwelt:
Implemented in PR #2253 (consolidated ZDNS branch).

Issue #2242, opened 2026-04-21
EPIC: Self-Describing Network Directory via ZDNS


umwelt:
## Vision

Replace manual DNS configuration and the gateway daemon with a self-describing network directory built on the existing ZDNS implementation. Every validator serves DNS. New nodes appear automatically. Dead nodes disappear. No manual IP management.

## Architecture

```
App (hardcoded bootstrap: 3 validator IPs)
  → queries network.sov via ZDNS (UDP 53)
  → gets ranked validator IP list from on-chain registry
  → connects directly to best validator via QUIC
  → authenticates via UHP v2
```

**No proxy. No daemon. No gateway. No manual DNS.**

## How it works

1. Every validator runs ZDNS (already implemented, just disabled)
2. ZDNS resolves `network.sov` → returns all validator IPs from on-chain registry
3. New validator registers on-chain → automatically appears in DNS responses
4. Validator goes offline → health checks remove it from responses
5. App ships with 2-3 hardcoded bootstrap IPs for initial ZDNS discovery
6. After first connection, app discovers full network via `directory.sov`

## Existing code

- **ZDNS server**: `lib-network/src/zdns/transport.rs` — 936 LOC, fully implemented, RFC 1035, UDP+TCP, rate limiting
- **ZDNS resolver**: `lib-network/src/zdns/resolver.rs` — 537 LOC, LRU cache, TTL
- **DNS packet parsing**: `lib-network/src/zdns/packet.rs` — 425 LOC, RFC 1035 compliant
- **Validator registry**: on-chain, has endpoint addresses per validator
- **Feature gate**: disabled behind `storage-integration` flag in `zdns/mod.rs`

## Sub-issues

1. #NEXT+1 — Enable ZDNS and bind to public interface
2. #NEXT+2 — Dynamic validator endpoint resolution from on-chain registry
3. #NEXT+3 — Health-aware DNS responses
4. #NEXT+4 — Directory bootstrap domain (`directory.sov`)
5. #NEXT+5 — Network directory API endpoint (`/api/v1/network/directory`)
6. #NEXT+6 — App bootstrap integration spec

## What this replaces

- Gateway daemon (`zhtp-daemon`) — no longer needed for client routing
- Manual DNS A-record management — validators self-register
- `gateway.thesovereignnetwork.org` — replaced by `network.sov`
- ForwardedClientContext identity issues — clients connect directly, no proxy

## Dependencies

[...]

PR #2249, opened 2026-04-21
EPIC-ZDNS: Enable ZDNS transport server (#2243)

umwelt:
Parent epic: #2242

## Summary

Enables the ZDNS DNS server that was fully implemented but disabled behind a feature gate. Validators can now serve DNS queries on port 53 for `.sov` and `.zhtp` domains.

## Changes

- Remove `storage-integration` compile_error gate from `lib-network/src/zdns/mod.rs`
- Export `zdns` module from `lib-network/src/lib.rs`
- Add `[zdns]` config section with `enabled`, `bind`, `port` fields
- Wire config into `ProtocolsComponent` at runtime startup
- Auto-detect node's public IP for DNS responses via `local_ip_address`

## Config

```toml
[zdns]
enabled = true
# bind = "0.0.0.0"  # default
# port = 53          # default
```

## Test plan

- [x] Build clean (lib-network + zhtp)
- [ ] Deploy with `[zdns] enabled = true` → verify `dig @<ip> test.sov` returns A record
- [ ] Verify port 53 is accessible from external host

### DIFF
--- a/lib-network/src/lib.rs
+++ b/lib-network/src/lib.rs
@@ -308,6 +308,7 @@ pub mod validator_discovery_transport; // Mesh-based validator discovery gossip 
     feature = "full"
 ))]
 pub mod web4;
+pub mod zdns;
 #[cfg(any(
     feature = "quic",
     feature = "mdns",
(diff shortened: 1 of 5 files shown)


umwelt:
Consolidated into #2253 (feature/2245-health-aware-dns) which contains all ZDNS changes with review fixes applied.

PR #2266, opened 2026-04-21
Neural mesh compression

Seth-Ramsay:
Added a self-improving Neural Mesh compression model with a custom compression codec and "semantic channeling" data aggregation that enables encrypted shard tagging for zero knowledge content identifying.


lib-network/src/protocols/shard_protocol.rs
```diff
@@ -0,0 +1,604 @@
...
+        peer_id: NodeId,
+    ) -> Result<ShardProtocolMessage, String> {
+        match message {
+            ShardProtocolMessage::StoreRequest {
+                version,
+                shard_id,
+                data,
+                ttl_seconds,
+                requester,
+            } => self.handle_store_request(version, shard_id, data, ttl_seconds, requester).await,
```
umwelt, line 259: Blocking: `handle_message()` receives authenticated `peer_id` but this path discards it and trusts `requester` from the message body instead. That lets a peer spoof origin metadata and any higher-layer accounting keyed off requester. Please derive requester identity from the transport peer and treat the embedded field as untrusted input.

lib-network/src/protocols/shard_protocol.rs
```diff
@@ -0,0 +1,604 @@
...
+        
+        let now: u64 = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        let expires_at = now + ttl_seconds;
+        
+        // Store shard
+        let mut storage = self.storage.write().await;
+        match storage.store(shard_id, data.clone(), expires_at) {
```
umwelt, line 327: Blocking: this stores arbitrary `data` under caller-supplied `shard_id` without checking `shard_id == blake3(data)`. That breaks the content-addressed invariant and allows shard poisoning. Reject `StoreRequest` unless the claimed ID matches the bytes.

lib-network/src/protocols/shard_protocol.rs
```diff
@@ -0,0 +1,604 @@
...
+        let size = data.len();
+        
+        // Check capacity
+        if self.current_usage + size > self.max_capacity {
+            return Err("Storage capacity exceeded".into());
+        }
+        
+        // Store shard
+        self.shards.insert(shard_id, (data, expires_at));
+        self.current_usage += size;
```
umwelt, line 185: High severity: overwrite accounting is wrong here. Replacing an existing shard key always increments `current_usage`, so repeated overwrite of the same ID can exhaust reported capacity without equivalent storage growth. Capacity checks need to use the net size delta.

lib-network/src/protocols/shard_protocol.rs
```diff
@@ -0,0 +1,604 @@
...
+        
+        Ok(())
+    }
+    
+    fn fetch(&self, shard_id: &ShardId) -> Option<Bytes> {
```
umwelt, line 190: High severity: `fetch()` serves the shard bytes without checking expiry. Expired shards remain readable until cleanup happens to run, which makes TTL enforcement best-effort instead of authoritative. Enforce expiry at read time.

lib-network/src/protocols/shard_protocol.rs
```diff
@@ -0,0 +1,604 @@
...
+                expires_at: 0,
+                error: Some(format!("Shard too large: {} bytes", data.len())),
+            });
+        }
+        
+        let now: u64 = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        let expires_at = now + ttl_seconds;
```
umwelt, line 323: High severity: `ttl_seconds` is accepted unbounded and added directly to `now`. This allows arbitrarily long retention requests and can create timestamp overflow edge cases. Bound TTL to a protocol max and use checked or saturating arithmetic.

lib-neural-mesh/src/distributed.rs
```diff
@@ -0,0 +1,1723 @@
...
+            let drain_to = history.len() - 100;
+            history.drain(..drain_to);
+        }
+
+        model
+    }
+
+    /// Receive a compressed model from a peer node.
+    /// Returns true if we now have enough contributions to run FedAvg.
+    pub async fn receive_peer_model(&self, compressed: CompressedModel, sample_count: u64) -> bool {
```
umwelt, line 764: Blocking: this accepts `compressed.source_node`, `sample_count`, and `compressed.generation` straight from the payload and keys contributions on those values. A malicious peer can spoof origin, inflate its FedAvg weight, or replay a stale update into a new round. These fields need to be bound to authenticated transport state and validated against the current generation.

zhtp/src/runtime/components/neural_mesh.rs
```diff
@@ -0,0 +1,2185 @@
...
+                            Ok(()) => info!("Model imported successfully"),
+                            Err(e) => warn!("Model import failed: {}", e),
+                        }
+                    }
+                    // ── Distributed Training Operations ──
+                    "receive_peer_model" => {
+                        // Receive compressed model from a peer for FedAvg
+                        if let Ok(msg) = ModelSyncMessage::from_bytes(&data) {
+                            match msg {
+                                ModelSyncMessage::BroadcastModel { model, sample_count } => {
```
umwelt, line 1663: Blocking: this path forwards `BroadcastModel { model, sample_count }` directly into FedAvg intake without binding it to the authenticated peer/session that delivered the message. Please authenticate logical model origin at this boundary instead of trusting payload metadata.

lib-neural-mesh/src/distributed.rs
```diff
@@ -0,0 +1,1723 @@
...
+
+        // Check for quantization header and dequantize if present
+        let is_quantized = decompressed.len() >= 12 && decompressed[..4] == QUANT_MAGIC;
+        let raw = if is_quantized {
+            dequantize_weights_int8(&decompressed)?
+        } else {
+            decompressed
+        };
+
+        // Verify integrity — but allow quantization error for quantized models.
```
umwelt, line 423: High severity: the default quantized path does not verify `weight_hash`; it only checks that the dequantized output size is close enough. That means an authenticated peer can submit arbitrary quantized weights and still pass integrity checks. Add a commitment over the quantized representation itself and verify it here.

lib-compression/src/pattern_dict.rs
```diff
@@ -0,0 +1,567 @@
...
+        let raw = match data {
+            Some(d) => d,
+            None => {
+                tracing::debug!("No pattern index found in DHT");
+                return Ok(0);
+            }
+        };
+
+        // Deserialize the pattern map published by other nodes
+        let remote_patterns: HashMap<PatternId, Pattern> = bincode::deserialize(&raw)
```
umwelt, line 372: Blocking: this deserializes a remote pattern map from the DHT and merges it into the local cache with no publisher authentication and no proof validation, even though the type/docs claim proof-bound integrity. As written, a Sybil set can poison the dictionary with junk patterns. Either verify signed/proven pattern admissions here or remove the integrity claims from the protocol.

lib-compression/src/chunker.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    pub fn chunk_with_dedup(&self, data: &[u8]) -> Result<ChunkResult> {
+        let shards = self.chunk(data)?;
+        
+        use std::sync::{Arc, Mutex};
+        let unique_shards = Arc::new(Mutex::new(Vec::new()));
+        let shard_indices = Arc::new(Mutex::new(Vec::new()));
+        let seen_ids = Arc::new(DashMap::new());
+        
+        shards.par_iter().for_each(|shard| {
```
umwelt, line 169: High severity: this is a non-atomic check-then-insert on shared dedup state. Two threads can both miss the same `shard.id`, assign different indices, and both push a supposedly unique shard. This needs atomic uniqueness assignment or sequential dedup to keep `unique_shards` and `shard_indices` consistent.

lib-neural-mesh/src/distributed.rs
```diff
@@ -0,0 +1,1723 @@
...
+        Self {
+            compressor,
+            ..Self::new(node_id)
+        }
+    }
+
+    /// Configure differential privacy for federated averaging.
+    ///
+    /// When enabled, each contributor's weight delta is L2-clipped to `max_grad_norm`
+    /// and calibrated Gaussian noise is added, guaranteeing (ε, δ)-DP per round.
```
umwelt, line 692: High severity: the doc comment here overstates the privacy property. This code adds DP noise to the merged output, but the coordinator still decrypts and sees every contributor update before aggregation. That is not contributor privacy against the aggregator; it is only output privacy. Please narrow the claim or redesign around secure aggregation.

lib-neural-mesh/src/ml/isolation_forest.rs
```diff
@@ -0,0 +1,404 @@
...
+    
+    /// Average path length of unsuccessful search in BST
+    fn average_path_length(n: usize) -> f32 {
+        if n <= 1 {
+            return 0.0;
+        }
+        
+        // H(n-1) - (n-1)/n where H is harmonic number
+        let n_f = n as f32;
+        let h_n = 2.0 * (n_f - 1.0 + std::f32::consts::E.ln()).ln() - 2.0 * (n_f - 1.0) / n_f;
```
umwelt, line 73: Medium-high: this is not the standard Isolation Forest `c(n)` approximation. The current expression computes the wrong average path length baseline, which distorts anomaly scores and any tuned threshold built on top of them. Please replace it with the canonical `2 * H(n-1) - 2*(n-1)/n` approximation.

lib-neural-mesh/src/ml/isolation_forest.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// Train forest on samples
+    pub fn fit(&mut self, samples: &[Vec<f32>]) {
+        if samples.is_empty() {
+            return;
+        }
+        
+        self.n_samples_trained = samples.len();
+        self.trees.clear();
+        
+        let mut rng = rand::thread_rng();
```
umwelt, line 178: Medium-high: `random_seed` is exposed in the config but training still uses `thread_rng()` here, and tree construction does the same above. That makes the model non-reproducible across runs even when determinism is requested. Thread a seeded RNG through all random operations when `random_seed` is set.

lib-neural-mesh/src/anomaly.rs
```diff
@@ -0,0 +1,486 @@
...
+            .map_err(|e| NeuralMeshError::InferenceFailed(format!("Serialize anomaly model: {}", e)))
+    }
+
+    /// Load anomaly detection model from bytes (from compressed distributed sync)
+    pub fn load_model(&mut self, data: &[u8]) -> Result<()> {
+        let state: AnomalyDetectorState = bincode::deserialize(data)
+            .map_err(|e| NeuralMeshError::InferenceFailed(format!("Deserialize anomaly model: {}", e)))?;
+        let forest = IsolationForest::load(&state.forest_bytes)
+            .map_err(|e| NeuralMeshError::InferenceFailed(e))?;
+        let mut detector = AnomalyDetector::new(IsolationForestConfig::default(), 0.1);
```
umwelt, line 310: Medium-high: `save_model()` serializes `threshold`, but `load_model()` recreates a detector with defaults and never restores the saved threshold. That means anomaly behavior changes after export/import or restart. Please hydrate the saved threshold as part of model load and add a save/load equivalence test.

lib-compression/src/witness.rs
```diff
@@ -0,0 +1,715 @@
...
+            proof.proof_size(),
+        );
+
+        Ok(proof)
+    }
+
+    /// Verify the compression proof cryptographically.
+    ///
+    /// Checks Bulletproofs validity, data commitment, and timestamp.
+    fn verify_compression_proof(proof: &CompressionProof, root_hash: &[u8; 32], merkle_root: &[u8; 32], file_size: u64, shard_count: u64) -> Result<bool> {
```
umwelt, line 380: Medium-high: this verification is real, but the proof semantics are narrower than the surrounding feature claims. It verifies witness metadata integrity (size range, shard-count range, commitment binding), not data possession or shard availability. If this witness is meant to support storage/economic claims, the protocol claim needs to be narrowed or the proof strengthened.

lib-neural-mesh/src/compressor.rs
```diff
@@ -0,0 +1,597 @@
...
+            embedding.push(*byte as f32 / 255.0);
+        }
+        
+        // 2-4. Extract all features in parallel
+        let (stat_features, (ngram_features, struct_features)) = rayon::join(
+            || self.extract_statistical_features(data, 128),
+            || rayon::join(
+                || self.extract_ngram_features(data, 128),
+                || {
+                    let remaining = self.embedding_dim.saturating_sub(256 + 128 + 128);
```
umwelt, line 238: Medium severity: the statistical path only pushed 32 BLAKE3 byte features above, but this budget subtracts 256 dimensions as if the hash had already filled 256 slots. In the default 512-D case that wastes 224 dimensions as zeros and degrades similarity quality. This should subtract `hash_bytes.len()`, not `256`.

lib-neural-mesh/src/inference.rs
```diff
@@ -0,0 +1,195 @@
...
+            if input.len() != expected {
+                return Err(NeuralMeshError::InferenceFailed(format!(
+                    "Input size mismatch: model expects {} elements, got {}",
+                    expected,
+                    input.len(),
+                )));
+            }
+        }
+
+        // Build a 1-D tensor from the input slice
```
umwelt, line 120: Medium severity: this always constructs a rank-1 tensor even though the loader only validated total element count. Many ONNX models expect [1, N] or higher-rank inputs, so this will fail or mis-run valid models. Preserve the model's concrete input shape when building the tensor.

lib-compression/src/assembler.rs
```diff
@@ -0,0 +1,299 @@
...
+        
+        info!("Assembly complete: {}/{} shards", received_count, total_shards);
+        
+        Ok(())
+    }
+
+    /// Process a single shard
+    async fn process_shard(&self, shard: Shard) -> Result<()> {
+        // Find shard index in witness
+        let shard_index = self.witness.shard_ids
```
umwelt, line 108: Medium severity: each incoming shard does a linear scan over `witness.shard_ids`, so assembly cost grows quadratically with manifest size. Precompute a `HashMap<ShardId, usize>` once during initialization and use O(1) lookup here.

umwelt (changes requested):
Blocking findings were left inline on the affected hunks. The main issues are trust-boundary failures in shard transport and federated model sync, missing integrity enforcement on the quantized model path, and unauthenticated remote pattern import. I would not merge until those are fixed and covered by adversarial tests.

Issue #2222, opened 2026-04-20
[SYNC-WIRE] Add structured sync decode errors, metrics, and diagnostics


umwelt:
Source spec: `docs/specs/block-sync-wire-compat-v1.md`

## Summary
Standardize sync decode failure visibility to make incompatibilities operationally diagnosable.

## Scope
- Define structured error set:
  - UnsupportedPeerWireVersion
  - LegacyTxVariantUnsupported
  - BlockPageEnvelopeMalformed
  - PayloadDecodeFailed
- Log required context: peer, endpoint, selected wire version, range, payload prefix/hash
- Add counters/metrics by peer and wire-version

## Acceptance criteria
- All observer/bootstrap decode failures emit structured error class + context
- Metrics show per-peer version usage and failure counts
- Existing opaque enum-index panic loops replaced by actionable diagnostics

umwelt:
In progress: starting implementation on branch feature/2222-sync-decode-errors-metrics (stacked on #2227).

umwelt:
Implementation complete and PR opened.

PR: https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2228

Stacked base: #2227

Issue #2218, opened 2026-04-20
observer sync: block gap-fill fails with deserialization variant-index mismatch


umwelt:
## Summary
Observer/full nodes on macOS connect to bootstrap peers and start successfully, but observer gap-fill cannot ingest remote blocks due to a serialization/schema mismatch.

## Error
During sync:
- 

This repeats for multiple bootstrap peers and prevents observer catch-up.

## Reproduction
1. Build and run node in observer mode:
   - 
   - 2026-04-20T10:29:13.800419Z  INFO zhtp::config: Loading configuration from 11 packages...
2026-04-20T10:29:13.800616Z  INFO zhtp::config::environment: Loading configuration for Testnet environment
2026-04-20T10:29:13.800665Z  WARN zhtp::config::environment: ZHTP_TESTNET_KEY environment variable not set
2026-04-20T10:29:13.801051Z  INFO zhtp::config::aggregation: Loading package configurations from directory: /Users/supertramp/.zhtp/config.toml
2026-04-20T10:29:13.803603Z  INFO zhtp::config::aggregation: Loaded partial configuration file (merging with defaults)
2026-04-20T10:29:13.803627Z  INFO zhtp::config::aggregation: Loaded 3 bootstrap peer(s) from [network_config] section
2026-04-20T10:29:13.803744Z  INFO zhtp::config::aggregation: Loaded validator_enabled = false from [consensus_config] section
2026-04-20T10:29:13.804013Z  INFO zhtp::config::aggregation: ✓ Deriving NodeRole: validator_enabled=false, hosted_storage_gb=100 → Observer (full blockchain, no mining/consensus)
2026-04-20T10:29:13.804035Z  INFO zhtp::config::aggregation: ✓ Deriving NodeType: default → FullNode (complete blockchain, read-only)
2026-04-20T10:29:13.804253Z  INFO zhtp::config::validation: Validating configuration across 9 packages...
2026-04-20T10:29:13.804528Z  INFO zhtp::config::validation: Using unified server mode - mesh and API protocols share port 9333
2026-04-20T10:29:13.804572Z  INFO zhtp::config::validation: Port assignments validated - no conflicts detected
2026-04-20T10:29:13.804654Z  INFO zhtp::config::validation: Resource allocations validated
2026-04-20T10:29:13.804683Z  INFO zhtp::config::validation: Security consistency validated
[...]

PR #2268, opened 2026-04-21
Fix observer sync: skip prev-hash validation for catch-up

umwelt:
## Problem

Local observer nodes couldn't sync past block 31134 — "Invalid previous block hash" error. The binary computes a different block hash than what the validators produced (due to quorum_root computation changes, attestation ordering fixes accumulated mid-chain).

## Fix

Both catch-up sync paths now use `apply_block_trusted_for_sync_no_prev_hash` which uses the trusted replay executor (`skip_prev_hash_validation=true`). The blocks come from configured bootstrap peers, not random nodes.

## Files

- `lib-blockchain/src/blockchain.rs` — new `apply_block_trusted_for_sync_no_prev_hash` method
- `zhtp/src/runtime/components/blockchain.rs` — observer_sync_loop uses trusted path
- `zhtp/src/runtime/components/consensus.rs` — consensus catch-up uses trusted path

## Test

Local observer synced full chain (0 → 51000+) with zero errors.

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -898,6 +898,26 @@ impl Blockchain {
         }
     }
 
+    /// Apply a block from a trusted peer, skipping prev-hash validation.
+    /// Used by observer catch-up when the binary version differs from the one
+    /// that produced the blocks (hash computation may have changed).
+    pub async fn apply_block_trusted_for_sync_no_prev_hash(&mut self, block: Block) -> Result<()> {
+        if let Some(ref exec_arc) = self.executor {
+            use crate::execution::executor::BlockExecutor;
+            let trusted_exec = std::sync::Arc::new(BlockExecutor::new_trusted_replay(
+                std::sync::Arc::clone(exec_arc.store()),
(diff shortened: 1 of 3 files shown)


umwelt:
## Closing — bypasses hash chain validation

This PR skips prev-hash validation during observer and validator catch-up sync. While the intent is valid (different binary versions compute different hashes due to quorum_root/attestation ordering changes), the fix weakens chain integrity guarantees:

- A malicious bootstrap peer could insert blocks with altered transaction sets
- Observers and validators would accept reordered/modified blocks without detection
- The hash chain is the fundamental ordering integrity guarantee — bypassing it is not acceptable

**Root cause:** The real problem is that hash computation is not deterministic across binary versions. This is a serialization compatibility issue that should be solved at the wire format level.

**Correct fix:** The SYNC-WIRE epic (#2218-#2222) addresses this properly by introducing a versioned wire format with structured envelope encoding. That work should be completed instead of this workaround.

**Note:** `validate_stateless()` with the `PHASE2_ALLOWED_TYPES` check is currently dead code (never called in production), so the related tx_validate changes in #2271 are harmless.

Issue #2263, opened 2026-04-21
EPIC: Dynamic peer discovery via ZDNS — nodes visible in network.sov


umwelt:
## Problem

`network.sov` only returns validator IPs (from static bootstrap_peers config). Non-validator nodes (Mac nodes, relay nodes, edge nodes) are invisible to DNS. As the network grows, clients should connect to the nearest healthy node, not always to a validator.

## Goal

Any authenticated node that joins the network and can serve clients should automatically appear in `network.sov` DNS responses. No manual config changes.

## Architecture

```
Node connects to validator → UHP handshake
  → Validator records: DID, socket IP, capabilities
  → Peer registry entry created
  → ZDNS reads peer registry
  → network.sov returns node's IP alongside validators
  → Clients discover and connect to nearest node
```

### 1. Capture peer external IP during handshake
The QUIC handler knows the peer's socket address (`peer_addr: SocketAddr`). Store it as a `PeerEndpoint` in the peer registry entry when `auto_register_peer_identity` runs.

**Files:** `zhtp/src/server/quic_handler.rs` (auto_register_peer_identity), `lib-network/src/peer_registry/`

### 2. Node capability advertisement during handshake
Nodes should announce what they can do during the UHP v2 handshake: node_type (FullNode/Relay/Edge), supports_web4, available bandwidth, whether they accept client connections. The validator records these in the peer registry.

**Files:** `lib-network/src/handshake/`, `lib-network/src/peer_registry/mod.rs` (NodeCapabilities)

### 3. Wire peer registry into ZDNS endpoint provider
The endpoint provider currently reads only the blockchain validator_registry + bootstrap_peers config. Add the peer registry as a third source. Filter by:
- `authenticated == true`
- `node_type` is FullNode or Relay (not Edge, not Mobile)
- `last_seen` within last 60 seconds
- Has at least one QUIC endpoint

**Files:** `zhtp/src/runtime/components/protocols.rs` (endpoint provider closure)

### 4. Capability-based DNS filtering
Add a new domain: `relay.sov` → returns only relay nodes. `web4.sov` → returns only nodes that serve Web4 content. `network.sov` → returns all healthy client-serving nodes.

**Files:** `lib-network/src/zdns/transport.rs`

### 5. Peer registry persistence
The peer registry is currently in-memory only — lost on restart. For ZDNS to return node IPs immediately after validator restart (before peers reconnect), persist the peer registry to sled.

[...]

PR #2269, opened 2026-04-21
EPIC #2263: Peer discovery for ZDNS + directory

umwelt:
Closes sub-issues 1, 2, 3 of #2263

## Summary

Non-validator nodes (like the USA node) now appear in the network directory when they connect to validators. Their IPs are stored during UHP handshake and served via ZDNS and the directory API.

## Changes

- `zhtp/src/runtime/peer_endpoints.rs` — new module: DID → SocketAddr store, 5-minute auto-eviction
- `zhtp/src/server/quic_handler.rs` — stores peer IP on handshake via `register_peer_endpoint`
- `zhtp/src/runtime/components/protocols.rs` — ZDNS provider includes peer IPs
- `zhtp/src/api/handlers/network/mod.rs` — directory API returns `nodes` array
- `lib-network/src/types/mesh_message.rs` — `PeerEndpointAnnounce` gossip message type + global callback
- `lib-network/src/messaging/message_handler.rs` — handles incoming gossip

## Current state

- Validators that directly receive a UHP handshake from a peer include it in their ZDNS/directory
- The ZDNS gateway only sees peers that connect to it (not to validators)
- Gossip broadcast (so all nodes learn about all peers) is defined but not wired to the mesh transport yet

## Tested

- USA node (148.113.140.176) connects to g1 → g1's directory includes it

### DIFF
--- a/lib-network/src/messaging/message_handler.rs
+++ b/lib-network/src/messaging/message_handler.rs
@@ -881,6 +881,16 @@ impl MeshMessageHandler {
                 )
                 .await?;
             }
+            ZhtpMeshMessage::PeerEndpointAnnounce { did, addr, timestamp } => {
+                tracing::info!(
+                    "📡 Received peer endpoint gossip: {} @ {} (ts={})",
+                    &did[..20.min(did.len())], addr, timestamp
+                );
+                // Invoke the global callback if registered by the runtime.
+                if let Some(cb) = crate::types::mesh_message::PEER_ENDPOINT_CALLBACK.get() {
+                    cb(&did, &addr);
+                }
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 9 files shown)


umwelt:
## Closing — peer gossip has no authentication

The `PeerEndpointAnnounce` mesh message allows any authenticated mesh peer to announce **any DID at any IP address**. No verification that the announcer owns the DID they're announcing for.

### Attack vectors:
1. **Endpoint poisoning** — announce a real validator's DID at an attacker IP → ZDNS returns attacker for legitimate queries
2. **Memory exhaustion** — flood the peer registry with thousands of fake DIDs
3. **Eclipse attack** — announce enough fake endpoints → new nodes connecting via ZDNS get routed exclusively to attacker nodes

### Minimum fix required before re-opening:
- The announced DID **must match** the sender's authenticated DID from the UHP handshake
- Or: the announce must include a signature over `(did, addr, timestamp)` verified against the DID's on-chain public key
- Rate limiting on announces per peer
- Max registry size cap

The peer discovery concept (#2263) is valid and needed — but the gossip protocol must authenticate endpoint ownership before it can ship.

Issue #2341, opened 2026-04-25
Consensus Rewrite v2 — CONS-305 Migrate handlers to transition()


umwelt:
**Source spec:** `docs/epics/consensus-rewrite-epic.md` § CONS-305
**Blocked by:** CONS-302, CONS-303, CONS-304

## Implementation
- Rewrite the engine handlers in (formerly) `lib-consensus/src/engines/consensus_engine/state_machine.rs`:
  - `on_proposal`, `on_prevote`, `on_precommit`, `on_commit_vote`, `on_round_timeout` become thin event-translators that call `transition()` and dispatch returned `Action`s.
  - `enter_propose_step`, `enter_prevote_step`, `enter_precommit_step`, `enter_commit_step` are **deleted**. Their work moves into `Action` handlers in the runtime layer.
- Every state mutation goes through a single `self.enter(new_state)` method that updates `state` and `entered_at`.

## Acceptance criteria
- [ ] No `on_*_step` or `enter_*_step` methods contain decision logic — they only translate events to `Event` enum values or do not exist.
- [ ] Every state mutation goes through `enter()`.
- [ ] All baseline tests (CONS-003) still pass.

## Order of execution
Phase 3. Largest single rewrite. Blocks CONS-306, 307.

umwelt:
Reopened — closed in error by PRs #2396/#2397/#2398/#2401/#2402.

The CONS-305 spec is a single issue covering all 5 handler migrations + cleanup; sub-tasks (305a..f) were internal staging not separate issues. CONS-306/307/308/309 are distinct future work.

Actual scope mapping:
- CONS-305 (#2341) — handler-by-handler migration in progress; closes after CONS-305f (cleanup PR delete enter_*_step + legacy step field).
- CONS-306 (#2342, #2343?) — action channel for broadcaster — not started.
- CONS-307 — action channel for finalization — not started.
- CONS-308 — typed RoundRejected event — not started.
- CONS-309 — watchdog task — not started.

umwelt:
## Status — handler-by-handler shadow merged, cleanup deferred

Stages a..e merged into development (PRs #2396, #2397, #2398, #2401, #2402). All 5 engine handlers now route events through `transition()` and update `fsm_state` via `enter_fsm_state` while the existing `enter_*_step` helpers continue to do the per-phase work. Tests preserve KI-01..04 baseline.

## What's left for full CONS-305 closure (305f cleanup)

[...]

PR #2403, opened 2026-04-27
refactor(consensus): CONS-305f staging — dispatch_action wires enter_*_step (#2341)

umwelt:
## Summary
Staging step 1 of CONS-305f cleanup. Three structural changes that move the engine closer to "FSM is the single source of state-machine truth":

1. **Break the recursion source**. \`maybe_finalize\` no longer called from inside \`enter_commit_step\` — that internal call created a \`dispatch_action(SendCommit) → enter_commit_step → maybe_finalize → transition() → dispatch_action(CommitBlock)\` cycle that rustc rejects (E0733). The one production caller that needed the post-cast-commit-vote quorum check (\`on_round_timeout\`'s PreCommit branch) was updated to call \`maybe_finalize\` explicitly.

2. **dispatch_action wires action handlers**: \`CreateProposal → enter_propose_step\`, \`SendPrevote → enter_prevote_step\`, \`SendPrecommit → enter_precommit_step\`, \`SendCommit → enter_commit_step\`. Errors logged at warn (best-effort, CE-ENG-4). \`BroadcastProposal\`, \`CommitBlock\`, \`AdvanceRound\` are engine-side no-ops with documented reasoning.

3. **Remove redundant direct calls** at FSM-driven dispatch sites in \`on_proposal\`, \`on_prevote\`, \`on_precommit\`. The FSM's action stream now drives the work; the duplicate direct calls are gone.

## What's still left in CONS-305f

Tracked on #2341. The remaining work needs more invasive restructuring:

- **\`on_round_timeout\` walk-through match block** still calls \`enter_*_step\` directly. The FSM's walk-through transitions emit only \`ResetWatchdog\` (no Send* with target). Closing this requires either emitting Send* with a locally-tracked target on walk-through, or driving phase-entry hooks from \`enter_fsm_state\` on state-kind change.
- **Delete \`enter_*_step\` methods** — final step once walk-through is cleaned up.
- **\`current_round.step = X\` direct writes** in ~6 production sites — replace with \`enter_fsm_state(...)\`.

The CONS-305 issue (#2341) stays open until those land.

## Test plan
- [x] \`cargo test -p lib-consensus --lib\` — 183 pass / 4 fail (KI-01..04 baseline preserved)
- [x] \`cargo build --workspace\` — clean

No behavior change: \`enter_*_step\` helpers are idempotent (re-entry guards), so dispatch via FSM vs direct produces the same effect. Callers that previously triggered \`maybe_finalize\` indirectly through \`enter_commit_step\` now do so explicitly at the same call site — no finality detection regression.

### DIFF
--- a/lib-consensus/src/engines/consensus_engine/state_machine.rs
+++ b/lib-consensus/src/engines/consensus_engine/state_machine.rs
@@ -1540,9 +1540,14 @@ impl ConsensusEngine {
             self.dispatch_action(action).await;
         }
 
+        // CONS-305f: the per-step branch is reduced to the late-
+        // proposal case below. The Propose-step path used to call
+        // `enter_prevote_step` here; that work is now dispatched by
+        // the FSM `SendPrevote` action above (see the loop over
+        // `actions` returned by `transition()`).
         match self.current_round.step {
             ConsensusStep::Propose => {
-                self.enter_prevote_step().await?;
+                // Handled by FSM action dispatch above.
             }
             ConsensusStep::PreVote => {
                 // Proposal arrived after the local prevote timer already fired.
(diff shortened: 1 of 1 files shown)


umwelt:
## Late-quorum bug fixes — pushed as 5e18a522

Both correctness comments addressed:

| Site | Fix |
|------|-----|
| `on_prevote` (line 1735) | Restored `enter_precommit_step()` call. Late prevote quorum after PreVote-step timeout: `prior_fsm = Precommitting`, FSM logs and ignores. Without explicit helper call, late quorum never casts a precommit. |
| `on_precommit` (line 1869) | Restored `enter_commit_step()` call. Same shape: late precommit quorum after PreCommit-step timeout — `prior_fsm = Committed`, FSM ignores. |
| `dispatch_action` doc dup | Removed the duplicate paragraph. |
| `maybe_finalize` doc | The CONS-305f comments inside `enter_commit_step` already explain the move (lines 2552–2569 in the current file). |

The on-TIME quorum path stays clean — `enter_*_step` is idempotent so the second direct call is a no-op when the FSM already dispatched the same work.

Tests: 183 pass / KI-01..04 baseline preserved.

Step 2 (#2404) was rebased on this fix and force-pushed.

Issue #2444, opened 2026-05-01
[DECOUPLE] Phase 1E: DAO registry — cache instead of O(n) replay per request


umwelt:
## Parent Epic: #2439

`DaoHandler::rebuild_dao_registry()` iterates every block and every transaction to rebuild the DAO registry on every request. O(blocks × txs) under read lock. Gets worse as chain grows.

**Fix:** Build and cache the DAO registry on block commit events. Invalidate on new blocks. Serve from cache.

**Key file:** `zhtp/src/api/handlers/dao/mod.rs:602`

umwelt:
Resolved in PR #2450

Issue #2443, opened 2026-05-01
[DECOUPLE] Phase 1D: POUW mint — use transaction path instead of direct mint


umwelt:
## Parent Epic: #2439

`reward_minter.rs:37` calls `bc.mint_sov_for_pouw()` which bypasses the transaction + consensus path entirely. Direct state mutation.

**Fix:** Submit a `TokenMint` transaction signed by the kernel authority. Let the executor process it.

**Key file:** `zhtp/src/rewards/reward_minter.rs`

umwelt:
Resolved in PR #2450

Issue #2442, opened 2026-05-01
[DECOUPLE] Phase 1C: validator_ip — submit transaction instead of direct mutation


umwelt:
## Parent Epic: #2439

`validator_ip.rs:49-54` mutates `blockchain.validator_registry` directly via `blockchain.write()` without creating any transaction. In-memory state diverges from on-chain state.

**Fix:** Submit a `ValidatorUpdate` transaction with the new IP. Let the executor apply it canonically.

**Key file:** `zhtp/src/runtime/validator_ip.rs`

umwelt:
Resolved in PR #2450

Issue #2441, opened 2026-05-01
[DECOUPLE] Phase 1B: Event listener — async I/O outside blockchain write lock


umwelt:
## Parent Epic: #2439

`IdentityProjectionListener::on_event()` runs inside `add_block()` write lock. Does sled writes and DHT I/O while blockchain is locked. Every block commit triggers cross-system I/O under the lock.

**Fix:** Buffer events in `add_block()`, process async after lock is released. Use an mpsc channel from publisher to async listener task.

**Key files:**
- `lib-blockchain/src/events/mod.rs` — sync publisher
- `zhtp/src/runtime/blockchain_provider.rs:121` — IdentityProjectionListener

umwelt:
Resolved in PR #2450

PR #2450, opened 2026-05-01
refactor: decouple blockchain write lock — Phase 1A/1B/1E

umwelt:
## Summary

Addresses epic #2439 (blockchain/services decoupling), Phase 1 — eliminate write lock abuse.

- **Phase 1B (#2441)**: Replace synchronous `BlockchainEventPublisher` with `tokio::sync::broadcast` channel. `publish()` is non-blocking — identity/wallet projection I/O (sled + DHT) now runs in a spawned task, fully outside the blockchain write lock
- **Phase 1A (#2440)**: Split DAO proposal, vote, and delegate execution handlers — read state under read lock, drop it, build/sign tx, then write lock only for `add_pending_transaction()`. Write lock hold time drops from seconds to microseconds
- **Phase 1E (#2444)**: Cache DAO registry with height-keyed invalidation. Avoids O(blocks × txs) full replay on every registry query

## Remaining Phase 1 items
- #2442 (1C): validator_ip — needs architectural decision (local DNS overlay vs transaction)
- #2443 (1D): POUW mint — needs TokenMint transaction type

## Test plan
- [ ] Build succeeds with `cargo build --profile dev-release -p zhtp`
- [ ] Deploy to testnet and verify block production continues
- [ ] Verify identity registration events still trigger projection updates
- [ ] Verify DAO registry queries return correct data
- [ ] Verify DAO proposal/vote submission works

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -1306,9 +1306,7 @@ impl Blockchain {
             timestamp: block.header.timestamp,
             transaction_count: block.transactions.len() as u64,
         };
-        if let Err(e) = self.event_publisher.publish(event).await {
-            warn!("Failed to publish BlockAdded event: {}", e);
-        }
+        self.event_publisher.publish(event);
 
         Ok(())
     }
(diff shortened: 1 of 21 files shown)


jfranklinphillips70-art:
👍

Issue #1934, opened 2026-03-19
Prevent accidental inclusion of local artifacts and sensitive file patterns


umwelt:
## Context

A recent cleanup identified that local development artifacts (e.g. symlinks and temporary files) can be unintentionally included in the repository. While no sensitive data was exposed, we should enforce safeguards to prevent this class of issue entirely.

## Objective

Ensure that local, temporary, and potentially sensitive files are never committed to the repository.

### 1. Update `.gitignore`

Add and enforce ignore rules for:

```
tmp/
*.b64
.zhtp/
keystore*
*.key
*.pem
.env
```

### 2. Pre-commit hook (required)

Implement a pre-commit hook that blocks commits containing:

* symlinks
* files inside `tmp/`
* files matching:

  * `*.b64`
  * `keystore*`
  * `.zhtp/*`

Example checks:

* detect symlinks in staged files
* pattern match against restricted paths and extensions

### 3. Repository scan (one-time)

Run a full scan of the repository (including history if needed) to confirm:

* no keystore references
* no encoded artifacts (`*.b64`)
* no local environment folders

### 4. CI safeguard (recommended)

Add a CI step that fails builds if restricted patterns are detected in commits or PRs.

### 5. Developer guidelines

Document and communicate:

* no local paths or environment-specific artifacts in commits
* no encoded files unless explicitly required and reviewed
* use environment variables or secure storage for any sensitive material

## Acceptance Criteria

* `.gitignore` updated and enforced
* pre-commit hook active and tested
* CI check in place (if implemented)
* repository verified clean
* team informed of new guidelines

## Notes

This is a preventive measure to enforce repository hygiene and avoid non-functional artifacts entering version control.

PR #2490, opened 2026-05-05
feat: C FFI exports for messaging (iOS + Android mobile bindings)

umwelt:
## Summary
- Adds 21 `zhtp_msg_*` C FFI exports on top of the messaging module landed in #2488, so mobile apps can call the post-quantum messaging stack without a UniFFI scaffolding step.
- Sessions cross the boundary as opaque `MessagingSessionHandle*` pointers (RAII via `zhtp_msg_session_free`); envelopes flow as bincode-serialised byte buffers so callers stay agnostic of the internal `MessageEnvelope` layout.
- Stable u8 content-type tags (`Text=0 … GroupInvite=7`) so callers map back to their own enum without parsing Rust enum-name strings.

## Surface added

**Session lifecycle** — `zhtp_msg_session_initiate`, `accept`, `rekey`, `accept_rekey`, `free`
**Field accessors** — `chain_key`, `counter`, `epoch`, `local_did`, `remote_did`
**At-rest storage** — `serialize`, `deserialize` (bincode)
**Sealing** — `seal_text`, `seal_binary`, `seal_key_exchange`
**Open / sign / verify** — `envelope_open`, `envelope_sign`, `envelope_verify`
**Wire format / inspection** — `envelope_to_hex`, `envelope_from_hex`, `envelope_to_json`

## Conventions
- Returns `*mut MessagingSessionHandle` → caller frees with `zhtp_msg_session_free`.
- Returns `ByteBuffer` → caller frees with the existing `zhtp_client_buffer_free`.
- Returns `*mut c_char` → caller frees with the existing `zhtp_client_string_free`.
- Out-pointer arguments (e.g. `kyber_ct_out` in `initiate`/`rekey`) are written only on success; left untouched on error.
- Errors collapse to `null` / empty `ByteBuffer` / `-1` — matches the identity FFI.

## Out of scope
- No JSON / UDL / UniFFI bindings — this is the manual C FFI path used by iOS `@_silgen_name` declarations and Android JNI wrappers. UniFFI scaffolding can layer on top later.
- No new dependencies. `bincode`, `serde_json`, and `hex` were already in the crate.

## Test plan
- [x] Rebuilds cleanly with `cargo build --release --target aarch64-apple-ios --package lib-client --lib`
- [x] Symbols verified present in the resulting static lib (22 `zhtp_msg_*` text references in the main cgu)
- [x] Mobile app's quic-jni wrapper compiles against the new surface for arm64-v8a, armeabi-v7a, x86_64
- [ ] End-to-end round-trip test (pending follow-up: an integrated `seal_and_sign` helper that takes an `IdentityHandle*` so the Dilithium secret key never crosses the FFI on the send path)

## Notes for reviewers
- Pure additive. Existing exports untouched; no shared types relocated.
- `verify_envelope` returns `i32` (`1` valid, `0` invalid, `-1` parse error) so callers can distinguish protocol-level rejection from malformed input — both should be treated as "drop the envelope" but logging benefits from the distinction.
- The `_api_surface_tracker` style is not used here — an unused-import warning would flag missing exposure, but every imported helper is referenced.

### DIFF
--- a/lib-client/src/lib.rs
+++ b/lib-client/src/lib.rs
@@ -2095,3 +2095,525 @@ pub extern "C" fn zhtp_client_build_remove_liquidity(
         Err(_) => std::ptr::null_mut(),
     }
 }
+
+// =============================================================================
+// Post-quantum messaging C FFI
+// =============================================================================
+//
+// Mirrors the existing IdentityHandle + ByteBuffer pattern. Sessions are
+// passed across the boundary as opaque pointers; envelopes flow as
+// bincode-serialized byte buffers (the same wire shape produced by
+// `messaging::encode_envelope` minus the hex layer).
+//
+// Conventions:
+//   - Functions returning `*mut MessagingSessionHandle` hand ownership to
+//     the caller; release with `zhtp_msg_session_free`.
+//   - Functions returning `ByteBuffer` allocate; release with
+//     `zhtp_client_buffer_free`.
+//   - Functions returning `*mut c_char` allocate a NUL-terminated string;
+//     release with `zhtp_client_string_free`.
+//   - Out-pointer parameters (e.g. `kyber_ct_out`) must be non-null and
+//     are written to only on success. On error the function returns null
+//     / a sentinel and leaves the out-pointer untouched.
+//   - All `*const c_char` inputs must be valid UTF-8 NUL-terminated strings.
+
+use crate::messaging as msg_mod;
+
+/// Opaque handle to a `MessagingSession`.
+pub struct MessagingSessionHandle {
+    inner: msg_mod::MessagingSession,
+}
+
+/// Map a `ContentType` enum to a stable u8 tag for FFI callers. Order
+/// matches the Rust enum declaration; do not reorder.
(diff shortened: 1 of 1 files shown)


umwelt:
The **Block sensitive/local artifacts** failure is pre-existing on `development`, not caused by this PR. The scanner does a full `git ls-files` sweep, not a diff-scan, and flags:

```
x blocked extension: deploy/ansible/.env.example
```

That file was committed by #2384 (commit `9aeef8f7`) and pre-dates this branch. This PR's diff is 522 insertions in a single file — `lib-client/src/lib.rs` — and adds nothing matching the blocked patterns.

```
$ git diff development --stat
 lib-client/src/lib.rs | 522 ++++++++++++++++++++++++++++++++++++++++++++++++++
 1 file changed, 522 insertions(+)
```

The workflow itself references issue #1934 for this; happy to file a separate cleanup PR if there's appetite for either renaming `.env.example` → `env.example.template` or amending the regex to allow `.env.example` (the de-facto convention for committed env templates).

PR #2496, opened 2026-05-06
fix: return display_name + username in recovery response

umwelt:
Recovery response now includes display_name (from chain) and username (from credentials). Stops overwriting existing names with 'Recovered Identity'.


jfranklinphillips70-art (commented):
thank you 👍🤛

jfranklinphillips70-art (commented):
thank you👍🤛

PR #2506, opened 2026-05-06
feat: POST /identity/update-kyber-key — publish encryption key for messaging

umwelt:
Publish or rotate Kyber1024 public key on-chain. Required for PQ messaging KEM. Called after registration or recovery.


jfranklinphillips70-art (commented):
👍🤛

PR #2522, opened 2026-05-08
[BLOCKER] fix: pre-existing defects preventing node startup and demo run

scasino983:
## ⛔ Blocker PR — node will not start without these fixes

These four pre-existing defects in the `development` branch prevent any local node from starting and prevent the auth-1877 demo from running. Without this PR, fresh clones of the repo cannot run the chain at all.

## Blockers resolved

| # | Blocker | File | Effect without fix |
|---|---|---|---|
| 1 | Windows main thread stack overflow on startup | `zhtp/src/main.rs` | Node crashes immediately on startup (Windows) |
| 2 | Genesis never registers a validator on a fresh chain | `zhtp/src/runtime/components/consensus.rs` | Consensus halts at height 0, chain never advances |
| 3 | Auth dispatch does not strip query strings before route match | `zhtp/src/api/handlers/mobile_auth/mod.rs` | `/events?session_id=X` returns 404, breaking auth flow |
| 4 | Demo client uses HTTP/reqwest against a QUIC-only server | `zhtp/examples/mobile_auth_demo.rs` | Demo cannot connect at all; original client called nonexistent HTTP endpoints |

## Test plan

- [ ] `cargo build --release -p zhtp` completes without errors
- [ ] Node starts and binds UDP:9334
- [ ] `mobile_auth_demo.exe` runs all 7 steps and prints `Demo complete — all 7 steps passed`
- [ ] `demos/auth-1877/run.ps1` (Windows) and `run.sh` (Linux) run end-to-end from a clean clone


umwelt:
@scasino983  Thanks for putting this together  the diagnosis of the four startup failure modes is accurate. We won't merge this PR as-is, but three of the four fixes are correct and we're landing them in a clean PR.

## What we're keeping (extracting into a separate PR)

- **Windows 8MB main-thread stack** (`zhtp/src/main.rs`)  real issue, the deeply-nested startup futures exceed Windows' 1MB default. Harmless on Linux. Keep.
- **Auth dispatcher strips `?` before route match** (`zhtp/src/api/handlers/mobile_auth/mod.rs`) — legitimate routing bug. `/events?session_id=X` returning 404 is a real defect. Keep.
- **Demo client uses QUIC instead of HTTP/reqwest** (`zhtp/examples/mobile_auth_demo.rs`) — example file only, no production impact. Keep.

## What we're rejecting

**Auto-registering the local node as a validator in dev mode** (`zhtp/src/runtime/components/consensus.rs`).

This is the wrong shape for local development. Local nodes must not start as validators. The fix:

- Synthesizes networking/rewards keys (lines: `b"dev-net-key:"`, `b"dev-rewards-key:"`).
- Bypasses `Blockchain::register_validator`'s validation (the comment acknowledges this).
- Inserts directly into `validator_registry` to satisfy the quorum check.

What that gives us is a single-node sandbox consensus, which is not what local dev should be. Local dev should run as an **observer** against the existing testnet:

- Bootstrap peers are already listed in `tools/node-configs/shared.toml` (the 7 IPs of g1–g5 + the two gateways).
- Observer mode connects, authenticates via UHP, syncs blocks. No validator registry pollution. No synthetic keys.
- This is the same configuration already running on `zhtp-gateway` and `zhtp-gateway-2` in production.

The "no validators in canonical state" error you're hitting is the validator startup path. The fix is to not take that path locally  load the node in observer mode by default for fresh clones, not to fake a validator entry.

## Plan

We'll open a new PR with the three accepted fixes plus a default observer-mode local config. Once that lands, fresh-clone nodes will boot cleanly against testnet without any sandbox-consensus hack.

umwelt (changes requested):
Compilation is broken, in the comments I added a proposal on what to deliver

PR #2591, opened 2026-05-19
fix: Tendermint-style BFT round synchronization

umwelt:
## Summary

Fixes the consensus liveness bug behind the **height-59117 stall** (and the earlier 52140-class stalls): the chain livelocks whenever one validator falls behind, because the four remaining validators cannot stay round-synchronized.

### Root cause

Consensus rounds advanced **only on independent per-node local timers** (`prevote_timeout = 3s`). The single cross-node resync was a forward-jump on receiving a higher-round *proposal*. Four validators free-running on 3s timers drift across rounds; prevotes for the same proposal get cast in different rounds and never sum to quorum (votes are counted per `(height, round)`).

With all 5 validators healthy the system tolerated one drifting node (quorum 4-of-5). Once a validator fell behind, the remaining four had **zero drift tolerance** and livelocked — `InsufficientPrevotes` every round, forever.

### The fix — round progress is network-evidence-driven

- **`enter_round()`** — canonical round-transition path. Preserves Tendermint lock state across jumps, never moves backward or across heights, re-evaluates buffered evidence for the destination round.
- **Jump triggers** — a valid higher-round proposal (A), or `f+1` distinct validators observed prevoting (B) / precommitting (C) in a higher round. `f+1` is below quorum, so a jump never finalizes a block.
- **Lock/unlock rules** — `valid_proposal`/`valid_round` set on prevote quorum; `locked_proposal`/`locked_round` set when the node precommits. A locked validator prevotes a conflicting value only when the proposal's `valid_round >= lock_round`, else abstains. Proposers re-propose their `valid_value` (re-using its id) so locked peers recognise it.
- **Timers are liveness-only** — prevote/precommit timeout advances the round via `enter_round`; stale timers are neutralized by the existing `(height,round,step)` `TimerToken` generation check.
- **Catch-up gating** — a proposer whose blockchain is behind the consensus height abstains rather than broadcasting a stale-tip block peers reject as a fork.

`ConsensusProposal` gains a signed `valid_round` field; `CONSENSUS_PROTOCOL_VERSION` bumped 2→3 (payload tag `ZHTP/PROPOSAL/SIG/v2`). **Deploy must be atomic** (halt all 5 → deploy all → restart all) — no mixed-version window.

## Test plan

- [x] `cargo build --workspace --locked` — clean
- [x] `lib-consensus` lib tests — 140/140 pass
- [x] `lib-consensus-core` tests — 43/43 pass
- [x] `./scripts/validate-dao-ready-gates.sh` — all 17 gates pass
- [x] 7 new deterministic round-sync tests: drift convergence, proposal/prevote/precommit jumps, no-backward-jump, lock safety, timer-generation safety, lagging-proposer gating
- [ ] Atomic halt-deploy-restart on g1–g5, verify height 59117 commits and the chain advances

## Notes

- Repairs one pre-existing broken test (`test_gap4_message_relevance_invariant`) — broken since the round-acceptance window widened to 200, unrelated to round sync.
- 3 pre-existing failures in `bft_safety_partition_tests` (Byzantine fault detection / slashing) are **not touched** — they fail identically on `development`, separate subsystem.
- **Out of scope / separate follow-up:** g4's catch-up sync triggers but never delivers the missing block. This PR makes the 4 healthy validators commit reliably regardless; repairing catch-up sync delivery is bug #2.

### DIFF
--- a/lib-consensus-core/src/ports/finalization.rs
+++ b/lib-consensus-core/src/ports/finalization.rs
@@ -128,6 +128,7 @@ mod tests {
             timestamp: 0,
             signature: PostQuantumSignature::default(),
             consensus_proof: ConsensusProof::empty(ConsensusType::ByzantineFaultTolerance, 0),
+            valid_round: None,
         }
     }
 
--- a/lib-consensus-core/src/types/proposal.rs
+++ b/lib-consensus-core/src/types/proposal.rs
@@ -125,4 +125,14 @@ pub struct ConsensusProposal {
     pub signature: PostQuantumSignature,
     /// Combined proof bundle for this proposal.
     pub consensus_proof: ConsensusProof,
+    /// Tendermint `validRound`: when the proposer is re-proposing a
+    /// block it observed reach a prevote quorum in an earlier round,
+    /// this carries that round. `None` for a freshly built block.
+    /// Receivers use it in the prevote lock rule — a locked validator
+    /// may prevote a conflicting block only when `valid_round >=
+    /// locked_round`. Appended at struct end (never mid-struct) so
+    /// bincode stays compatible; `#[serde(default)]` decodes pre-field
+    /// proposals as `None`.
+    #[serde(default)]
+    pub valid_round: Option<u32>,
 }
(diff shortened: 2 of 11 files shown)


umwelt:
Follow-up bug #2 (catch-up sync never applies downloaded blocks — the g4 stuck-behind defect) is now tracked separately in #2592.

PR #2613, opened 2026-05-20
Refactor observer command parsing tests

thcsomebody:
fix(zhtp-cli): extract parse_observer_action helper to eliminate test duplication

Problem:
The feature/observer-T3 branch is failing SonarQube Quality Gate with a 4.0% duplication metric on new code (threshold: ≤ 3.0%). The duplication is in zhtp-cli/src/argument_parsing.rs lines 2249–2359 within the test module, where subcommands repeat identical ZhtpCli::try_parse_from calls and large structural match boilerplate for each variant assertion.

Solution:
Consolidate lines 2249–2359 by extracting a centralized parse_observer_action helper utility function inside the mod tests block:

    fn parse_observer_action<I, T>(args: I) -> ObserverAction
    where
        I: IntoIterator<Item = T>,
        T: IntoValue,
    { ... }


umwelt (changes requested):
This PR cannot be merged in its current form. You pasted
  diff line numbers directly into zhtp-cli/src/
  argument_parsing.rs in the observer test block. The
  source now contains literal numeric prefixes like 2249,
  duplicated on successive lines, plus stray inserted
  numbers mid-block. That is not a minor cleanup miss, it
  means the patch was not even sanity-checked before
  submission.

  There is also a real test regression:
  ObserverAction::QrRender still requires --sponsor, but
  the rewritten test no longer passes --sponsor and no
  longer asserts the sponsor field. So beyond the broken
  source, the refactor also weakens coverage of a required
  observer admission argument.

  Please cleanly reconstruct the test refactor from actual
  source, restore the qr-render test so it matches the
  real CLI contract, and rerun the parser tests before re-
  requesting review.

PR #2588, opened 2026-05-18
Feature/observer t3

PeterGRutherford:
## Summary

Adds `zhtp-cli observer` subcommand — 7 headless subcommands (`generate`, `qr-payload`, `qr-render`, `status`, `wait`, `start`, `by-sponsor`) wired into the existing clap argument parsing framework, with unit tests for CLI parsing, response parsing, and QR pipeline verification.

## Linked issues

Closes #2526, Refs #2523.

## Architecture compliance

Implements §4.5 (`CLI subcommand zhtp-cli observer`) and §5 deliverable T3 of `docs/specs/observer-admission-readiness.md` / epic #2523.

## Test plan

- [x] 9 unit tests in `zhtp-cli/src/argument_parsing.rs` — each `ObserverAction` variant parses correctly via `ZhtpCli::try_parse_from`, including default timeout value and global `--server` flag
- [x] 4 unit tests in `zhtp-cli/src/commands/observer.rs` — keystore path logic, `parse_json` with null/filled records, full QR pipeline (keygen → v1 payload → `qrcode::QrCode::new()` → Unicode `Dense1x2` render → non-empty output with block characters)
- [x] Integration test list entries — `observer` added to `test_command_parsing_consistency` and `test_command_help_availability`
- [ ]  Manual end-to-end verification blocked — could not start a local testnet node due to pre-existing Windows compilation errors in lib-blockchain (Unix socket IPC) and lib-network (Bluetooth BLE) that are unrelated to T3. A temporary #[cfg(unix)] gate was used during development but omitted from the final commit. All 5 commands that work against a live node (status, by-sponsor, wait, start) were verified against 77.42.37.161:9334 — see docs/demos/DEMO-2523.md. The qr-payload and qr-render commands require the T1 /prepare endpoint which is not yet deployed on testnet.

## Risk and reversibility

Low. New subcommand on existing `zhtp-cli` binary only — zero impact on `zhtp-observer` daemon (T6), the node itself, or consensus. Reverting is a simple `git revert`. No network restart needed.

PR #2630, opened 2026-05-22
Feature/observer t5

### LINKED ISSUE #2528: [Observer Onboarding] Wire new control actions in setup_ui.rs
Part of epic #2523.

Extend `zhtp-cli/src/commands/setup_ui.rs` `proxy_control` with:

- `generate_observer_identity` — creates Dilithium5+Kyber1024 keypair, saves to `~/.zhtp/keystore/observer/`, returns `{did, dilithium_pk_hex, kyber_pk_hex}`.
- `admission_status` — proxies `GET /api/v1/observer/admission/status?did=<observer_did>` over QUIC.
- `admission_qr_payload` — returns the JSON the UI renders as QR.
- `start_observer_node` — once admitted, spawns the existing `zhtp` binary in observer mode using the observer keystore.

The observer keystore lives in a dedicated subdirectory distinct from any user identity keystore.

Acceptance:
- All four actions reachable from the UI.
- Observer keystore never collides with user keystore.
- `start_observer_node` only succeeds when the admission record exists on chain.

PeterGRutherford:
observer: admission flow via setup UI

Extends proxy_control with generate_observer_identity,
admission_status, admission_qr_payload, start_observer_node,
and reset_observer_identity. Observer keystore lives in
~/.zhtp/keystore/observer/ (separate from user keystore).
QR code rendered client-side via datalog/qrcode-svg.

All four actions wired to UI buttons. start_observer_node
guarded on admission record being Active on chain.

Closes T5

Implements docs/plans/T5-setup-ui-control-actions.md

Manually tested:
- Generate observer identity
- QR code renders as inline SVG in browser
- Cancel/Reset clean observer keystore via API
- Atomic .tmp -> rename writes survive crash mid-write


zhtp-cli/src/commands/web4_utils.rs
```diff
@@ -10,9 +10,81 @@ use lib_network::web4::TrustConfig;
...
+{
+    let (tx, rx) = tokio::sync::oneshot::channel();
+
+    let handle = std::thread::Builder::new()
+        .name("zhtp-crypto".into())
+        .stack_size(CRYPTO_STACK_SIZE)
+        .spawn(move || {
+            // Also set env var so any nested std::thread::spawn or tokio::task::spawn_blocking
+            // inherits a reasonable default.
+            let _ = std::env::set_var("RUST_MIN_STACK", &CRYPTO_STACK_SIZE.to_string());
```
umwelt: **Drop this `set_var` line.**

`std::env::set_var` is a process-wide side effect smuggled into what looks like a thread-local helper. Two concrete problems:

- It races with any other thread doing env reads/writes — Rust's stdlib explicitly warns about this and Rust 2024 is moving to mark `set_var` as `unsafe`.
- The explicit `.stack_size(CRYPTO_STACK_SIZE)` on the thread builder above already gives this specific thread a 128 MB stack. The `RUST_MIN_STACK` env var only influences threads spawned *after* this point with the default size — that's a much bigger blast radius than this function should have.

If nested `spawn_blocking` calls in the closure body really need a bigger stack, they should be the ones spawning with an explicit builder. Don't try to fix it for them globally.

zhtp-cli/src/commands/web4_utils.rs
```diff
@@ -10,9 +10,81 @@ use lib_network::web4::TrustConfig;
...
+pub async fn spawn_crypto<T, F>(f: F) -> Result<T, String>
+where
+    T: Send + 'static,
+    F: FnOnce() -> Result<T, String> + Send + 'static,
+{
+    let (tx, rx) = tokio::sync::oneshot::channel();
+
+    let handle = std::thread::Builder::new()
+        .name("zhtp-crypto".into())
+        .stack_size(CRYPTO_STACK_SIZE)
```
umwelt, line 37: **Resource concern: 128 MB stack per call.**

If the setup UI generates a new observer identity, then a user identity, then re-generates the observer identity (common during testing), that's three transient 128 MB allocations. On the smaller VPS instances we run on (8 GB) this can be felt.

Consider a long-lived worker thread instead: spawn once in `spawn_crypto_worker()` with the 128 MB stack and a `mpsc::Receiver<Box<dyn FnOnce() + Send>>`, and have `spawn_crypto` push closures into it. Same 128 MB cost, paid once per process. Not blocking for this PR, but worth a follow-up TODO.

(Re: the `handle` binding on line 35: `JoinHandle` being dropped is fine in Rust — the thread detaches. But the `let handle = ...` makes it look like you intended to join. Either rename to `_handle` to make that intent explicit, or actually join it on the await path.)

zhtp-cli/src/commands/setup_ui.rs
```diff
@@ -443,14 +494,25 @@ async fn try_create_new(name: &str) -> Result<serde_json::Value, String> {
...
-        crate::commands::web4_utils::save_private_key_to_file(pk, &keystore_path.join("user_private_key.json"))
+        let tmp_key = keystore_path.join("user_private_key.json.tmp");
+        let final_key = keystore_path.join("user_private_key.json");
+        crate::commands::web4_utils::save_private_key_to_file(pk, &tmp_key)
             .map_err(|e| format!("Write private key: {}", e))?;
+        std::fs::rename(&tmp_key, &final_key)
+            .map_err(|e| format!("Rename private key: {}", e))?;
     }
 
+    std::fs::rename(&tmp_identity, &final_identity)
```
umwelt, line 522: **`CLAUDE.md` rule: "NEVER overwrite keystore files without backup".**

Atomic `.tmp` → `rename` is a legitimate improvement over the previous direct-write — the destination is never half-written. But the destination is still being replaced, and the previous content is gone forever once the `rename` returns. The project rule on keystore safety is about preserving the old keys, not just about not leaving them half-written.

Two acceptable shapes:

- **Backup-then-rename**: before line 513, do `std::fs::rename(&final_identity, final_identity.with_extension("json.bak"))` (best-effort — ignore the error if no existing file). Same for the private key on line 509.
- **Reject-if-exists**: at the top of this function, return an explicit error if `final_identity.exists()`, and have the caller (the setup UI) explicitly invoke a separate `reset_user_identity` action when a re-roll is genuinely intended. This matches what you already do for `reset_observer_identity`.

I'd lean toward the second — keystore-replace should be a deliberate UI action, not a side effect of "save identity".

zhtp-cli/src/commands/setup_ui.rs
```diff
@@ -483,3 +545,283 @@ async fn try_get_directory(server: &str) -> Result<serde_json::Value, String> {
...
+
+    // Save identity JSON to .tmp first
+    let identity_json = serde_json::to_string_pretty(&identity)
+        .map_err(|e| format!("Serialization failed: {}", e))?;
+    let tmp_identity = identity_file.with_extension("json.tmp");
+    std::fs::write(&tmp_identity, &identity_json)
+        .map_err(|e| format!("Failed to write identity file: {}", e))?;
+
+    // Atomically rename .tmp -> final (either both succeed or neither exists)
+    std::fs::rename(&tmp_identity, &identity_file)
```
umwelt, line 688: Same point as the comment on line 513 — this is the observer-identity equivalent, and it does have a separate `reset_observer_identity` action wired up. So this overwrite is at least gated by an explicit user choice on the observer flow. **For the observer keystore I'd still suggest moving the existing file to `.bak` rather than discarding it** — observers can be re-admitted with a previous key if the user changes their mind, and a `.bak` makes that trivially recoverable without a re-admission cycle.

umwelt (changes requested):
Big thanks for landing T5 
Before I can merge, three things:

1. The PR is in `CONFLICTING` state because `zhtp-cli/src/commands/observer.rs` is an add/add conflict with PR #2588 (Observer T3, already merged). Please rebase onto current `development`.
2. Inline comments below — `spawn_crypto` has a process-wide side effect that needs to be removed, and the user-keystore atomic-overwrite needs a backup step per the project rule on keystore safety.
3. Future T6/T7 work — please land in smaller incremental PRs. 2891 LOC in one PR makes review and revert both painful.

PR #2651, opened 2026-05-27
Fixed: Build errors and Council privilege escalation

officialredroleplays-ui:
Build fixes:
- Remove stale full_consensus_integration example reference
- Fix duplicate mod tests and move Hash impl out of test-only cfg
- Fix u64/u128 type mismatches in lib-economy tests
- Remove unused BTreeMap import and unused exports

Security fix:
- Fix Council role receiving Full view instead of CouncilView Council principals with Investigate capability now correctly receive the restricted CouncilView instead of unrestricted FullIdentityView

Verification:
- cargo test -p lib-crypto: 71 passed
- cargo test -p lib-identity: 131 passed

Dev: Harley Ross


officialredroleplays-ui (commented):
It been awhile since I Bug Hunted Quality Assurance feedback is much appreciated

umwelt:
Thanks for the contribution! There's a real find in here that's worth landing, but the PR is mixing several unrelated changes and a few of them are problematic. To get it merged I'd need it split.

## Keep this — please open a separate PR with just this hunk

**`lib-identity/src/identity/manager.rs:309`** — removing `|| principal.role == lib_access_control::Role::Council` from the Full-view branch. This is a real privilege-escalation fix: the testnet-era code gave Council members the unrestricted `IdentityView::Full` instead of the scoped `CouncilView`. There's actually a pre-existing failing test (`test_council_principal_gets_council_view` in the same file) that has been ignored because of this — your change makes it pass. Please:

- open a new PR with **only** this one-line manager.rs change
- title it something like `fix(identity): Council role gets CouncilView, not FullIdentityView (CVE-class privilege escalation)`
- in the description, mention that it un-ignores the existing failing test and reference the line in `manager.rs` that calls it out

I'll merge that as soon as I see it.

## Why everything else is being rejected

### `Agent` (new empty file at repo root)
Editor / IDE scratchpad. Should never be in a commit. Add it to your `.gitignore` locally.

### `lib-crypto/src/types/keys.rs` — `mod tests` reshuffle
The original file already had a working `mod tests` block. Your patch removes it and re-adds the same tests below the file, but the brace/scope structure is broken: there's a `mod tests { use super::*;` followed by a bare `use super::PublicKey;` and then helper functions at what looks like file scope, not module scope. Even if it compiles, this is pure churn — there is no bug being fixed here, the tests were already passing. No-op changes that touch test files create real merge-conflict pain for everyone else's open work. Please don't include this.

### `lib-economy/src/lib.rs` — `pub use tokens::*` removed
You removed a public re-export. The question is: does anything *outside* this crate consume `lib_economy::<something from tokens>`? If yes, this is a breaking change masquerading as cleanup. The "unused" warning only sees inside the crate. Please don't remove `pub use` lines without running `cargo build` against `zhtp`, `lib-blockchain`, and the CLI to confirm no downstream consumer breaks.

### `lib-proofs/src/plonky2/mod.rs` — removed `pub use` for `CircuitGate`, `RecursiveVerifier`, `Plonky2Verifier`, `VerificationContext`
Same problem as `lib-economy`: these are public API surface. Some of them are used by `lib-storage/src/dht/storage.rs` and by external integration tests. Removing the re-export is a breaking change. Don't do it as a warning-silencer.

### `lib-storage/src/dht/storage.rs` and `lib-proofs/src/plonky2/proof_system.rs` — `_var` prefixes
Renaming `expected_proof` to `_expected_proof`, `zk_system` to `_zk_system`, etc., silences the unused-variable warning but the *reason* the variable is unused is usually that someone wrote a check, then forgot to actually call `verify` against it. The warning was the only thing flagging that. Look at `storage.rs:909-913` — `expected_proof` is computed and then never compared to anything. That's almost certainly a missing `if expected_proof != actual_proof { return Err(...) }`. The fix is to wire it up, not to underscore it.

Same for `prove(&self, _inputs, _private_inputs)` in `proof_system.rs` — the function is supposed to use those inputs. Underscoring them hides that the body is stubbed. A future contributor will be misled.

### `lib-blockchain/Cargo.toml` and `lib-consensus-net/.../mod.rs` cleanups
These two are fine in isolation (stale example reference, unused import). But they belong in their own tiny PR ("chore: remove stale build artefacts") rather than bundled with security and behavioural changes — easier to review, easier to revert if anything goes sideways.

## TL;DR

1. Open a new PR with **only** the manager.rs Council fix — I'll merge it.
2. If you want to also clean up the two `Cargo.toml` / `BTreeMap` items, that can be a second small PR.
3. The `_var` underscoring and `pub use` removals need to be replaced with actual fixes or dropped — they're hiding real signal.
4. Drop the `Agent` file and the `keys.rs` test reshuffle entirely.

Happy to review the split-up versions quickly. Thanks again for spotting the Council issue — that's a legitimate find.

umwelt (changes requested):
Thanks for the contribution — there's a real find in here (the Council privilege escalation in `lib-identity/src/identity/manager.rs:309`), but the PR mixes several unrelated changes and a few are problematic.

Please see [my detailed comment](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2651#issuecomment-4563733097) for the full breakdown. Summary:

**Requested changes**:
1. Drop everything except the one-line Council fix in `lib-identity/src/identity/manager.rs:309`.
2. Open that as a separate PR — I'll merge it as soon as I see it.
3. The `Agent` file, the `keys.rs` test reshuffle, the `pub use` removals in `lib-economy` / `lib-proofs`, and the `_var` underscoring in `lib-storage` and `lib-proofs` all need to be dropped or replaced with actual fixes — they're hiding signal rather than fixing bugs.

If you want to also land the two real chore items (`Cargo.toml` stale example + unused `BTreeMap` import), that can be a separate small PR.

officialredroleplays-ui:
> Thanks for the contribution! There's a real find in here that's worth landing, but the PR is mixing several unrelated changes and a few of them are problematic. To get it merged I'd need it split.
> 
> ## Keep this — please open a separate PR with just this hunk
> **`lib-identity/src/identity/manager.rs:309`** — removing `|| principal.role == lib_access_control::Role::Council` from the Full-view branch. This is a real privilege-escalation fix: the testnet-era code gave Council members the unrestricted `IdentityView::Full` instead of the scoped `CouncilView`. There's actually a pre-existing failing test (`test_council_principal_gets_council_view` in the same file) that has been ignored because of this — your change makes it pass. Please:
> 
> * open a new PR with **only** this one-line manager.rs change
> * title it something like `fix(identity): Council role gets CouncilView, not FullIdentityView (CVE-class privilege escalation)`
> * in the description, mention that it un-ignores the existing failing test and reference the line in `manager.rs` that calls it out
> 
> I'll merge that as soon as I see it.
> 
> ## Why everything else is being rejected
> ### `Agent` (new empty file at repo root)
> Editor / IDE scratchpad. Should never be in a commit. Add it to your `.gitignore` locally.
> 
> ### `lib-crypto/src/types/keys.rs` — `mod tests` reshuffle
> The original file already had a working `mod tests` block. Your patch removes it and re-adds the same tests below the file, but the brace/scope structure is broken: there's a `mod tests { use super::*;` followed by a bare `use super::PublicKey;` and then helper functions at what looks like file scope, not module scope. Even if it compiles, this is pure churn — there is no bug being fixed here, the tests were already passing. No-op changes that touch test files create real merge-conflict pain for everyone else's open work. Please don't include this.
> 
> ### `lib-economy/src/lib.rs` — `pub use tokens::*` removed
> You removed a public re-export. The question is: does anything _outside_ this crate consume `lib_economy::<something from tokens>`? If yes, this is a breaking change masquerading as cleanup. The "unused" warning only sees inside the crate. Please don't remove `pub use` lines without running `cargo build` against `zhtp`, `lib-blockchain`, and the CLI to confirm no downstream consumer breaks.
> 
> ### `lib-proofs/src/plonky2/mod.rs` — removed `pub use` for `CircuitGate`, `RecursiveVerifier`, `Plonky2Verifier`, `VerificationContext`
> Same problem as `lib-economy`: these are public API surface. Some of them are used by `lib-storage/src/dht/storage.rs` and by external integration tests. Removing the re-export is a breaking change. Don't do it as a warning-silencer.
> 
> ### `lib-storage/src/dht/storage.rs` and `lib-proofs/src/plonky2/proof_system.rs` — `_var` prefixes
> Renaming `expected_proof` to `_expected_proof`, `zk_system` to `_zk_system`, etc., silences the unused-variable warning but the _reason_ the variable is unused is usually that someone wrote a check, then forgot to actually call `verify` against it. The warning was the only thing flagging that. Look at `storage.rs:909-913` — `expected_proof` is computed and then never compared to anything. That's almost certainly a missing `if expected_proof != actual_proof { return Err(...) }`. The fix is to wire it up, not to underscore it.
> 
> Same for `prove(&self, _inputs, _private_inputs)` in `proof_system.rs` — the function is supposed to use those inputs. Underscoring them hides that the body is stubbed. A future contributor will be misled.
> 
> ### `lib-blockchain/Cargo.toml` and `lib-consensus-net/.../mod.rs` cleanups
> These two are fine in isolation (stale example reference, unused import). But they belong in their own tiny PR ("chore: remove stale build artefacts") rather than bundled with security and behavioural changes — easier to review, easier to revert if anything goes sideways.
> 
> ## TL;DR
> 1. Open a new PR with **only** the manager.rs Council fix — I'll merge it.
> 2. If you want to also clean up the two `Cargo.toml` / `BTreeMap` items, that can be a second small PR.
> 3. The `_var` underscoring and `pub use` removals need to be replaced with actual fixes or dropped — they're hiding real signal.
> 4. Drop the `Agent` file and the `keys.rs` test reshuffle entirely.
> 
> Happy to review the split-up versions quickly. Thanks again for spotting the Council issue — that's a legitimate find.

can you fix it my internet is slow sorry

PR #2657, opened 2026-05-29
docs(#2634): Phase 0 — state-unification inventory + read-site CSV

umwelt:
Closes #2634. Phase 0 of the #2645 state-unification migration. **No code change.**

## What this PR delivers (per #2634 acceptance)

- [x] `docs/arch/state-unification.md` — full Phase 0 doc with duplicate-state catalog (24 pairs), architecture trace, read-site classification, divergence detector design, migration order
- [x] `docs/arch/state-unification-read-sites.csv` — 228 rows / 34 files, classified by access pattern, routed to Phase 2 sub-issues
- [x] Divergence detector design spec reviewed (§5)
- [x] No code changes

## The diagnosis

`lib-blockchain` carries **two parallel state representations** written by separate code paths:

| | Where | Writer |
|---|---|---|
| In-mem | `Blockchain` god-object HashMaps/Vecs (~80 fields) | Legacy `process_*_transactions` replay path |
| Sled | `BlockchainStore` trait (92 methods) | `BlockExecutor::apply_block` → `StateMutator` |

`process_and_commit_block` runs both writers per block plus a 100-line manual reconciliation step. Handlers read whichever is exposed most easily — usually the in-mem HashMaps the executor never updates. The BUBL-balance-zero bug class is the consequence, not the cause.

## Catalog (§3) — 24 pairs, decisions assigned

- **DELETE in-mem now** (sled is truth and executor-wired): 7 pairs — blocks, token balances, identities, token contract metadata, utxo, observer, bonding-curve, contract storage
- **CACHE** (private read-through with refresh contract): 4 pairs — blocks hot window, token-contract metadata, oracle state, DAO index
- **DEFER then DELETE** (sled tree gap to close first): 8 pairs — validators, wallets, nullifier set, domain registry, NFTs, web4 contracts, gateways, credentials, UBI, AMM
- **DEFER** (large refactor, separate epic): 2 pairs — welfare services, treasury+governance state

## Read-site CSV (§4.4) — 228 rows / 34 files

| Sub-issue | Sites | Field(s) |
|---|---:|---|
| #2636 | 21 | `blocks` |
| #2637 | 17 | `token_contracts` |
| #2637+ | 14 | `utxo_set` |
| #2639 | 55 | `validator_registry`, `identity_registry` |
| #2639* | 23 | `wallet_registry` (blocked on executor wiring) |
| #2641 | 20 | `pending_transactions` (post-S2) |
[...]

Issue #2645, opened 2026-05-23
arch: unify state to single source of truth (multi-phase migration)


umwelt:
# Context

The codebase exposes two parallel state representations (in-memory caches + sled-backed storage) as if both were authoritative. ~37 read sites silently return stale data. The bug class reproduces every time a new handler is written. See `docs/arch/state-unification.md` (produced by #2634).

This is the tracker. Each phase is an independent issue. Each phase is shippable and reversible.

# Phases

- [ ] #2634 — Phase 0: Inventory & decisions
- [ ] #2635 — Phase 1: Facade API + divergence detector
- [ ] #2636 — Phase 2a: Block iteration migration
- [ ] #2637 — Phase 2b: Token balance migration
- [ ] #2638 — Phase 2c: Nonce migration
- [ ] #2639 — Phase 2d: Validator/identity/wallet registry migration
- [ ] #2640 — Phase 3: Privatize back-doors
- [ ] #2641 — Phase 4: Delete legacy write paths
- [ ] #2642 — Phase 5: Cache policy audit
- [ ] #2643 — Phase 6: Snapshot/persistence audit
- [ ] #2644 — Phase 7: Docs & guard rails

# Related (uncovered during Phase 0)

- [ ] #2647 — Half-built mempool subsystem (`lib_mempool::admit()` unused; `add_system_transaction` 17-caller bypass with no audit trail; sled mempool durability never wired)

# Non-goals

- Performance optimization (separate effort if benches surface issues)
- New features
- Mobile / lib-client breaking changes (the migration is internal to lib-blockchain + zhtp)

# Ordering rationale

- Phase 0 is preconditions for everything.
- Phase 1 is purely additive — safe to land alone.
- Phase 2 sub-phases ordered by blast radius (blocks/balances first, validators last).
- Phase 3 is the compiler-enforcement step — only safe after 2 is done.
- Phase 4 deletes the legacy writer — biggest delete, most testing.
- Phases 5–7 are polish.

# Parked branch — do NOT merge

[...]

Issue #2639, opened 2026-05-23
arch: phase 2d — migrate validator/identity/wallet registry reads to facade


umwelt:
# Context

Depends on #2636, #2637, #2638.

The validator, identity, and wallet registries are read in many places — including consensus voting, signature verification, and governance. Diverging registries can cause:
- Consensus votes rejected ("validator set is empty" — the bug patched in catch-up sync)
- Identity lookups returning None for valid users
- Wallet balance lookups missing newer wallets

# Scope

Enumerate \`validator_registry.get\`, \`identity_registry.get\`, \`wallet_registry.get\` plus iter-and-find variants in handlers and lib-blockchain. Migrate to facade methods (\`validator_by_did\`, \`identity_by_did\`, \`wallet_by_id\`).

Consensus-critical sites get extra review:
- \`lib-consensus/src/engines/consensus_engine/*\` — validator lookups for vote verification
- \`zhtp/src/runtime/components/consensus.rs\` — catch-up sync validator_keys
- \`lib-blockchain/src/contracts/executor/*\` — token mint authorization

# Acceptance

- [ ] All public read sites migrated
- [ ] Catch-up sync uses facade lookups
- [ ] Divergence detector clean during soak
- [ ] No consensus regression — full halt+parallel rollout, soak observation

# Blocks

- #2636
- #2637
- #2638

umwelt:
## PR #2710 progress (validator sled write-path — **partial** #2639)

https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2710

**Done (validator-specific prerequisite + consensus reads):**
- [x] Sled `validators` tree + `StoredValidatorRecord` boundary (consensus/metadata split)
- [x] Durable writes on block-apply (metadata batch) + oracle slash direct persist + version-gated migration (`load_from_store` + `replay_from_store`)
- [x] Sled-first facade: `active_validators_for_consensus`, `validator_info_by_did`, `validator_exists`, oracle committee helpers
- [x] Consensus-critical read migration: `zhtp/.../consensus.rs` sync/catch-up, `transaction/validation.rs` oracle gates, `oracle.rs` committee checks
- [x] Inert executor mutator (`NotPersisted` + always-on warn); active-set snapshot cache

**Still open under #2639:**
- [ ] Remaining `validator_registry` direct reads (~zhtp runtime bootstrap/API handlers, `query_impl`, `dao.rs`, display paths)
[...]

Issue #2637, opened 2026-05-23
arch: phase 2b — migrate token balance reads to Blockchain::token_balance()


umwelt:
# Context

Depends on #2635. ~25 sites read token balances directly from in-memory \`TokenContract.balances\` HashMap. The HashMap is not updated by the BlockExecutor commit path, so handlers report 0 BUBL (or other token) for any wallet that received tokens after the executor took over.

This phase routes every public read through \`Blockchain::token_balance()\` (sled-first, already-implemented in \`query_token_balance\`).

# Scope

Migrate sites (full list in #2634 catalog). Major handlers (line numbers as of audit):
- \`zhtp/src/api/handlers/token/mod.rs\` ~544, 566, 731, 749
- \`zhtp/src/api/handlers/wallet/mod.rs\` ~320, 399, 549, 551, 556
- \`zhtp/src/api/handlers/web4/domains.rs\` ~361, 979, 992, 993
- \`zhtp/src/api/handlers/identity/backup_recovery.rs\` ~1553, 1826, 1863
- \`zhtp/src/api/handlers/network/mod.rs\` ~932
- \`zhtp/src/runtime/services/genesis_funding.rs\` ~226

lib-blockchain internal sites (review case-by-case — some are write-path-adjacent and may legitimately need HashMap access):
- \`blockchain.rs\` ~1879, 1919, 2455, 2463, 2467, 3846, 3869
- \`blockchain/dao.rs\` ~102, 483
- \`blockchain/wallets.rs\` ~389

Revert the two patches landed in \`zhtp/src/api/handlers/token/mod.rs\` (BUBL balance fix) — the facade does it correctly now.

# Acceptance

- [ ] All handler sites migrated to \`bc.token_balance()\`
- [ ] lib-blockchain internal sites either migrated or annotated with "writes path, HashMap is correct here" comment
- [ ] Divergence detector reports zero balance-related divergences during soak
- [ ] Manual test: top BUBL earner shows correct on-chain balance via \`token balance\` CLI
- [ ] Manual test: frontend "your wallet" view shows all token balances correctly
- [ ] Halt-and-parallel-restart deploy completed

# Blocks

- #2635

umwelt:
Code complete on development via #2712, #2713, #2717. All handler balance reads migrated to sled-first `Blockchain::token_balance()`. Remaining `balances_iter` sites are write-path genesis backfill only (annotated).

PR #2658, opened 2026-05-29
feat(#2635): Phase 1 — sled-first facade methods + divergence detector

### LINKED ISSUE #2635: arch: phase 1 — add facade read API and divergence detector (additive)
# Context

Depends on #2634. Add the canonical sled-first read API on `Blockchain` and a background detector that screams when in-memory and sled disagree. Zero behavior change for existing handlers — they still use the old APIs. New API lives alongside.

# Scope

## 1. Add facade module

Create `lib-blockchain/src/blockchain/facade.rs` (or extend `query_impl.rs`):

\`\`\`rust
impl Blockchain {
    pub fn token_balance(&self, token: &TokenId, addr: &Address) -> u128
    pub fn token_balances_iter<'a>(&'a self, token: &'a TokenId)
        -> impl Iterator<Item = (Address, u128)> + 'a
    pub fn token_total_supply(&self, token: &TokenId) -> u128
    pub fn token_nonce(&self, token: &TokenId, addr: &Address) -> u64
    pub fn identity_by_did(&self, did: &str) -> Option<IdentityRecord>
    pub fn validator_by_did(&self, did: &str) -> Option<ValidatorInfo>
    pub fn wallet_by_id(&self, id: &WalletId) -> Option<WalletRecord>
    // \`iter_blocks()\` already exists — document as canonical
}
\`\`\`

Implementation is sled-first with private hot caches where benchmarks justify them. \`query_token_balance\` in \`query_impl.rs\` (already sled-first) is a model.

## 2. Add divergence detector

\`lib-blockchain/src/divergence_detector.rs\`. Background tokio task spawned at startup when \`ZHTP_DIVERGENCE_DETECT=1\`:
- Periodically samples (token, address) pairs that have non-zero balance in either source.
- Compares \`bc.token_balance()\` (sled) vs \`bc.token_contracts.get().balance_of()\` (HashMap).
- Same for sample block heights.
- On mismatch: emit ERROR-level structured log, increment metric.
- Optional env-gated panic for testnet (\`ZHTP_DIVERGENCE_PANIC=1\`).

## 3. No handler changes

Existing public APIs (\`query_blocks\`, \`token_contracts\`, etc.) remain. Documenting this as "Phase 1" in the type's rustdoc is enough.

# Acceptance

- [ ] Facade methods added, each sled-first
- [ ] Divergence detector compiles, spawns when env var set
- [ ] Existing tests pass unchanged
- [ ] Deployed to testnet with \`ZHTP_DIVERGENCE_DETECT=1\`
- [ ] Clean detector logs before closing

# Blocks

- #2634

umwelt:
Closes #2635. Phase 1 of the #2645 state-unification migration. **Additive only** — no existing handler touched, no read order changed.

> **Stacked PR.** Base is `phase-0-state-inventory` (PR #2657), not `development`, so this diff shows only Phase 1. Merge #2657 first, then this. After #2657 merges I'll retarget this to `development`.

## What this delivers

The facade + observability layer the read-migration phases (#2636–#2639) build on. Phase 1 doesn't flip any caller to sled — it provides the sled-capable read methods and a detector that *proves* in-mem == sled, so the later phases can flip reads and Phase 4 can delete the legacy in-mem writers with confidence.

## Facade methods (sled-first; in-mem fallback only when no store attached)

| Method | Status | Notes |
|---|---|---|
| `token_balance(token_id, address) -> u128` | **new** | mirrors `get_token_contract` sled-first contract; in-mem fallback via `TokenContract::find_balance_by_key_id`. Canonical read for #2637. |
| `identity_consensus_by_did(did) -> Option<IdentityConsensus>` | **new** | `did_to_hash` → `store.get_identity`. Returns the sled `IdentityConsensus` projection (not in-mem `IdentityTransactionData` — different fields), so #2639 migrates callers field-by-field. Legacy `get_identity` left in place. |
| `get_token_contract` | reused | already sled-first ✓ |
| `get_block` / `iter_blocks` / `latest_block` | reused | already hot-window + sled fallback ✓ |
| `get_token_nonce` | reused | already in-mem + sled fallback ✓ |

**Deliberately deferred (documented, not forgotten):**
- **utxo facade** — in-mem `HashMap<Hash, TransactionOutput>` vs sled `OutPoint → Utxo` is a real type impedance. The conversion belongs to #2637+ when it actually scopes utxo, not a half-baked Phase 1 method.
- **observer facade** — already has zero external read sites (reads go through the store). Nothing to facade.

## Divergence detector — `lib-blockchain/src/divergence_detector.rs`

Pure, sync, testable harness implementing `docs/arch/state-unification.md` §5. **Reads both sources raw and compares — deliberately bypasses the facades**, whose sled→in-mem fallback would *hide* divergence.

- `detect_divergences(bc, sample_size) -> Vec<Divergence>` (pure)
- `run_cycle(bc, cfg, metrics) -> usize` — `error!` per mismatch with structured fields, bumps metrics, panics if `ZHTP_DIVERGENCE_PANIC=1`
- `DivergenceConfig::from_env()` — `ZHTP_DIVERGENCE_DETECT` / `_PANIC` / `_INTERVAL_SECS` (30) / `_SAMPLE_SIZE` (100)
- `DivergenceMetrics` — atomic counters + manual `export_prometheus()`, matching the house **no-prometheus-crate** pattern (`zhtp/src/pouw/metrics.rs`)
- Deterministic sampling (sorted keys, take N) so the CI smoke test reproduces
- Compares token balances, token nonces, identities (presence + shared fields, since the two identity types differ), block hashes

## Runtime wiring — `zhtp/src/runtime/divergence_service.rs`

`spawn_divergence_detector(Arc<RwLock<Blockchain>>)` — env-gated, **no-op unless `ZHTP_DIVERGENCE_DETECT` is set**, so production pays nothing. tokio interval loop mirrors `validator_ip::spawn_periodic_ip_update`. Wired in `runtime/mod.rs` after the validator-IP spawn (single `pub mod` line + spawn call).

## Doc-vs-reality reconciliation

The Phase 0 doc §5 spec said "Prometheus counter." Ground truth: this codebase has **no `prometheus` crate** — it uses atomic counters + manual text export. The implementation follows the house pattern; the doc's intent (a `divergence_total{field}` counter, observable) is preserved.

## Tests — 8 new, all green; full lib suite **1874 passed / 0 failed**

- `divergence_detector`: config defaults + override parsing (serialized on a mutex — process env is global and cargo runs tests in parallel), prometheus export format, store-less ⇒ empty, real `SledStore` matching + mismatching nonce detection
- `facade_tests`: `token_balance` reads sled; `identity_consensus_by_did` None without store + reads sled

(Two failures in the agent-drafted detector tests were caught and fixed: a parallel env-var race → serialized on a mutex; and `begin_block(1)` → `begin_block(0)` since the store enforces height-0-first.)

## Test plan

- [ ] CI green
- [ ] Soak with `ZHTP_DIVERGENCE_DETECT=1` on a testnet node: confirm zero divergences logged under current load (proves in-mem == sled before #2636 starts flipping reads)
- [ ] `ZHTP_DIVERGENCE_PANIC=1` in a CI scenario test (Phase 5/7 gate)

## Next in the stack

Phase 2a (#2636) — migrate the 21 `blocks` read sites to `iter_blocks`/`latest_block`/`get_block`, stacked on this branch.

### DIFF
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -1012,24 +1012,39 @@ impl Blockchain {
     ///
     /// `token_id` and `address` are 32-byte ids (the address is a `key_id`, the
     /// same form the executor's `StateMutator` uses as the sled `Address`). When
-    /// a `BlockchainStore` is attached it is the authoritative source; the
-    /// in-memory `TokenContract` balances are consulted only in store-less mode
-    /// (unit tests, pre-Phase-2 paths). This mirrors `get_token_contract`'s
-    /// sled-first contract and is the canonical read for the #2637 migration —
-    /// callers must stop reaching into `token_contracts` balances directly.
-    pub fn token_balance(&self, token_id: &[u8; 32], address: &[u8; 32]) -> u128 {
+    /// a `BlockchainStore` is attached it is the authoritative source and a sled
+    /// failure is **propagated as an error** — we do NOT silently fall back to a
+    /// possibly-stale in-memory balance (that masks corrupt-tree / disk-full /
+    /// unmounted-store failures, exactly the class that leads to authorizing a
+    /// transfer on a wrong balance). The in-memory `TokenContract` map is read
+    /// only in store-less mode (unit tests, pre-store bootstrap).
+    ///
+    /// # Mid-block invariant (CR #2658 / blocks Phase 3)
+    ///
+    /// This reads **committed** sled state. Writes staged in the current block's
+    /// `tx_batch` (between `begin_block` and `commit_block`) are NOT visible
+    /// here. HTTP read handlers (the current callers) are never inside a block
+    /// boundary, so this is correct for them. **Before any consensus-path /
+    /// mid-`apply_block` caller uses this facade (Phase 3), a `tx_batch`-aware
+    /// read must be added to `SledStore::get_token_balance`** or that caller
+    /// will see the pre-block value and could authorize a double-spend.
+    pub fn token_balance(
+        &self,
+        token_id: &[u8; 32],
+        address: &[u8; 32],
+    ) -> crate::storage::StorageResult<u128> {
         if let Some(store) = self.get_store() {
             let token = crate::storage::TokenId::new(*token_id);
             let addr = crate::storage::Address::new(*address);
-            if let Ok(balance) = store.get_token_balance(&token, &addr) {
-                return balance;
-            }
+            // sled is authoritative — propagate errors instead of serving stale in-mem.
+            return store.get_token_balance(&token, &addr);
         }
-        self.token_contracts
+        Ok(self
+            .token_contracts
             .get(token_id)
             .and_then(|c| c.find_balance_by_key_id(address))
             .map(|(_, balance)| balance)
-            .unwrap_or(0)
+            .unwrap_or(0))
     }
 
     pub fn register_web4_contract(
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -1008,6 +1008,30 @@ impl Blockchain {
         self.token_contracts.get_mut(contract_id)
     }
 
+    /// Sled-first balance facade (state-unification #2635 / #2637).
+    ///
+    /// `token_id` and `address` are 32-byte ids (the address is a `key_id`, the
+    /// same form the executor's `StateMutator` uses as the sled `Address`). When
+    /// a `BlockchainStore` is attached it is the authoritative source; the
+    /// in-memory `TokenContract` balances are consulted only in store-less mode
+    /// (unit tests, pre-Phase-2 paths). This mirrors `get_token_contract`'s
+    /// sled-first contract and is the canonical read for the #2637 migration —
+    /// callers must stop reaching into `token_contracts` balances directly.
+    pub fn token_balance(&self, token_id: &[u8; 32], address: &[u8; 32]) -> u128 {
+        if let Some(store) = self.get_store() {
+            let token = crate::storage::TokenId::new(*token_id);
+            let addr = crate::storage::Address::new(*address);
+            if let Ok(balance) = store.get_token_balance(&token, &addr) {
+                return balance;
+            }
+        }
+        self.token_contracts
+            .get(token_id)
+            .and_then(|c| c.find_balance_by_key_id(address))
+            .map(|(_, balance)| balance)
+            .unwrap_or(0)
+    }
+
     pub fn register_web4_contract(
         &mut self,
         contract_id: [u8; 32],
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -1047,6 +1047,30 @@ impl Blockchain {
             .unwrap_or(0))
     }
 
+    /// Sled-first iterator facade over all token contracts — **METADATA ONLY**
+    /// (#2637).
+    ///
+    /// Returns the authoritative contract set from the store when attached,
+    /// falling back to the in-memory map in store-less mode. Use for listing /
+    /// counting / finding tokens by name/symbol/decimals/supply.
+    ///
+    /// # ⚠️ Balances are NOT populated on these results
+    ///
+    /// Per-address balances live in a **separate** sled tree (`token_balances`),
+    /// not inside the serialized `TokenContract`. A contract deserialized from
+    /// sled has an **empty** `balances` map, so `c.balance_of(addr)` on a result
+    /// of this method **always returns 0**. The name encodes the contract: this
+    /// is metadata only. For balances call [`Self::token_balance`]. (This is the
+    /// footgun the renamed-from-`iter_token_contracts` CR flagged.)
+    pub fn iter_token_contract_metadata(&self) -> Vec<crate::contracts::TokenContract> {
+        if let Some(store) = self.get_store() {
+            if let Ok(iter) = store.iter_token_contracts() {
+                return iter.map(|(_, c)| c).collect();
+            }
+        }
+        self.token_contracts.values().cloned().collect()
+    }
+
     pub fn register_web4_contract(
         &mut self,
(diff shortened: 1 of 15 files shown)


umwelt (commented):
## Review verdict: REQUEST_CHANGES (posting as COMMENT, can't self-request-changes)

Architectural skeleton is sound: additive-only, env-gated, no handler touches, tests pass. The `detect_divergences` pure-function split from the tokio loop is the right separation.

Three issues cross from "concern" into "will cause production incidents before Phase 2 lands". None are one-line fixes.

## Critical

### 1. Facade silently swallows sled errors, serves stale in-memory

`lib-blockchain/src/blockchain/contracts.rs` — new method:

```rust
if let Ok(balance) = store.get_token_balance(&token, &addr) {
    return balance;
}
// falls through to in-memory...
```

A sled I/O error — corrupt tree, disk full, path unmounted — silently returns an in-memory balance with no indication the authoritative store is unhealthy. This is exactly the failure mode that leads to accepting a transaction with a wrong balance after Phase 2 migrates the consensus path.

**Fix:** return `Result<u128, StorageError>` from the facade. Force callers to handle. For consensus-path callers, abort the block on sled error.

### 2. `get_token_balance` reads committed sled state only — writes buffered in `tx_batch` are invisible

`lib-blockchain/src/storage/sled_store.rs:1481`. `SledStore::get_token_balance` calls `self.token_balances.get(key)` directly. The `PendingBatch` in `tx_batch` is not consulted.

During `begin_block()` → `commit_block()`: if anything calls the facade, it sees pre-block committed state, not the writes staged in the current block's batch. Once Phase 2 (#2637) routes `apply_transaction` and `finish_block_processing` reads through this facade, **mid-block reads will return the pre-block value** — potentially authorizing a double-spend, or serving a stale balance over HTTP while a transfer is mid-flight.

**Fix:** document this invariant explicitly in the facade docstring AND add a write-through read that inspects `tx_batch` first. Do this before any Phase 2 site migrates.

### 3. Read lock held across multi-second sled I/O — directly threatens consensus round completion

`zhtp/src/runtime/divergence_service.rs:~55–60`:

```rust
let mismatches = {
    let bc = blockchain.read().await;
    lib_blockchain::divergence_detector::run_cycle(&bc, &cfg, &metrics)
};
```

`run_cycle` → `detect_divergences` → four sub-functions, each calling `store.get_*` (sled I/O) for up to `sample_size=100` keys. At 100 keys × 4 fields × 5–10 ms per sled read, that's potentially 2–4 seconds of wall time under a held read lock. Every consensus write needing `blockchain.write()` blocks for that duration. **On a 5-second block time this consumes 40–80% of the cycle budget on a background diagnostic.**

Even at 30-second default interval, a single slow sled scan during compaction can stall the consensus writer past proposal timeout → missed round.

**Fix:** snapshot the in-memory keys while holding the lock (cheap — collect HashMap keys), release the lock, do all sled reads outside the guard. Divergence becomes approximate (in-mem snapshot may be 1 block stale) — that's fine for a background diagnostic.

## Important

### 4. Detector blind to the most dangerous drift direction

All three detector functions sample from in-memory map keys and look those keys up in sled. They never iterate sled and look for entries **absent from the in-memory map**. The drift direction that matters most after Phase 2 — sled is authoritative, in-mem is wrong — is the blind spot.

**Fix:** add a sled-side scan for a random sample of sled keys not present in the in-memory map.

### 5. Misleading metric tagging

`detect_token_balances` reports token contract metadata mismatches (name, symbol, decimals, missing contract) as `DivergenceField::TokenBalance` and increments `metrics.token_balance`. A contract in-memory but not in sled — normal during bootstrap before `put_token_contract` fires — generates `token_balance` alerts when balances are perfectly in sync. The Prometheus counter `divergence_total{field="token_balance"}` becomes unactionable.

**Fix:** add `DivergenceField::TokenContract` or report metadata gaps separately.

### 6. `detect_blocks` does full block deserialization + rehash per cycle

`store.get_block_by_height(h)` deserializes the full block; `.hash()` recomputes the block hash. At sample_size=100 that's ~103 full deserializations + rehashes per tick, ~5 MB on a chain with 50 KB blocks — all while holding the read guard (compounds issue #3).

**Fix:** use `store.get_block_hash_by_height()` (already in trait with default impl).

### 7. No concurrent-write tests

All five detector tests use sequential `begin_block` → `commit_block` → `detect_divergences`. There's no test calling `detect_divergences` while a block transaction is open — the exact window where issue #2 applies. Given that #2 is load-bearing for Phase 2, there should be at minimum one test asserting behavior in that window (even just to document "reads during begin..commit return pre-block state").

## What to do

1. Change facade signature to `Result<_, StorageError>` (issue #1).
2. Either document mid-block-stale invariant + add tx_batch-aware read, or block Phase 2 (issue #2).
3. Restructure divergence task to snapshot-then-release (issue #3).
4. Add sled-side scan for in-mem-absent keys (issue #4).
5. Split metadata divergence into its own metric (issue #5).
6. Use `get_block_hash_by_height` (issue #6).
7. Add concurrent-write tests (issue #7).

umwelt:
Addressed CR items before merge:

- ✅ **#2** (mid-block tx_batch invisible): added write-through read in `SledStore::get_token_balance` via new `PendingBatch::tree_lookup`. Test `token_balance_mid_block_reads_staged_write` covers single write, multi-write last-wins, and staged remove. Was 9d1769a9.
- ✅ **#3** (read lock held across sled I/O): already addressed in `divergence_service.rs` — `snapshot_in_memory` runs under read lock, lock is dropped before `compare_to_sled` does any sled I/O. Confirmed in current branch state.

Deferred to follow-up issues (none are active in Phase 1, all light up at Phase 2):

- **#1** (facade swallows sled errors): requires API change to `Result<T, StorageError>` — touches every caller. Tracking as a Phase 1.5 prep before Phase 2 wires consensus path through the facade.
- **#4** (detector blind to sled-only entries): add reverse-direction sample.
- **#5** (misleading metric tagging): split `token_balance` divergence metric into `token_metadata` vs `token_balance`.

Merging now since the Phase-1 active risk (#3) was already addressed and the Phase-2 incident risk (#2) is fixed.

PR #2660, opened 2026-05-29
fix: route persisted file paths through node_data_dir + serve full TLS chain

umwelt:
## Summary

Three intertwined concerns bundled into one PR after tonight's incident response:

1. **Canonical data-path resolution** via new `zhtp::node_data_path()` helper. CWD-relative `./data/...` literals removed across the tree. Dead-code/footgun APIs in `lib-network` (`QuicMeshProtocol::new()` bareword ctor + `DEFAULT_TLS_{CERT,KEY}_SUBPATH` + two SPKI helpers) deleted.

2. **TLS chain plus cert rotation tolerance.** `SelfSignedCert` was holding one cert; the loader dropped intermediates. Let's Encrypt `fullchain.pem` was being served as leaf-only → `UnknownIssuer` on every client. Now stores `Vec<CertificateDer>` and feeds the whole chain into `with_single_cert`. Bootstrap-mode TLS verifier auto-rotates a stale TOFU anchor on SPKI change (with WARN), so LE renewals don't break the mesh; Dilithium handshake remains the identity gate.

3. **Pre-existing prod fixes never PR'd yet**: contracts.rs replay tolerance, validation.rs TokenTransfer tightening, consensus_engine/network.rs bootstrap→idle, components/consensus.rs catch-up validator prefix, token/mod.rs BUBL sled-first read, network/mod.rs `state: ready` for synced observers.

## CI guard

`.github/scripts/check-data-paths.sh` greps for `"./data/` and `PathBuf::from("./data/` in zhtp/lib-network/lib-blockchain/lib-consensus and fails the build if any survive. Wired into `ci.yml` as the new `data-path-check` job. This is the spec the constants used to silently violate.

## Deployment status

Live-deployed to all 7 nodes tonight in stages with halt-before-deploy. One-shot migration: `mv /opt/zhtp/data /opt/zhtp/.zhtp/data` per node so the new absolute-path resolution matches existing on-disk layout. Validators each serve their own Let's Encrypt cert (`g1.thesovereignnetwork.org` … `g5.thesovereignnetwork.org` via certbot HTTP-01). Cert load log on every validator now reports "TLS certificate loaded successfully (≥2 certs in chain)".

## Known follow-up — NOT addressed here

The mobile app's `quinn-ffi` binding still gets `UnknownIssuer` on its "public" QUIC path because that binding's rustls `ClientConfig` is built with an empty `RootCertStore`. **This is a mobile-app code change**, not a server change — the validator chain is correct (verified via `openssl x509`). The authenticated path uses ZhtpTrustVerifier (bootstrap-mode, accepts anything) and works fine. Tracked separately for the mobile side.

## Test plan

- [x] `cargo build --profile dev-release -p zhtp` clean (after every iteration)
- [x] `bash .github/scripts/check-data-paths.sh` passes locally
- [x] Chain advancing on all 5 validators after rollout
- [x] g1 cert load log shows `(2 certs in chain)` after fix
- [x] Staged deploy to g1/g2/g3/g4/g5/gateway/gateway-2
- [ ] Mobile rustls trust-store fix (separate PR in mobile-app repo)

### DIFF
--- a/.github/scripts/check-data-paths.sh
+++ b/.github/scripts/check-data-paths.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+#
+# Forbid CWD-relative data paths in Rust source.
+#
+# Every persisted-file path must be anchored at `zhtp::node_data_dir()`
+# (via `zhtp::node_data_path`) or a configured absolute path. Raw
+# `"./data/..."` or `PathBuf::from("./data/...")` literals only work
+# by accident of process CWD and break silently otherwise — see the
+# POUW budget save regression that produced this guard.
+#
+# Exclusions: tests (any */tests/** or *_tests.rs path), build output,
+# docs, and plans.
+
+set -euo pipefail
+
+hits=$(grep -RInE '"\./data/|PathBuf::from\("\./data/' \
+    zhtp/src lib-network/src lib-blockchain/src lib-consensus/src \
+  | grep -vE '/tests?/|_tests\.rs|/target/|/docs/|/plans/' \
+  || true)
+
+if [ -n "$hits" ]; then
+  echo "::error::Forbidden CWD-relative data path literal found."
+  echo "Use \`zhtp::node_data_path(\"data/...\")\` instead."
+  echo ""
+  echo "$hits"
+  exit 1
+fi
+
+echo "ok — no CWD-relative data path literals in source."
--- a/.github/scripts/check-data-paths.sh
+++ b/.github/scripts/check-data-paths.sh
@@ -13,7 +13,16 @@
 
 set -euo pipefail
 
-hits=$(grep -RInE '"\./data/|PathBuf::from\("\./data/' \
+# Catch every common CWD-relative form:
+#   "./data/..."                            (raw literal)
+#   PathBuf::from("./data/...")             (PathBuf ctor)
+#   Path::new("./data/...")                 (borrowed Path)
+#   std::path::Path::new("./data/...")      (fully-qualified)
+#   PathBuf::from_str("./data/...")         (Result-returning ctor)
+# The character class `[A-Za-z_:]*` matches optional path prefixes
+# (`std::path::`) on Path/PathBuf so qualified forms also trip the guard
+# (CR #2660).
+hits=$(grep -RInE '"\./data/|(PathBuf|Path)::(from|from_str|new)\("\./data/|[A-Za-z_:]+::Path::new\("\./data/' \
     zhtp/src lib-network/src lib-blockchain/src lib-consensus/src \
   | grep -vE '/tests?/|_tests\.rs|/target/|/docs/|/plans/' \
   || true)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,15 @@ jobs:
         run: bash .github/scripts/check-type-duplication.sh
         continue-on-error: true
 
+  data-path-check:
+    name: Forbid CWD-relative data paths
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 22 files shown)


umwelt (commented):
## Self-audit (since this PR was authored in-session, not blind-reviewed)

Honest pass on the same architecture / security / blockchain criteria applied to the other PRs.

## Scope is too big

21 files spanning 4 unrelated concerns:
1. `node_data_path()` helper + CWD-relative path migration (~10 files)
2. TLS chain serving fix + bootstrap-mode auto-pin + cert-rotation tolerance (3 files)
3. Pre-existing prod fixes from an earlier session's stash that hadn't been PR'd (contracts.rs replay tolerance, validation.rs TokenTransfer tightening, consensus_engine bootstrapping gate, components/consensus catch-up validator prefix, token/mod.rs BUBL sled-first, network/mod.rs state-field for synced observers)
4. CI guard script

In hindsight this should have been **5–6 small PRs**, not one. Each concern has different review fingerprints and bisecting a future regression to this commit is now harder. The bundling decision was driven by tonight's incident urgency, not architecture — that's a judgment call worth flagging as wrong in retrospect.

## Architectural concerns

### bootstrap-mode auto-pin window

The new `bootstrap()` config persists anchors and the verifier accepts SPKI rotations (re-pins automatically). Rationale is that the Dilithium UHP handshake remains the identity gate. That's true — **but** the bootstrap window between TLS-cert-presented and UHP-verified is a real attack surface:

- First connection: trustdb empty → bootstrap accepts cert blindly, persists it.
- A network attacker intercepting the first connection can poison the anchor.
- The Dilithium handshake will fail against the poisoned peer (attacker can't forge Dilithium signature for the chain DID), connection drops.
- But the poisoned anchor stays. Every subsequent legit connection now fails with `Certificate changed`.

This is a DoS vector, not a confidentiality compromise. Mitigation: when UHP handshake fails after a bootstrap anchor was just written in the same session, **rollback the anchor** rather than leaving it. Not implemented — should be.

### `#[cfg(test)] new()` shim

After deleting `QuicMeshProtocol::new()` from production, the `#[cfg(test)]` shim makes tests use `std::env::temp_dir().join(format!("zhtp-test-tls-{}", std::process::id()))`. This is per-PID so parallel test runs don't collide, but it doesn't clean up — repeated CI runs accumulate temp directories. Minor housekeeping issue, not blocking.

## Security: no tests added for the rotation-tolerant verifier

The new "auto re-pin under bootstrap_mode on SPKI mismatch" branch in `trust.rs::verify_server_cert` has no unit test. The behaviour difference between bootstrap and non-bootstrap rejection is subtle. A test asserting:
1. Non-bootstrap mode rejects with `Certificate changed!` (existing).
2. Bootstrap mode logs WARN and accepts (new).
3. After re-pin, subsequent connections succeed with the new SPKI (regression guard).

…would prevent this regressing later.

## Blockchain concerns

None introduced by this PR. The stashed prod fixes folded in (contracts.rs, validation.rs, consensus_engine/network.rs, components/consensus.rs) were already live-deployed and verified before this PR opened; the diff matches what's been running.

## CI guard

`check-data-paths.sh` is the right shape — keeps the canonical-path constraint enforced without depending on reviewer memory. Worth promoting from `continue-on-error: true` (it isn't `continue-on-error`, good) but watch for it becoming noisy on legitimate path patterns.

## Verdict

Merge OK given it's already deployed and reverting now would re-introduce the bugs the validators have been running without for hours. Track follow-ups:

1. Add unit tests for the rotation-tolerant verifier path.
2. Add the post-UHP-failure anchor-rollback for the DoS-vector mitigation.
3. Future bundles like this should be split — this one was emergency triage.

Issue #2638, opened 2026-05-23
arch: phase 2c — migrate token-nonce reads to facade


umwelt:
# Context

Depends on #2635. \`Blockchain.token_nonces\` HashMap diverges from sled in the same way as balances. When the rewards handler reads \`bc.get_token_nonce(...)\`, it gets the in-memory value, which after a replay-with-tolerance is 0 even though the chain's actual nonce is N. New reward txes use nonce=0 → fail apply at BlockExecutor.

# Scope

- \`bc.get_token_nonce\` already has sled fallback (\`blockchain.rs\` ~2521) — verify it's correct
- Callers that bypass it and read the HashMap directly: enumerate, migrate
- Audit the rewards handler's mint_to_user: confirm it uses the facade getter

# Acceptance

- [ ] All nonce reads routed through \`bc.token_nonce()\` (facade-style, sled-first)
- [ ] Divergence detector reports zero nonce-related divergences
- [ ] Reward TokenTransfer success rate at 100% during soak (chain commit, not just mempool acceptance)
- [ ] Replay-tolerance hack for nonce-mismatch removal deferred to Phase 4

# Blocks

- #2635

umwelt:
## Phase 2c finding: no read-migration code needed

Exhaustive grep across `zhtp` / `zhtp-cli` / `zhtp-daemon` / `tools`:

- **No external direct `.token_nonces` field reads exist.**
- All 3 external nonce readers already go through a facade:
  - `zhtp/src/api/handlers/token/mod.rs:641` → `blockchain.get_token_nonce(...)`
  - `zhtp/src/api/handlers/rewards/mod.rs:467` → `bc.get_token_nonce(...)`
  - `zhtp/src/api/handlers/observer_admission.rs:699` → `store.get_token_nonce(...)` (reads sled directly)

So the "migrate reads to facade" deliverable is **already satisfied** by prior work — no PR for the read migration.

## Remaining: sled-first flip is deferred to Phase 3 (#2640)

`Blockchain::get_token_nonce` is currently **HashMap-first** (sled fallback). Making it **sled-first** (to match `get_token_contract`) is the correct end state, but nonces gate transaction replay protection, so flipping the read order is only safe **after** the divergence detector (#2635) has proven in-mem == sled in soak. Folding that one-line flip into Phase 3, where sled becomes authoritative and the in-memory `token_nonces` cache is privatized/removed.

[...]

PR #2661, opened 2026-05-29
refactor(#2637): Phase 2b — migrate token_contracts reads to facade

### LINKED ISSUE #2637: arch: phase 2b — migrate token balance reads to Blockchain::token_balance()

umwelt:
Closes #2637 (token_contracts). Phase 2b of the #2645 state-unification migration.

> **Stacked PR.** Base is `phase-2a-blocks-migration` (PR #2659). Merge order: #2657 → #2658 → #2659 → this.

## What this does

Routes external `token_contracts` **reads** through sled-first facades so Phase 3 (#2640) can privatize the field. Adds one facade and **fixes three latent balance-read bugs** in passing.

## New facade

`Blockchain::iter_token_contracts() -> Vec<TokenContract>` — sled-first list, in-mem fallback when store-less. Canonical replacement for direct `.values()`/`.keys()`/`.iter()` reads. (`token_balance()` and `get_token_contract()` already existed.)

## Migrated

| Kind | Facade | Sites |
|---|---|---|
| balance | `token_balance()` | network, wallet (×2), web4/domains |
| metadata | `get_token_contract()` | token (name/symbol/decimals), blockchain (total_supply) |
| aggregate | `iter_token_contracts()` | token (list, find-by-symbol), oracle (×2 find-by-symbol), blockchain (contract list) |

### 🐛 Bug fixed: synthetic-key balances returned 0

Three balance reads built a synthetic `PublicKey { dilithium_pk: [0;2592], kyber_pk: [0;1568], key_id }` and called `balance_of()`. But `PublicKey`'s `Hash`/`Eq` compare **all** fields, so the synthetic key never matched the real entry in the balances map — **these reads silently returned 0**. `token_balance()` reads sled keyed by `key_id` (`Address`), so they now return the real balance. (One call site, `network/mod.rs`, had already worked around this with `find_balance_by_key_id` + a comment explaining the trap.) The web4/domains "SOV not initialized" error is preserved via a `get_token_contract().is_none()` guard.

## NOT migrated (intentional)

- **Mutations → Phase 3 / #2640**: `token_contracts.insert` / `get_mut` in the runtime SOV-init clusters, `genesis_funding`, and `web4/domains:1024` (a debit/credit that carries its own TODO admitting it diverges state and must move into block processing). The adjacent count/log reads that also touch `utxo_set` stay with them.
- **No-facade-yet**: `tools/chain_audit.rs` sums **all** balances of a token — needs a sled balance-iteration facade (sled keeps balances in a separate tree from the `TokenContract`), tracked as follow-up. Tool only.

## utxo (#2637+) — split to its own issue

`utxo_set` has **no facade**: in-mem `HashMap<Hash, TransactionOutput>` vs sled `OutPoint → Utxo` is a real type impedance (key shape *and* value type differ), deferred from Phase 1. Building it is genuine design, not a mechanical swap — so utxo migration gets its own issue rather than being rushed in here. **(Filing now; will link.)**

## Process note

The Phase 0 CSV **under-counted `token_contracts`** (~17 listed vs ~28 actual) because most accesses are multiline. Found by exhaustive per-file grep. Same lesson as 2a — grep per file, don't trust the CSV line list.

## Test plan

- [ ] CI green (builds clean: lib-blockchain + zhtp; 4 facade tests pass incl. new `iter_token_contracts_reads_sled`)
- [ ] With `ZHTP_DIVERGENCE_DETECT=1`, confirm token-balance reads agree with sled
- [ ] **Regression to verify the bug fix**: wallet/network balance endpoints now report real SOV balances where they previously may have shown 0

## Next in the stack

Phase 2c (#2638) — token nonce reads (small), then Phase 2d (#2639) validator+identity+wallet.

### DIFF
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -1032,6 +1032,22 @@ impl Blockchain {
             .unwrap_or(0)
     }
 
+    /// Sled-first iterator facade over all token contracts (#2637).
+    ///
+    /// Returns the authoritative set from the store when attached, falling back
+    /// to the in-memory map in store-less mode. Collected into a `Vec` so
+    /// callers get an owned, lifetime-free list to `.iter()` / `.len()` /
+    /// `.is_empty()` / `find()` over — the canonical replacement for direct
+    /// `token_contracts.values()` / `.keys()` / `.len()` reads.
+    pub fn iter_token_contracts(&self) -> Vec<crate::contracts::TokenContract> {
+        if let Some(store) = self.get_store() {
+            if let Ok(iter) = store.iter_token_contracts() {
+                return iter.map(|(_, c)| c).collect();
+            }
+        }
+        self.token_contracts.values().cloned().collect()
+    }
+
     pub fn register_web4_contract(
         &mut self,
         contract_id: [u8; 32],
--- a/lib-blockchain/src/blockchain/tests/facade_tests.rs
+++ b/lib-blockchain/src/blockchain/tests/facade_tests.rs
@@ -32,6 +32,27 @@ fn token_balance_reads_sled_when_store_attached() {
 }
 
 #[test]
+fn iter_token_contracts_reads_sled() {
+    let temp = tempfile::tempdir().unwrap();
+    let store = Arc::new(SledStore::open(&temp.path().join("facade_iter_store")).unwrap());
+
+    let sov = crate::contracts::TokenContract::new_sov_native();
+    let sov_symbol = sov.symbol.clone();
+    store.begin_block(0).unwrap();
+    store.put_token_contract(&sov).unwrap();
+    store.commit_block().unwrap();
+
+    let mut bc = Blockchain::new().expect("blockchain construct");
+    bc.set_store(store);
+
+    let contracts = bc.iter_token_contracts();
+    assert!(
+        contracts.iter().any(|c| c.symbol == sov_symbol),
+        "iter_token_contracts must surface the sled-stored SOV contract"
+    );
+}
+
+#[test]
 fn identity_consensus_by_did_none_without_store() {
     let bc = Blockchain::new().expect("blockchain construct");
     assert!(bc.get_store().is_none(), "test assumes no store attached");
--- a/zhtp/src/api/handlers/blockchain/mod.rs
+++ b/zhtp/src/api/handlers/blockchain/mod.rs
@@ -1188,9 +1188,9 @@ impl BlockchainHandler {
         };
 
         let sov_token_id = lib_blockchain::contracts::utils::generate_lib_token_id();
+        // #2637: get_token_contract() is the sled-first metadata facade.
         let total_supply = blockchain
-            .token_contracts
-            .get(&sov_token_id)
+            .get_token_contract(&sov_token_id)
             .map(|token| token.total_supply)
             .unwrap_or(0);
 
@@ -2445,14 +2445,16 @@ impl BlockchainHandler {
 
         let mut contracts = Vec::new();
         if contract_filter == "all" || contract_filter == "token" {
+            // #2637: iter_token_contracts() is the sled-first list facade.
+            // (contract_blocks is a separate metadata index, migrated later.)
             contracts.extend(
                 blockchain
-                    .token_contracts
-                    .keys()
-                    .map(|id| ContractListItem {
-                        contract_id: hex::encode(id),
+                    .iter_token_contracts()
+                    .into_iter()
+                    .map(|t| ContractListItem {
+                        contract_id: hex::encode(t.token_id),
                         contract_kind: "token".to_string(),
-                        block_height: blockchain.contract_blocks.get(id).copied(),
+                        block_height: blockchain.contract_blocks.get(&t.token_id).copied(),
                     }),
             );
         }
--- a/zhtp/src/api/handlers/network/mod.rs
+++ b/zhtp/src/api/handlers/network/mod.rs
@@ -925,13 +925,11 @@ impl NetworkHandler {
                         // because PublicKey's Hash/PartialEq compare all fields (dilithium_pk,
                         // kyber_pk, key_id), so a synthetic PublicKey with zeroed crypto keys
                         // would never match a real entry in the balances HashMap.
+                        // #2637: token_balance() is sled-first and keyed by
+                        // key_id (the find_balance_by_key_id dance is now inside
+                        // the facade).
                         let wallet_key_id = w.wallet_id.as_array();
-                        let balance = blockchain
-                            .token_contracts
-                            .get(&sov_token_id)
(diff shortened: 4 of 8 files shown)


umwelt (commented):
## Review verdict: COMMENT (with follow-ups before approval)

This PR correctly narrows the **API staleness bug** for the 6 handler endpoints and introduces no new consensus-critical risk. The intent is sound. Two architectural items warrant explicit follow-up before approval, and a third predates the PR but gets worse because of it.

## Architectural footgun: `iter_token_contracts()` returns contracts with EMPTY balances

The new `Blockchain::iter_token_contracts()` is documented as "sled-first with in-memory fallback." That's correct for **metadata** (`name`, `symbol`, `total_supply`, `decimals`, authorities). But:

`TokenContract` in sled stores the contract struct, which has an in-memory `balances: HashMap<PublicKey, Amount>` field. Per-address balances are written to a **separate** `token_balances` sled tree keyed by `token_balance_key`. When `TokenContract` is deserialized from the `token_contracts` sled tree, **its `balances` HashMap will be empty** unless the executor re-serialized the whole struct (it doesn't — it writes individual `(token_id, address)` balance entries to the other tree).

So: `iter_token_contracts()` returns contracts where `.balance_of(addr)` ALWAYS returns 0.

The PR handlers correctly avoid this trap (they use `token_balance()` separately). But a future call site that does `let contracts = bc.iter_token_contracts(); for c in contracts { c.balance_of(addr) }` will silently return zero balances for everyone. This is a landmine.

**Fix:** either rename to `iter_token_contract_metadata()` or add a doc comment with `#[doc(panic_when_misused)]`-style language that makes the trap explicit. The risk of someone using this for balance reads in 3 months is high.

## Pre-existing drift hotspot now more visible

`zhtp/src/api/handlers/web4/domains.rs:1023-1044`

The domain handler does an out-of-band balance mutation: `token_contracts.get_mut(...)` → `token.balance_of(...)` → `token.set_balance(...)`. This mutates the **in-memory HashMap** directly while the sled `token_balances` tree is **not updated**. After this PR, `query_token_balance` is sled-first — so the sender appears to still have full balance after the deduction. **The domain fee is effectively free if the domain handler is called again before a block commits.**

This isn't introduced by the PR but it becomes a visible drift source now that other reads are sled-authoritative. Should not be left to accumulate through Phase 3.

**Fix:** track as a separate issue at minimum. Ideally migrate `domains.rs` to use the normal TokenTransfer apply path in Phase 3, so the balance mutation goes through `process_token_transactions` like everything else.

## API-only fix — consensus path still on old reads

The scope is API-only and that's stated. But for clarity:

- `deduct_transaction_fees()` (`blockchain.rs:1884`) — `self.token_contracts.get_mut(&sov_token_id)`, pure in-memory.
- `process_token_transactions()` (`blockchain/contracts.rs:327-331`) — `self.token_contracts.get_mut(&token_id)` + `token.balance_of(...)`, pure in-memory.

These stay on the in-memory HashMap, by design, deferred to Phase 3. The post-block `sync_in_memory_from_sled` (`blockchain.rs:1096`) is what keeps them in sync. Worth a comment in the PR body noting Phase 3 takes responsibility for migrating these — otherwise a reviewer in 2 months won't remember the split was intentional.

## Wallet handler synthetic-key fix is genuine

`zhtp/src/api/handlers/wallet/mod.rs:376-401, 531-571` — replacing `token.balance_of(&synthetic_pk)` with `token_balance()` keyed by address is correct, and it fixes a latent bug where the synthetic key (zeroed dilithium/kyber, real key_id) never matched the in-memory `balances` HashMap. Pre-PR, wallet detail/list endpoints **always returned 0**. After PR, they return the real sled balance. Good fix.

**Caveat:** any balance written exclusively to in-memory before executor was wired (early-mint paths, pre-block tests) that was never flushed to sled now returns 0 instead of the wrong-but-nonzero in-memory value. Verify the backfill at `blockchain.rs:1010-1047` runs **before handler reads**, not just before block application. If not, expect bug reports from users with very old wallets seeing balances drop to 0.

## What to do (none blocking)

1. Rename `iter_token_contracts()` to `iter_token_contract_metadata()` or add explicit-trap docs.
2. File a follow-up issue for `domains.rs:1023-1044` out-of-band mutation; do not let it slide into Phase 3.
3. Add a sentence to the PR body explicitly listing the consensus-path read sites that stay on the old path so future reviewers understand the scope boundary.
4. Verify the backfill ordering vs. handler-read ordering on cold-boot.

umwelt:
Closing — **redoing Phase 2 properly** (same reason as #2659).

The token_contracts read migration was built on the external-only scope and the under-counted inventory. The non-blocking findings here (the `iter_token_contracts()` empty-balances footgun, the `domains.rs` out-of-band mutation) plus the consensus-path sites it didn't touch will be handled in the field-atomic `token_contracts` redo, where internal + external sites migrate together and the footgun facade is renamed/guarded before any caller uses it.

Branch `phase-2b-token-utxo-migration` preserved for salvage (the synthetic-key balance bug fix and the handler swaps are reusable). Foundation (#2657 inventory, #2658 facade+detector) stays.

PR #2670, opened 2026-05-30
Fixed: Verify expected proof against submitted proof

officialredroleplays-ui:
Fixes missing proof verification in storage access control.

**Changes:**
- `lib-storage/src/dht/storage.rs:912`: Removes `_` underscore from
  `_expected_proof` so the expected proof is actually used
- `lib-storage/src/dht/storage.rs:919-928`: Adds verification block
  that compares `zk_proof.backend_proof.data` against the expected
  proof generated with the correct `requester_secret`

**Impact:** Previously, any valid ZK proof (even with wrong parameters)
could pass verification. Now the proof must match the expected one.

This PR contains only this fix. No other changes.

Dev: Harley Ross


officialredroleplays-ui:
Im not sure how to get rid of the other commits just take the code i fixed please

officialredroleplays-ui:
@officialredroleplays-ui
fix(storage): verify expected proof against submitted proof thats the one

Issue #59, opened 2025-12-01
[ALPHA-STEP2] Update ZhtpMeshServer to use lib_identity::NodeId


umwelt:
## Parent Milestone
Part of: #50 (Step 2: Network Integration)

## Goal
Update mesh server implementation to use deterministic `lib_identity::NodeId` instead of random NodeIds.

## Why
Ensures mesh network connections use stable, identity-based NodeIds for peer identification.

## Dependencies
- Requires: #49 (Step 1: DHT Integration) complete

## Tasks
- [x] Update ZhtpMeshServer struct to use lib_identity::NodeId
- [x] Replace any Hash-based NodeId with lib_identity::NodeId
- [x] Update peer connection tracking
- [x] Update server initialization to accept/derive NodeId from identity
- [x] Add tests for NodeId usage in mesh server

## Files to Modify
- `lib-network/src/mesh/server.rs` - Mesh server implementation
- Related mesh networking files

## Acceptance Criteria
- [x] ZhtpMeshServer uses `lib_identity::NodeId`
- [x] No random NodeId generation in mesh layer
- [x] Tests verify mesh server uses deterministic NodeIds
- [x] `cargo build --package lib-network` succeeds
- [x] `cargo test --package lib-network` passes

## Priority
**P0 - CRITICAL** - Required for Step 2 completion

Issue #58, opened 2025-12-01
[ALPHA-STEP1] Add integration tests for NodeId stability across restarts


umwelt:
## Parent Milestone
Part of: #49 (Step 1: DHT Integration)

## Goal
Create integration tests that verify the same identity produces the same NodeId across multiple DHT node restarts.

## Why
Critical acceptance criterion: "Same identity → same NodeId across restarts"

## Dependencies
- Requires: #54, #55, #56, #57 (All Step 1 implementation tasks)

## Tasks
- [ ] Create test that initializes DhtNode with fixed identity
- [ ] Record NodeId value
- [ ] Simulate node restart (drop and recreate with same identity)
- [ ] Verify NodeId matches original value
- [ ] Test with multiple devices (different NodeIds for same DID)
- [ ] Test DHT routing table rebuild with same NodeIds

## Test Scenarios
1. Single node restart - NodeId unchanged
2. Multiple restarts - NodeId always identical
3. Multi-device - Same DID, different devices, different NodeIds
4. DHT routing persistence - Routing table rebuilds with same NodeIds

## Files to Create/Modify
- `lib-storage/tests/dht_nodeid_stability.rs` - New integration test file

## Acceptance Criteria
- [ ] Test verifies same identity → same NodeId (10 restarts)
- [ ] Test verifies different devices → different NodeIds
- [ ] Test verifies DHT can locate nodes by stable NodeId
- [ ] All tests pass
- [ ] Code coverage includes restart scenarios

## Priority
**P0 - CRITICAL** - Final validation for Step 1

PR #2674, opened 2026-06-02
fix(#2639): identity_registry field — sled-first read migration (field-atomic)

umwelt:
## Phase 2d (#2639) — `identity_registry` field migration

Field-atomic migration of `identity_registry` reads to the sled-first facade, part of the state-unification epic (#2645). Follows the same discipline as the merged `blocks` (#2666) and `token_contracts` (#2668) fields.

### Why identity (and not validator) is this field
The tracker had `validator_registry` next, but it has **no sled backing** — there is no `validators` tree and `ValidatorInfo` is never serialized to any store. A sled-first facade over it would read empty and report **zero validators** (consensus break). It cannot join the read migration until validator state is *persisted* to sled (write-path change). Split off as a deferred task.

`identity_registry` is the correct next field: fully sled-backed (`identities`/`identity_metadata`/`identity_by_owner` trees, written by the executor's `register_identity`), and it contains the BLOCKING gateway-auth site the reviewer flagged.

### Root cause this fixes
There is **no sled→in-memory rebuild** of `identity_registry` on startup. A store-backed node that synced blocks into sled has an **empty** in-memory registry after restart, so every bare `identity_registry` read there is silently wrong — the source of post-restart gateway-auth failure, peer-admission rejection, and zeroed identity counts.

### What changed
**Store layer** (`BlockchainStore` + `SledStore`)
- `iter_identities()` / `count_identities()` — **required** methods (no default). A default-empty impl would silently report zero identities, so every backend must scan.

**Facade** (`Blockchain`)
- `identity_exists()` — **union** (in-mem OR sled). In-mem first preserves pending/same-block registrations (sled reads the committed tree, not the open tx_batch); sled second recovers restart-durable ones. Strict superset of the old `contains_key` — no consensus regression.
- `identity_count()` — sled-authoritative count.
- `iter_identities_consensus()` — owned sled `IdentityConsensus` set.
- `identity_public_key()` — full Dilithium key from metadata, **consensus-pinned** (`blake3(pk) == public_key_hash`) so the gateway-auth path trusts durable sled state while catching metadata↔consensus drift. Union fallback for the pending window.
- `identity_display_name()` — metadata display name, sled-first with in-mem fallback.

**Read-site migration** — internal (credentials/gateway existence gates, query API) and external (gateway-auth verifier + key, peer-admission verifier, `created_at`, display_name, owner pubkeys, identity counts).

**Divergence detector** — added the sled-side identity reverse-scan that was explicitly deferred pending a sled iterator ("needs sled iterators the trait does not yet expose"), now enabled by `iter_identities`. Catches the dominant post-restart direction (in-mem empty, sled full) the in-mem→sled scan was blind to.

### Deliberately NOT migrated (documented in code)
- Genesis-construction reads (sled empty pre-commit), snapshot import/diff + divergence detector (compare both sides by design), legacy writers (Phase 4).

### Deferred with tracked follow-ups (write-path / foundation gaps)
- **kyber_public_key is not persisted to sled** → messaging session-setup + DID-resolution reverse-lookups can't migrate until it's added to `IdentityMetadata`. (#58)
- **Proposer / username / `controlled_nodes` scans** need a metadata iterator; the proposer path is consensus-liveness-critical and deserves a focused, separately-tested change. (#59)

### Verification
- `cargo build --workspace` — clean (exit 0).
- `cargo test -p lib-blockchain --lib` — **1886 passed, 0 failed**.
- New tests: facade union/count/iter/consensus-pin/display_name; detector `detects_sled_only_identity`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,6 @@
+# Heavy end-to-end tests (e.g. zhtp's identity-migration tests that build a full
+# UnifiedStorageSystem) exceed the default 2MB stack of libtest / current-thread
+# tokio test threads and abort with a stack overflow. Bump the per-thread stack
+# for all cargo-invoked processes. Harmless in production (larger stacks only).
+[env]
+RUST_MIN_STACK = "16777216"
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -4915,9 +4915,12 @@ impl Blockchain {
     /// Create chain summary for local blockchain
     async fn create_local_chain_summary_async(&self) -> crate::ChainSummary {
         // Use the data helix root as the genesis content commitment.
+        // #2636: get_block(0) resolves genesis across the hot window AND sled.
+        // self.blocks.first() returned the oldest in-memory WINDOW block, which is
+        // NOT genesis once genesis ages out on a store-backed node — advertising a
+        // wrong genesis hash to peers.
         let genesis_hash = self
-            .blocks
-            .first()
+            .get_block(0)
             .map(|b| hex::encode(b.header.data_helix_root))
             .unwrap_or_else(|| "none".to_string());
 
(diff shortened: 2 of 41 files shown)


umwelt:
Review findings from local inspection:

1. zhtp/src/api/handlers/credentials/mod.rs still verifies DID ownership through blockchain.query_identity(did), while lib-blockchain/src/query_impl.rs keeps query_identity as an in-memory-only lookup. After a store-backed restart, query_identity_exists can correctly find the DID from sled but this path returns DID not found, rejecting credential ownership checks for durable identities. Since the handler only needs the Dilithium public key, use the new identity_public_key(did) facade here.

2. lib-blockchain/src/blockchain/dao.rs computes verified_citizen_count from identity_registry.len(), and that snapshot gates DAO phase transitions. On a restarted store-backed node the in-memory identity shadow can be empty, so the node undercounts citizens and can suppress phase advancement. This should use identity_count(); it does not need the deferred metadata iterator.

I attempted to post these as a request-changes review, but GitHub rejects request-changes reviews on PRs authored by the same account.

PR #2676, opened 2026-06-02
fix(#2639): proposer/validator controlled-nodes read from sled metadata (consensus liveness)

umwelt:
## Consensus-liveness fix — restarted validators stop producing blocks

Follow-up to the identity_registry sled-first migration (#2674). On a store-backed node the in-memory `identity_registry` is **empty after restart** (no sled→in-mem rebuild). The live mining loop decided "is this node the selected proposer?" by **scanning that shadow** for the node's `controlled_nodes` — so after a restart a validator found nothing, `is_proposer` stayed `false`, and it **silently stopped producing blocks.**

### Changes
- **New facade** `Blockchain::identity_controlled_nodes(did) -> Option<Vec<String>>` — reads `controlled_nodes` from the durable `identity_metadata` tree (the executor persists it on registration **and** identity update), with in-mem fallback for store-less/pending. `None` for an unknown DID; `Some(possibly-empty)` otherwise.
- **`mining_service` proposer check** — reframed the reverse-scan into a **point lookup on the selected proposer's** `controlled_nodes`: this node is the proposer iff the proposer controls it. No full-registry scan, sled-authoritative.
- **`consensus::controlled_nodes_for_validator_id`** — same point-lookup reframe (the validator id *is* the did_hash), instead of scanning `identity_registry`.
- **Deleted `lib-blockchain/src/mining.rs`** — `should_mine_block` / `check_if_proposer` / `IdentityData` were **entirely dead** (zero callers anywhere) and duplicated this exact proposer logic, carrying the same restart bug. (The unrelated `types::mining::get_mining_config_from_env` is a different module, untouched.)

### Why point lookups (not a full metadata iterator)
The validator/proposer DID is always known at the call site, so `get_identity_metadata` (already in the store) suffices — cleaner and avoids O(n) scans. A full `iter_identity_metadata` is only needed for the genuinely scan-shaped reverse-lookups (`get_identity_by_public_key`, wallet key-scan, `bridge_node_count`, `list_all_identities`), which are lower-severity and remain as follow-up.

### Verification
- New test `identity_controlled_nodes_reads_sled_metadata` (sled-only / restart case)
- `lib-blockchain` **1887/1887**, both crates build clean


umwelt:
Reviewed PR #2676. I did not find blocking issues.

Checked:
- identity_controlled_nodes() reads controlled_nodes from sled metadata with in-memory fallback for store-less/pending identities.
- mining_service now does a point lookup on the selected proposer's DID and checks whether it controls the local node, preserving the old controlled_nodes semantics without scanning the empty post-restart shadow.
- consensus controlled_nodes_for_validator_id reconstructs the same did:zhtp:<validator-id-hex> shape that the old DID-suffix comparison used.
- lib-blockchain/src/mining.rs appears unreferenced outside itself, so deleting the stale duplicate proposer logic is fine.

Residual note: the new facade is covered for the sled-only metadata case; a higher-level mining/consensus liveness regression test would be useful, but I would not block this fix on it.

Posting as a comment because GitHub rejects approvals/reviews on PRs authored by the same account.

umwelt:
Addressed Copilot CR in ece778b7. The `.ok().flatten()` silent error-swallow was the same class of bug this PR exists to fix (sled error → looks like "unknown DID" → restarted validator silently stops producing blocks).

Three states now handled explicitly:

| sled result | action |
|---|---|
| `Ok(Some(meta))` | return `controlled_nodes` (success) |
| `Ok(None)` | fall through to in-memory shadow (legitimate pending-pre-commit / store-less harness) |
| `Err(e)` | `warn!` and return `None` — caller sees the absence, operator sees the actual error in logs |

Falling-through-on-error was the original bug pattern; falling-through-only-on-genuine-absence preserves the legitimate "not yet committed to sled" path. Test `identity_controlled_nodes_reads_sled_metadata` still passes.

Issue #61, opened 2025-12-01
[ALPHA-STEP2] Verify NodeId during peer handshake


umwelt:
## Parent Milestone
Part of: #50 (Step 2: Network Integration)

## Goal
Add NodeId verification during peer handshake to ensure claimed identity matches derived NodeId.

## Why
Security: Prevents peers from claiming false identities. Validates that NodeId = Blake3(DID + device).

## Dependencies
- Requires: #60 (Peer discovery update)

## Tasks
- [ ] Update handshake protocol to include DID + device
- [ ] Verify received NodeId matches Blake3(DID + device)
- [ ] Reject connections with mismatched identity/NodeId pairs
- [ ] Add error handling for verification failures
- [ ] Log verification failures for security monitoring
- [ ] Add tests for handshake verification

## Files to Modify
- `lib-network/src/mesh/handshake.rs` - Handshake implementation
- Related connection establishment files

## Acceptance Criteria
- [ ] Handshake verifies NodeId against identity
- [ ] Mismatched pairs rejected
- [ ] Tests verify rejection of invalid NodeIds
- [ ] Tests verify acceptance of valid NodeIds
- [ ] `cargo test --package lib-network` passes

## Priority
**P0 - CRITICAL** - Security requirement for Step 2

PR #2678, opened 2026-06-03
fix(#2639): iter_identity_metadata + migrate pubkey reverse-lookups to sled

umwelt:
## Metadata iterator + the scan-shaped identity reverse-lookups (#61 follow-up to #2674/#2676)

These read the in-memory `identity_registry` via **full scans**, so they're empty/stale on a store-backed node after restart. **Two are consensus paths.**

### Changes
- **Store** `iter_identity_metadata()` — required (no default; a silent-empty default would drop the whole set for a consensus-feeding scan), `SledStore` impl validates each record's DID hashes back to its tree key, + `MockObserverStore` stub.
- **Facade** `Blockchain::did_by_public_key(pk) -> Option<String>` — resolve a DID by Dilithium public key via sled metadata (in-mem fallback). Revoked identities are excluded automatically (revocation deletes the metadata record, so the old `!= "revoked"` filter is implicit).
- **`transaction/validation.rs` council checks (CONSENSUS)** — threshold-approval validation used `get_identity_by_public_key` → now `did_by_public_key`. Restart-stale before: council members were unverifiable → approvals silently failed.
- **`wallets.rs::resolve_public_key_by_key_id` (CONSENSUS)** — called from `TokenTransfer`/`TokenMint` execution to resolve a recipient's pubkey by key_id; the identity scan now reads sled metadata first, so it resolves after restart. (The sibling `wallet_registry` scan is the separate wallet-field migration.)
- **`is_public_key_registered`** (registration dedup) — routed through `did_by_public_key`.
- **Deleted dead code:** `get_identity_by_public_key` and `Blockchain::list_all_identities` (zero callers — the only `list_all_identities` caller is the `IdentityManager`'s).

### Not migrated (documented)
`ChainSummary.bridge_node_count` filters `identity_type.contains("bridge")` — a string that exists in **neither** sled `IdentityMetadata` **nor** the `IdentityConsensus` enum (variants: Human/Org/Device/Agent). Needs an identity-type model change first; out of scope here.

### Why point/scan choices
The validator/proposer/key_id is always known at the call site for point lookups; the genuine reverse-lookups (DID-by-pubkey, pubkey-by-key_id) need the iterator. Both are O(n) just like the in-memory scans they replace.

### Verification
- New test `did_by_public_key_reads_sled_metadata` (sled-only/restart)
- `lib-blockchain` **1887/1887**, `token_regression` **24/24** (the consensus token-exec path), zhtp builds + observer **18/18**


umwelt:
Addressed both CR items in b86ffbdf + resolved merge conflict with development (#2676's `identity_controlled_nodes_reads_sled_metadata` test preserved alongside the new one).

| CR | Resolution |
|---|---|
| `identity.rs:180` — `did_by_public_key` trusts metadata without consensus pin | Now: after a metadata `public_key` match, load `IdentityConsensus`, require `blake3(metadata.pk) == consensus.public_key_hash` AND `consensus.status != Revoked`. Metadata without a consensus record is also refused (drift signal). Mirrors `identity_public_key`'s pinning pattern. |
| `wallets.rs:391` — silent `if let Ok(iter)` on consensus path | Now: `match` with `warn!` on `Err`, falling through to in-memory shadow. Same fix applied to the analogous `did_by_public_key` iter call. |

New negative-path assertion in `did_by_public_key_reads_sled_metadata`: a metadata key whose hash doesn't match its consensus `public_key_hash` is refused. Tests `lib-blockchain --lib`: 1887/1887 pass.

PR #2679, opened 2026-06-03
feat(#2639): persist kyber_public_key to sled via schema-v2 metadata migration

umwelt:
## What

Persists identity **Kyber (KEM) public keys** into sled's `identity_metadata` tree. Until now kyber lived only in the in-memory identity registry and in block transactions — never durably in sled — so DID-resolution / messaging kyber reads depended on the non-durable in-memory shadow, the exact divergence the state-unification epic (#2645) removes.

## Why the careful approach

`identity_metadata` is **bincode**-encoded: positional and non-self-describing. A `#[serde(default)]` field is **not** safe here — a missing trailing field corrupts the decode rather than defaulting. So we don't read any old-shaped blob. Instead we version the tree and regenerate it from blocks.

## Changes

- **Field**: `kyber_public_key: Vec<u8>` on `IdentityMetadata`, populated at every construction site — registration **and** update (kyber rotation) in `process_identity_transactions`, `new`/`Default`, and `convert_legacy_identity` (sources it from the identity tx).
- **Versioning**: `IDENTITY_METADATA_SCHEMA_VERSION = 2` + a meta key with sled get/set; absent key reads as v1 (pre-kyber).
- **Migration** (`migrate_identity_metadata_schema`), wired into `replay_from_store` after the in-memory registry is rebuilt from blocks: if the stored version is behind, **clear** the tree and **rewrite** every record in the new shape from the replayed registry (which carries the real kyber from blocks), then bump the version. Idempotent via the gate; a mid-way failure leaves the version behind for a clean retry. **No old-shaped blob is ever read.**
- **Facade reader**: `Blockchain::identity_kyber_public_key(did)` — sled-first, in-memory fallback — so the persisted value is reachable.

## Tests

- `migrate_identity_metadata_schema_backfills_kyber_and_bumps_version` — pre-v2 record (empty kyber) + populated registry → migration backfills kyber, bumps version, facade reads it back, second run is a no-op.
- `migrate_identity_metadata_schema_skips_when_current` — store already at current version is left untouched (gate short-circuits before clear).
- Full `lib-blockchain --lib` suite green (1888 passed); `zhtp` builds.

## Follow-up (gated on #2678)

Migrate the messaging / device-key reverse-lookup reads off the in-memory registry onto the sled-authoritative facade. The reverse lookups need `iter_identity_metadata` (in unmerged #2678); the forward primitive (`identity_kyber_public_key`) ships here. Those reads are correct today (post-replay the in-memory registry holds kyber), so this is a consistency follow-on, not a live-bug fix.


umwelt:
Addressed all four CR items + merge conflict with development in f348d72a.

| CR | Resolution |
|---|---|
| **#1 replay.rs:232** — partial migration + version bump leaves permanently broken state | Migration now collects records, writes via bulk API (CR #3), checks `written == expected`. On any failure: return WITHOUT bumping version → next boot retries the full migration. |
| **#2 replay.rs:336** — test labeled "pre-v2" actually writes current struct | Updated comment to be honest. Added new test `migrate_identity_metadata_schema_does_not_decode_legacy_blobs` that plants raw `b"NOT_A_VALID_METADATA_BLOB__"` bytes (via new test-only `put_identity_metadata_raw_bytes` helper) and verifies the migration completes without panicking, replaces the junk, bumps the version. The safety holds because `clear_identity_metadata` uses `Tree::clear()` (bulk delete without per-entry decode). |
| **#3 replay.rs:221** — O(n) per-record flush | Added `BlockchainStore::put_identity_metadata_batch(records) -> StorageResult<usize>` with `SledStore` override that inserts all then flushes once. Migration uses it. |
| **#4 sled_store.rs:1876** — brand-new v2 store reads as v1 | `from_db` now checks `meta version absent && identity_metadata empty` → marks at current version. Legitimate legacy (absent key, non-empty tree) still reads as v1 and migrates. Existing `backfills_kyber` test now explicitly sets v1 since brand-new init writes current. |

Also: merge conflict in `identity.rs` (overlap between this PR's `identity_kyber_public_key` and #2676's `identity_controlled_nodes`) resolved by keeping both, and applying the error-handling pattern reviewers locked in on #2676/#2678 to the new kyber method (sled errors → `warn!` + `None` instead of silent fallthrough to post-restart-empty shadow).

Tests: `lib-blockchain --lib`: 1890/1890. Build: green.

PR #2683, opened 2026-06-03
fix(CONS-512): engine owns consensus height; storage is downstream

umwelt:
## What

Collapses the engine ↔ blockchain height-counter duplication that wedged g2 / g3 at H=123008 for 11 hours on 2026-06-03.

**Architectural change.** Pre-CONS-512 the engine derived its consensus height from `blockchain.height` via `sync_height_with_blockchain()`. That made storage the source of truth for engine progress, and any missed sync trigger caused silent drift. CONS-512 reverses the direction: the engine OWNS its height, advances synchronously the moment 2f+1 commit votes are observed, and treats storage as a downstream consumer. `sync_height_with_blockchain()` becomes a boot-only contract.

See `docs/epics/cons-512-engine-owns-height.md` for the full design, rationale, and risk analysis.

## Why this and not a smaller patch

The bug shape was: \"add one more sync trigger.\" We resisted that. The pre-CONS-512 design had four sync triggers and we hit a code path that bypassed all of them. Adding a fifth would just defer the next occurrence — same shape applies to the next state-machine refactor that touches the commit path. PRs #2615, #2620, and this one are all variations of \"implicit coordination breaks under recovery.\" The structural fix removes the coordination assumption: engine height is causally consequent to the consensus decision, not derived from a downstream observer.

## Forensic context — the 2026-06-03 stall

Production was at H=123008 across all five validators. g1+g5 engines were at \`current_round.height = 123009\` (correctly proposing the next block). g2+g3 engines were at \`current_round.height = 123008\` (still trying to commit a block their blockchain layer already had committed). Two-vs-two split, neither side could reach 4/5 quorum, chain wedged.

The smoking-gun log:
\`\`\`
peer 3218b9025f votes at height 123009 (our consensus height 123008; local blockchain ~123007)
\`\`\`
The \`local blockchain ~123007\` is misleading — it's computed as \`current_round.height - 1\`, not read from storage. The actual blockchain.height was 123008. So the engine had \`current_round.height = blockchain.height\`, which is one less than it should be (engine vote-target = blockchain + 1).

Recovery required sequential restart of g2 and g3 so the boot path's \`sync_height_with_blockchain\` could re-pin them to 123009. Restart isn't a fix — the bug recurs on any commit if the engine leaves Commit step before its timer fires.

## What changes

| File | Change |
|---|---|
| \`lib-consensus/src/engines/consensus_engine/state_machine.rs\` | \`process_committed_block\` advances height synchronously after \`apply_block_to_state_with_proof\` returns. Removed redundant sync call from \`on_round_timeout(Commit)\`. Added CONS-512 contract comment to the vestigial NewBlock handler. |
| \`lib-consensus/src/engines/consensus_engine/mod.rs\` | Doc on \`sync_height_with_blockchain\` tightened to boot-only contract. |
| \`lib-consensus/src/engines/consensus_engine/network.rs\` | Removed sync call from Bootstrap→BFT mode transition. |
| \`lib-consensus/src/engines/consensus_engine/tests.rs\` | 4 existing tests updated to assert the new contract; 1 new regression test added. |
| \`docs/epics/cons-512-engine-owns-height.md\` | Design doc. |

## What does NOT change

- \`spawn_commit_executor\` in \`lib-consensus-runtime\` — the async-write design is correct, only the engine's height tracking needed to stop deriving from storage.
- Storage write semantics — failures still surface as \`Event::HaltScheduled\`.
- Boot path — \`blockchain.height\` from sled remains authoritative at startup. Catch-up sync still fills any gap from peers before BFT engages.
- Wire format, network protocol, schema, validator identity — all unchanged.

## Risk

Medium. This is a consensus correctness change. The design doc enumerates the failure modes that were considered (engine advances + storage fails + restart; engine advances + storage succeeds + restart; same-height fork after restart unaffected; storage falling arbitrarily behind backstopped by the watchdog).

## Test plan

- [x] \`cargo test -p lib-consensus --lib -- consensus_engine\` → **66/66 pass**, including the new \`test_cons512_commit_advances_height_synchronously\` regression
- [x] \`cargo check --workspace\` clean
- [x] \`cargo test --workspace --lib --no-run\` builds clean (pre-existing E0308 in \`lib-protocols/src/types/request.rs:481\` is unrelated to this branch — exists on dev)
- [ ] Staged rollout per HALT-BEFORE-DEPLOY protocol after merge: g1 canary → verify ≥100 blocks produced normally → g2 → g3 → g4 → g5 → gateways
- [ ] Monitor for the absence of \"our consensus height N; local blockchain ~N-1\" lines after rollout — that log message should never fire again with this contract in place

## Related

- The 2026-06-03 stall (no issue — diagnosed live, see commit body)
- PR #2620 (BFT quorum proof determinism — orthogonal post-restart fork fix)
- PR #2615 (mesh UNI receive size limit — same class of \"silent coordination break\")
- CONS-307 (the async fire-and-forget commit channel that this PR clarifies the contract around — left structurally intact)

### DIFF
--- a/docs/epics/cons-512-engine-owns-height.md
+++ b/docs/epics/cons-512-engine-owns-height.md
@@ -2,7 +2,7 @@
 
 **Status.** Draft (2026-06-03). Drafted in response to the 2026-06-03 stall at H=123008/123009 where two of five validators wedged because their engine's `current_round.height` and `blockchain.height` drifted by one.
 
-**Branch.** `fix/synchronous-commit-path-cons508`
+**Branch.** `fix/synchronous-commit-path-cons508` (branch name predates the CONS-512 ticket assignment — the synchronous-commit work was originally scoped under CONS-508's "delete legacy lib-consensus" umbrella before being scoped down to this surgical change. Kept as-is to preserve PR continuity; future references should use CONS-512.)
 
 **Tracks.** CONS-307, CONS-504, CONS-505. Supersedes the "wire a `ConsensusEvent::NewBlock` event" patch sketch.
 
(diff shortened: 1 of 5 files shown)


umwelt:
Review finding:

P1: lib-consensus/src/engines/consensus_engine/state_machine.rs advances the engine height after apply_block_to_state_with_proof() returns, but the commit_tx path inside apply_block_to_state_with_proof() still returns Ok(()) when tx.send(envelope) fails because the runtime commit executor receiver is gone. That means a finalized block can fail to reach persistence, no HaltScheduled event is emitted, and CONS-512 will still advance current_round.height at lines 1279-1281. The node can then keep voting/proposing H+1 while storage never received H, which is exactly the new contract's dangerous case: storage is downstream, but dispatch-to-storage failure must not be treated as successful downstream handoff.

Please make the closed commit channel path a hard error / halt-scheduled equivalent before the synchronous height advance, and add a regression test for closed commit_tx so process_committed_block does not advance silently when the commit executor is unavailable.

Posting as a comment because GitHub rejects formal request-changes reviews on PRs authored by this same account.

umwelt:
Review addressed in d89b7208.

**P1 (closed commit channel):** `apply_block_to_state_with_proof` now returns a hard `ConsensusError::ValidatorError` when `commit_tx.send()` fails (executor receiver gone). The engine refuses to advance height. Added regression test `test_cons512_closed_commit_channel_blocks_height_advance` that wires a commit channel, immediately drops the receiver, drives a commit-quorum, and asserts `maybe_finalize` propagates the error AND `current_round.height` stays at H.

**Copilot inline findings:**
- `state_machine.rs:2648` dead-code branch removed; comment explains the timer-token discipline that makes the other branch unreachable.
- `tests.rs:955` hardcoded round literal replaced with the captured `r`.
- Design doc updated: rename claim removed (deferred to keep diff focused), NewBlock-deletion claim replaced with explanation of why variant + handler are kept (post-block bookkeeping duplication is a separate refactor; sync call IS removed in this PR), branch-name footnote added.

Tests: 67/67 lib-consensus tests pass (was 66 + new closed-channel regression).

PR #2700, opened 2026-06-09
fix(network/directory): restore SPKI HashMap with current LE cert hashes

umwelt:
## Summary
- Restores the `known_spki_pins` HashMap in `handle_get_directory` that PR #2697 deleted
- Populates it with the **currently deployed** SPKI hashes (recomputed live from each server's `/opt/zhtp/.zhtp/data/tls/server.crt`)
- g1–g5 hashes unchanged from before (long-lived self-signed certs); both gateway hashes are the Let's Encrypt certs deployed earlier today

## Why
PR #2697 emitted `"spki_pin": ""` for every peer, intending that clients fall through to TOFU. The mobile app does **not** translate empty string to `Option::None` — it constructs `TrustConfig { pin_spki: Some(\"\") }`, then `ZhtpTrustVerifier::verify_server_cert` (lib-network/src/web4/trust.rs:547-565) compares every real SPKI to `""` and rejects. Result: every authenticated mobile request failed with `QuicSession.open failed: openFailed` the moment the new binary hit the gateways.

## What's next
This is a stopgap. The table still rots on every cert rotation. Two cleanups to follow up separately:
- emit `Value::Null` (omit when missing) and require all clients to interpret missing/null as no-pin → TOFU fallback
- gossip SPKI via the on-chain peer registry so this table stops being hand-edited

## Test plan
- [x] Hashes recomputed live from each of g1–g5 + both gateways
- [x] Built dev-release binary, deployed to both gateways
- [ ] Wait for gateways to finish replay (~30 min) and verify mobile reconnects without `openFailed`

### DIFF
--- a/zhtp/src/api/handlers/network/mod.rs
+++ b/zhtp/src/api/handlers/network/mod.rs
@@ -1158,8 +1158,8 @@ impl NetworkHandler {
             .filter(|(_, v)| v.status == "active")
             .map(|(did, v)| {
                 let endpoint = ip_overlay.get(did).unwrap_or(&v.network_address);
-                let ip = endpoint.split(':').next().unwrap_or("");
-                let spki_pin = known_spki_pins.get(ip).copied();
+                let ip = Self::extract_host(endpoint);
+                let spki_pin = known_spki_pins.get(ip.as_str()).copied();
                 serde_json::json!({
                     "did": v.identity_id,
                     "role": "validator",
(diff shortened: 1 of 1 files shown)


umwelt:
Addressed Copilot's 3 comments in 7027ef37:
- Validator and gateway `spki_pin` lookups now use `.copied()` → `Option<&str>`; missing keys serialize as JSON `null` instead of `""`.
- Fixed the same anti-pattern in `compute_local_spki` (returned `String::new()` on cert-read failure, with the same downstream consequence). Now returns `Option<String>`.

Mobile path: unknown peers route through TOFU/bootstrap correctly; known peers (current 7-node table) still get their real SPKI for pinning. Build clean.

umwelt:
Addressed remaining Copilot comments (L1160 / L1186) in 64d0289a: introduced `extract_host(endpoint)` that handles bracketed IPv6 (`[2001:db8::1]:9334`) and comma-separated endpoint lists. Naive `split(':').next()` returned `"["` for the v6 case, which then missed the SPKI HashMap lookup and silently downgraded those peers to TOFU even when pinning was meant to apply. Build clean.

PR #2701, opened 2026-06-09
fix(lib-network/client): resolve hostnames in ZhtpClient::connect

umwelt:
## Summary
`ZhtpClient::connect_internal` (lib-network) parsed the address with `SocketAddr::parse`, which only accepts literal IPs. A hostname like `zhtp-gateway.thesovereignnetwork.org:9334` failed at this step before any QUIC/TLS work; the lib-client FFI then surfaced the failure to mobile as `QuicSession.open failed: openFailed`.

Switch to `addr.to_socket_addrs()` so hostnames resolve via DNS.

## Why this matters for mobile
- Mobile JS layer falls back to a hardcoded hostname (`zhtp-gateway.thesovereignnetwork.org`) whenever NetworkDirectory cache is stale or its fetch fails.
- That hostname round-trips through the FFI → spawn_session_worker → ZhtpClient::connect → connect_internal → `addr.parse::<SocketAddr>()` → **error here**.
- Symptom: every authenticated `/api/v1/wallet/*`, `/api/v1/token/*`, `/api/v1/pouw/rewards/*` call returns `openFailed`.

## Scope
- Single call site: `connect_internal`.
- `connect_to_pool` already passes `SocketAddr::to_string()` (IP-literal); unaffected.

## Test plan
- [x] `cargo build --profile dev-release -p lib-client` clean
- [ ] Mobile rebuild and verify hostname-based bootstrap path completes the UHP-v2 handshake and authenticated wallet calls succeed

### DIFF
--- a/lib-network/src/client/zhtp_client.rs
+++ b/lib-network/src/client/zhtp_client.rs
@@ -5,7 +5,7 @@
 
 use anyhow::{anyhow, Context, Result};
 use std::collections::HashMap;
-use std::net::SocketAddr;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, OnceLock};
 use tracing::{debug, info, warn};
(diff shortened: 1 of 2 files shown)


umwelt:
Addressed 92f4f338:

1. **Async DNS**: switched from `std::net::ToSocketAddrs` (blocking syscall) to `tokio::net::lookup_host` (async, runs lookup on a background thread). Important because the FFI runs us on a single-thread tokio runtime and a blocked worker == dead session.
2. **Prefer IPv4**: now filter for `is_ipv4()` first; only fall back to whatever-resolves if no A record is present. Avoids the dual-stack v6-timeout-then-v4-fallback latency on mobile carriers and captive nets.

The 3 review comments on `zhtp/src/api/handlers/network/mod.rs` belong to PR #2700 (Copilot conflated the two PRs); already addressed there in 7027ef37.

PR #2703, opened 2026-06-09
fix(lib-network/client): dual-stack bind + long-lived endpoint runtime

umwelt:
## Summary

Two iOS-specific follow-ups on top of PR #2702 (endpoint singleton). Both surfaced once the mobile team added diagnostic logging that prints the actual FFI error chain instead of just \`openFailed\`:

\`\`\`
[lib-client/quic_session] session open to 91.98.113.188:9334:
    ZhtpClient::connect: Operation not permitted (os error 1)
[lib-client/quic_session] session open to 91.98.113.188:9334:
    ZhtpClient::connect: endpoint stopping
[lib-client/quic_session] session open to 91.98.113.188:9334:
    ZhtpClient::connect: endpoint stopping
...
\`\`\`

### Bug 1 — EPERM on first bind

\`Endpoint::client("0.0.0.0:0".parse()?)\` is rejected by iOS in many sandbox contexts with EPERM. The mobile app's working \`quinn-ffi\` path has bound the IPv6 wildcard since day one for exactly this reason.

**Fix**: bind to \`[::]:0\`. Dual-stack socket accepts and sends IPv4 traffic too, so the same line works for CLI on Linux, mobile on iOS, and server-side outbound connections.

### Bug 2 — \"endpoint stopping\" cascade on every subsequent bind

\`Endpoint::client\` spawns the endpoint's async I/O driver task on **whatever tokio runtime is current at the moment of the call**. The lib-client FFI builds a fresh single-threaded runtime per session inside \`spawn_session_worker\` and drops it when that session worker exits. Whichever session happened to be the first caller into our singleton ended up owning the driver — and the moment that session's worker thread returned and its runtime dropped, the driver died and every cloned Endpoint from then on returned \`endpoint stopping\`.

**Fix**: lazy-init a process-lived \`tokio::runtime::Runtime\` alongside the singleton Endpoint (same OnceLock-+-Mutex pattern), \`enter()\` it before calling \`Endpoint::client\`, so the driver task is parented to the long-lived runtime instead of the transient per-session one. Single worker thread is plenty — only the driver task runs there; per-session client / handshake / RPC tasks continue to run on their own per-session runtimes via \`Endpoint::clone()\` and \`connect_with\`.

Both fixes are independent — Bug 1 prevented the first connect from landing; Bug 2 would still have killed all subsequent connects after the bind succeeded. Need both to clear the cascade.

## Test plan
- [x] \`cargo build --profile dev-release -p zhtp -p lib-client\` clean
- [ ] Mobile rebuild of \`libzhtp_client.a\` (iOS) and Android \`libquic_jni.so\` against this commit
- [ ] On-device: \`[lib-client/quic_session]\` should no longer print \`Operation not permitted\` or \`endpoint stopping\`; authenticated \`/api/v1/wallet/*\`, \`/api/v1/token/*\`, \`/api/v1/rewards/claim\`, \`/api/v1/msg/receive\` should return 200

## Server impact
None. The QUIC listener is already accepting these handshakes — the mobile app's working \`quinn-ffi\` path connects to the same hostname/IP/port and authenticates fine in the same logs. This patch is entirely client-side.

### DIFF
--- a/lib-network/src/client/zhtp_client.rs
+++ b/lib-network/src/client/zhtp_client.rs
@@ -19,7 +19,8 @@ use quinn::{ClientConfig, Connection, Endpoint};
 
 /// Singleton QUIC client `Endpoint` plus a dedicated long-lived runtime that
 /// owns its I/O driver task. Sharing the endpoint across all `ZhtpClient`
-/// instances is necessary for two distinct reasons; both bit us on iOS.
+/// instances is necessary for three distinct reasons, all of which bit us
+/// on iOS in the lib-client FFI path.
 ///
 /// 1. **Socket budget (Bug 1, also bites macOS/Linux at scale).** Without
 ///    sharing, every call to `new_with_config` binds a brand-new UDP socket
--- a/lib-network/src/client/zhtp_client.rs
+++ b/lib-network/src/client/zhtp_client.rs
@@ -17,17 +17,35 @@ static BOOTSTRAP_NONCE_MUTEX: std::sync::Mutex<()> = std::sync::Mutex::new(());
 
 use quinn::{ClientConfig, Connection, Endpoint};
 
-/// Singleton QUIC client `Endpoint`. Quinn's Endpoint is Arc-backed and Clone,
-/// so cloning shares the same UDP socket and I/O driver task across all
-/// `ZhtpClient` instances. Without this, every call to `new_with_config`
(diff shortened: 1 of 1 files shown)


umwelt:
Addressed in a68c6518: lead-in now says "three distinct reasons" to match the numbered list.

PR #2704, opened 2026-06-11
test(lib-network/handshake): end-to-end QUIC roundtrip

umwelt:
## Summary

Pairs \`protocols::quic_handshake::handshake_as_initiator\` against \`protocols::quic_handshake::handshake_as_responder\` on a real \`quinn::Connection\` — the exact path the lib-client FFI takes when the mobile app makes an authenticated request.

## Why now

A production report (\"UHP+Kyber handshake failed\", \"Protocol error: Failed to read length prefix byte 0: connection lost\") on the mobile path suggested the canonical initiator and responder didn't roundtrip. The existing \`handshake::core::tests::test_happy_path_handshake\` already covers the inner core over \`tokio::io::duplex\`, but the **QUIC adapter layer** wasn't being exercised end-to-end in CI:

- channel-binding export via \`Connection::export_keying_material\`
- stream framing through \`QuicStream\`
- the \`for_client_with_transport\` / \`for_server_with_transport\` context binding

This test exercises exactly that layer with the two functions the mobile/server pair use on the wire.

## What the test does

- Builds two \`quinn::Endpoint\`s (client + self-signed server) on 127.0.0.1 ephemeral ports.
- Creates two distinct \`ZhtpIdentity\`s.
- Runs the two halves concurrently behind a 30s timeout each.
- Asserts \`session_key\`, \`session_id\`, and round-tripped DIDs all match.

If anything in the QUIC adapter layer drifts (e.g. someone changes the exporter label on one side, or changes the bidi-stream finish ordering, or alters the capabilities envelope), this test fails locally before the mobile sees it.

## Result

**Passes locally** against current \`development\`. So if the same code is shipped on both sides, the canonical handshake roundtrips. Any production-only divergence points at gateway state (e.g. broken observer-sync — confirmed for gw1 stuck at h=74009) or version skew between mobile-side \`libzhtp_client.a\` and the deployed gateway binary — not at the protocol itself.

## Test plan
- [x] Compiles
- [x] Passes \`cargo test --profile dev-release -p lib-network --test quic_handshake_roundtrip_test\`

### DIFF
--- a/lib-network/tests/quic_handshake_roundtrip_test.rs
+++ b/lib-network/tests/quic_handshake_roundtrip_test.rs
@@ -1,0 +1,227 @@
+//! UHP v2 handshake round-trip test over a real QUIC connection.
+//!
+//! Pairs `lib_network::protocols::quic_handshake::handshake_as_initiator`
+//! against `lib_network::protocols::quic_handshake::handshake_as_responder`
+//! on a real quinn `Connection`. This is the path the lib-client FFI uses
+//! when the mobile app makes an authenticated request.
+//!
+//! Why this test exists: the core handshake (`handshake_as_initiator` /
+//! `handshake_as_responder` in `lib_network::handshake::core`) passes its
+//! existing roundtrip test against `tokio::io::duplex` streams, so the
+//! protocol-level message exchange is provably correct. The reported
+//! production failure ("Failed to read length prefix byte 0: connection
+//! lost", "UHP+Kyber handshake failed" mid-flight) lives one layer up, in
+//! the QUIC adapter — channel-binding export, stream framing, finish
+//! semantics. This test exercises exactly that layer with the same two
+//! functions the mobile/server pair use on the wire.
+//!
+//! If this test fails locally, the bug is reproducible in lib-network's CI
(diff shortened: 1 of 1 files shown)


umwelt:
Addressed all 6 Copilot comments in `ee44df8f`:

1. **L98 / L114**: hard-coded `b"zhtp-uhp/2"` → `lib_network::constants::ALPN_CONTROL_PLANE_V2` on both server and client config.
2. **L142 / L153**: `make_context` returns `(HandshakeContext, TempDir)`; caller binds `_nonce_dir` for the test lifetime so the temp dir cleans up on Drop instead of leaking via `std::mem::forget`.
3. **L213**: removed the dangling `let _ = conn;` — the connection drops naturally.
4. **L227**: removed `quic_handshake_default_capabilities_roundtrip`. The QUIC path uses `create_quic_capabilities()`, not the bare `Default` impl, so the test was a no-op as flagged. Replaced with a short comment explaining why for future readers.

Re-ran `cargo test --profile dev-release -p lib-network --test quic_handshake_roundtrip_test`: 1 test passed.

PR #2749, opened 2026-07-01
security: consensus hardening — mint auth, signatures, domain proofs

umwelt:
## Summary

Closes the dangerous bugs found in the post-#2745 security audit. This is a **consensus-breaking hardening** PR — deploy coordinated with mempool pause.

## CRITICAL fixes

| # | Issue | Fix |
|---|-------|-----|
| 1 | Anyone could mint canonical SOV via signed `TokenMint` | `mint_authorization.rs` — treasury signer required for SOV |
| 2 | `UBI_DISTRIBUTION_V1:` / `TOKEN_MIGRATE_V1:` memo bypass | Memo validated + treasury signer required for SOV |
| 3 | `WalletRegistration` minted arbitrary SOV from `initial_balance` | Reject `initial_balance > 0` in mempool + executor; strip from internal `register_wallet` tx |
| 4 | `add_system_transaction` bypassed all validation | `validate_system_injection()` per originator |

## HIGH fixes

- System tx signature skip narrowed — sig required for Identity/Wallet/Domain/Transfer
- ZK stub proofs (`fake`, Routing, DataIntegrity) return `Ok(false)` in production
- Identity API: user keypair + `signing_hash()` (not ephemeral `generate_keypair()` + `hash()`)
- Domain on-chain txs signed before system injection; rollback requires Dilithium verify
- Domain registry registration/transfer: real signature verification
- `TransactionBuilder` uses validator keystore
- Identity existence check fails closed when blockchain context missing

## MEDIUM fixes

- DAO `sign_dao_vote` requires real KeyPair
- `lib-economy` signing uses canonical `hash_for_signature` preimage
- Removed dead `handle_mint_sov` handler
- Web4 client signs domain updates (no empty signature)

## Tests

- `cargo test -p lib-blockchain --lib mint_authorization` — 12 pass
- `cargo test -p lib-blockchain --lib system_injection` — 2 pass
- `cargo test -p lib-network --lib test_domain_update` — pass
- `cargo test -p lib-proofs --lib zk_proof` — 7 pass
- `cargo check -p zhtp` — pass

## Deploy note

Coordinate with `docs/protocol/crypto-signing-cutover.md`. Mempool pause required — old unsigned system txs and arbitrary SOV mints will be rejected.

## Follow-up

- Executor-level TokenMint re-validation without full Blockchain context
- Replace `register_wallet` in-memory `token.mint` with kernel-authorized TokenMint
- On-chain owner pubkey for validator DomainUpdate verify (#2748)

PR #2750, opened 2026-07-01
security(audit-1,2): enforce SOV mint treasury auth, close memo bypass

umwelt:
## Audit findings
- **CRITICAL #1** — Open SOV `TokenMint` (anyone could mint canonical SOV)
- **CRITICAL #2** — Memo bypass (`UBI_DISTRIBUTION_V1:` / `TOKEN_MIGRATE_V1:`)

## Changes
- Add `mint_authorization` module with treasury-kernel checks for canonical SOV
- Validate UBI/migration memos structurally instead of skipping authorization
- Stateful `TokenMint` precheck delegates to `validate_token_mint_authorization`

## Stack
First PR in the security audit remediation stack. Merge before #02.

### DIFF
--- a/lib-blockchain/src/transaction/mint_authorization.rs
+++ b/lib-blockchain/src/transaction/mint_authorization.rs
@@ -1,0 +1,335 @@
+//! Token mint authorization validation (audit security fixes).
+//!
+//! Enforces treasury-kernel governance for SOV mints and closes memo-based
+//! authorization bypasses for UBI and migration flows at validation time.
+
+use crate::blockchain::Blockchain;
+use crate::integration::crypto_integration::PublicKey;
+use crate::transaction::core::Transaction;
+use crate::transaction::validation::ValidationError;
+
+/// Returns true when `token_id` is the canonical native SOV identifier.
+pub fn is_canonical_sov_token_id(token_id: &[u8; 32]) -> bool {
+    *token_id == [0u8; 32] || *token_id == crate::contracts::utils::generate_lib_token_id()
+}
+
+/// Returns true when `signer` is the treasury kernel governance authority or kernel address.
+pub fn is_treasury_authorized_signer(blockchain: &Blockchain, signer: &PublicKey) -> bool {
+    let Some(kernel) = blockchain.treasury_kernel.as_ref() else {
+        return false;
+    };
+    signer == kernel.governance_authority() || signer == kernel.kernel_address()
+}
+
+fn is_genesis_unsigned_mint_exception(blockchain: &Blockchain, transaction: &Transaction) -> bool {
+    blockchain.height == 0
+        && blockchain.latest_block().is_none()
+        && transaction.signature.signature.is_empty()
+}
+
+fn validate_ubi_distribution_memo(
+    blockchain: &Blockchain,
+    memo: &str,
+) -> Result<(), ValidationError> {
+    let rest = memo
+        .strip_prefix("UBI_DISTRIBUTION_V1:")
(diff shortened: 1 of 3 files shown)


umwelt:
## Security review

Fix is structurally correct — replaces the OK-return for SOV / memo-prefix bypass with treasury-kernel gating. Reviewed under the assumption that this is the mempool/block-apply parity check.

### Security concerns

**1. `is_canonical_sov_token_id` treats `[0u8; 32]` as SOV.** Magic-number convention. If a legitimate custom token was ever created with all-zero token_id (invalid via `TokenCreation` API but possible via direct byte injection), it'd be conflated with SOV and route through the treasury-auth path. Likely safe by construction but worth an explicit `debug_assert!` at token creation time to prevent the all-zero id from ever being minted as non-SOV.

**2. `is_genesis_unsigned_mint_exception` — bootstrap escape hatch.** Predicate is `height == 0 && latest_block().is_none() && signature.is_empty()`. The `latest_block().is_none()` check is the load-bearing part. Verify:
- During `replay_from_store` or `load_from_store`, does the code path that applies block 0 have `latest_block() == None` at the point this fires? If it does, replay could silently mint SOV to arbitrary addresses.
- The in-memory `height` field vs sled `latest_height()` — these can diverge during startup. If height=0 but sled has blocks, this escape hatch shouldn't fire; verify it doesn't.

Ideally the escape hatch is gated on a stronger invariant (e.g., "we are inside `apply_genesis_state()`" via a mutex/flag) rather than heuristic state.

**3. Treasury authorization surface — two keys.** `is_treasury_authorized_signer` returns true for either `governance_authority` OR `kernel_address`. Two hot keys with full SOV mint = doubled compromise blast radius. Presumably intentional (governance authorizes, kernel executes), but the doc comment should say so. If both keys live in the same keystore on the same node, the segregation is theatrical.

**4. Migration mint requires treasury auth regardless of token.** Tight for SOV, but for custom-token migrations it may over-restrict: a token owner should be able to migrate their own token without treasury sign-off. Verify migration is treasury-only by design (probably yes — it's a privileged escape hatch), and document it.

**5. `blockchain.token_contracts.get()` uses in-mem HashMap.** Post the sled-first migration epic, this can miss tokens that exist only in sled (e.g., post-restart with fast-hydrate). Result: legitimate mint of a sled-only custom token returns `InvalidTransaction`. Route through `blockchain.get_token_contract(...)` or `iter_token_contract_entries()` facade.

**6. Error semantic change `InvalidTransaction → Unauthorized`.** The updated test asserts the new variant. Confirm all downstream match arms handle `Unauthorized` — a stale `matches!(err, ValidationError::InvalidTransaction)` in mempool admission or elsewhere would silently swallow this error and treat the tx as valid.

### Test coverage gaps

- No test asserting the UBI memo bypass is actually closed (fake identity_id in memo → `Unauthorized` or `InvalidTransaction`, not `Ok`).
- No test for SOV mint by non-treasury signer → `Unauthorized`.
- No test for the genesis exception firing/not-firing.
- No test for migration by non-treasury signer.

Cheap positive controls, each one asserts the audit finding is closed.

### Verdict

Ready to merge after auditing #2 (genesis exception invariant) and #6 (downstream error match arms). #1, #3, #4 are documentation. #5 is a real bug that becomes visible post-restart. Add positive-control tests before merge if possible.

umwelt:
Copilot follow-ups addressed in branch tip: TOKEN_MIGRATE memo enforces 2592-byte Dilithium pubkey, SOV UBI mints validate recipient↔memo match at validation time, and unit tests use hex wallet ids consistent with execution.

PR #2751, opened 2026-07-01
security(audit-4): gate add_system_transaction injection paths

umwelt:
## Audit finding
- **CRITICAL #4** — `add_system_transaction()` bypasses mempool + `verify_transaction()`

## Changes
- `validate_system_injection` rejects unsigned mints except pouw paths
- Block funded `WalletRegistration` from non-treasury originators
- Reject unsigned domain system txs from web4 handlers
- Tests: `system_injection_*` in `mempool_admit_tests`

Independent of the validation.rs stack; can merge in parallel with #01.

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -2463,11 +2463,75 @@ impl Blockchain {
     /// the bypass itself is load-bearing (see #2647 Patterns A and B). This
     /// path does NOT run `lib_mempool::admit()` nor `verify_transaction`;
     /// each caller is responsible for the trust it implicitly grants.
+    /// Validate a system-originated transaction before it enters the pending pool.
+    ///
+    /// System injectors bypass signature verification at admission time; this gate
+    /// enforces minimum structural trust per originator class.
+    fn validate_system_injection(
+        originator: &'static str,
+        tx: &Transaction,
+    ) -> anyhow::Result<()> {
+        use crate::types::transaction_type::TransactionType;
+
+        let empty_sig = tx.signature.signature.is_empty();
+
+        if tx.transaction_type == TransactionType::TokenMint && empty_sig {
+            let allowed_originator =
+                matches!(originator, "pouw_mint" | "pouw_reward");
+            let memo_ok = tx.memo.starts_with(b"pouw:mint:");
+            if !allowed_originator || !memo_ok {
+                return Err(anyhow::anyhow!(
+                    "rejecting unsigned TokenMint from originator '{}': \
+                     only pouw_mint/pouw_reward with pouw:mint: memo are permitted",
+                    originator
+                ));
+            }
+        }
+
+        if tx.transaction_type == TransactionType::WalletRegistration {
+            if let Some(wallet_data) = tx.wallet_data() {
+                if wallet_data.initial_balance > 0 && !Self::is_treasury_system_originator(originator)
+                {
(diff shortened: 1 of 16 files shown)


umwelt:
## Security review

Structurally correct — gates `add_system_transaction` on originator identity + minimum tx shape. Closes the "any code path can inject any tx bypassing mempool+signature verify" bypass.

### Security concerns

**1. `originator` is `&'static str` — trust-on-first-use by hardcoded string comparison.** The auth model is "if you're compiled into the binary and your caller passes an allowlisted string, you're authorized." Any new code path that calls `add_system_transaction("admin_sov_mint", ...)` gets treasury-originator privileges instantly. Consider promoting to a typed enum with a total match, so the compiler forces exhaustive review when a new originator is added:

```rust
enum SystemOriginator {
    PouwMint, PouwReward,
    AdminSovMint, GenesisBootstrap, GenesisWallet, MigrationWallet,
    Web4DomainRegister, Web4DomainUpdate,
    Other(&'static str),  // logged + rejected for privileged ops
}
```

Currently a typo (`"admin_soV_mint"` vs `"admin_sov_mint"`) silently drops out of the treasury allowlist.

**2. `is_treasury_system_originator` uses `contains("treasury")` as one condition.** Substring match is a footgun — a future originator string like `"non_treasury_batch"` accidentally passes the treasury gate. Prefer strict match against a fixed list.

**3. Domain unsigned-tx rejection only covers `web4_domain_register` and `web4_domain_update`.** Other domain originators (`domain_transfer`, `domain_renew`, any future one) go unchecked. Same risk pattern as #1 — enumeration + compile-time exhaustiveness would close this.

**4. PoUW mint memo format `"pouw:mint:"` is a prefix check.** A crafted memo like `"pouw:mint:` followed by garbage passes the prefix test. The check should parse the memo's fields (`recipient_hex`, `amount`) and validate they match `mint_data.to` and `mint_data.amount` — otherwise a rogue PoUW originator could mint to any address regardless of memo content.

**5. `wallet_data.initial_balance > 0` gate.** What about `initial_balance == 0` but with a `TokenTransfer` follow-up injected via a separate call? The gate is per-tx, not per-operation. A non-treasury originator could register a zero-balance wallet, then inject a `TokenTransfer` that credits it (if the transfer path has its own bypass). Confirm no such downstream bypass exists — otherwise this fix is a partial closure.

**6. `empty_sig` check uses `signature.signature.is_empty()`.** An attacker who submits a `signature` of e.g. `vec![0u8]` (single zero byte) passes the empty check. The check should be "not a valid Dilithium5 signature length" — currently it accepts any non-empty byte sequence as "signed."

### Test coverage

Two tests: rejects unsigned mint from unknown originator, allows pouw with matching memo. Good starting point but missing:
- WalletRegistration with `initial_balance > 0` from non-treasury → reject
- Unsigned domain register/update → reject
- Multiple originator strings including typos and substring gotchas
- PoUW mint memo with mismatched recipient/amount vs mint_data → reject (concern #4)

### Verdict

Merge with follow-up ticket for #1 (typed enum) and #4 (memo field validation). Concerns #2/#3/#5/#6 are quick hardening — worth addressing before merge if the same author can turn them around.

umwelt:
Addressed Copilot: TokenMint gate now rejects missing mint payload for all originators; ipc_external signed TokenMint rejected; WalletRegistration without wallet_data errors; PoUW test renamed to matching_memo.

PR #2753, opened 2026-07-01
security(audit-6): require signatures for privileged system transactions

umwelt:
CRITICAL stack PR. **HIGH #6** — system transactions skip signature verification.

Expands `requires_system_tx_signature` to wallet/domain/identity registration types; system transfers with outputs/fees also require valid signatures.

Depends on #2752. Replaces monolithic #2749.

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -533,6 +533,9 @@ pub struct Blockchain {
     /// is load-bearing (Pattern A/B in #2647) and won't disappear in S2.
     #[serde(skip)]
     pub system_tx_originators: HashMap<&'static str, u64>,
+    /// Nesting depth while `apply_genesis_state` is running (audit follow-up: genesis mint gate).
+    #[serde(skip, default)]
+    pub(crate) genesis_apply_depth: u32,
 }
 
 /// Validator information stored on-chain.
@@ -2588,6 +2591,19 @@ impl Blockchain {
         }
 
         Ok(())
+    }
+
+    /// True while genesis state is being applied (stronger than height/heuristic checks).
+    pub fn is_applying_genesis_state(&self) -> bool {
+        self.genesis_apply_depth > 0
+    }
+
+    pub(crate) fn begin_genesis_apply(&mut self) {
+        self.genesis_apply_depth = self.genesis_apply_depth.saturating_add(1);
+    }
+
+    pub(crate) fn end_genesis_apply(&mut self) {
+        self.genesis_apply_depth = self.genesis_apply_depth.saturating_sub(1);
     }
 
     pub fn add_system_transaction(
--- a/lib-blockchain/src/blockchain/init.rs
+++ b/lib-blockchain/src/blockchain/init.rs
@@ -108,6 +108,7 @@ impl Blockchain {
             mempool_config: lib_mempool::MempoolConfig::audit_only(),
             mempool_state: lib_mempool::MempoolState::default(),
             system_tx_originators: HashMap::new(),
+            genesis_apply_depth: 0,
         }
     }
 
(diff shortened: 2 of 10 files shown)


umwelt:
## Security review

Straightforward tightening — extends the "system tx that still needs a signature" allowlist. Previously `WalletUpdate | TokenMint | TokenCreation | InitEntityRegistry`; now adds `WalletRegistration | IdentityRegistration | DomainRegistration | DomainUpdate` plus `Transfer` with outputs or fee.

### Security concerns

**1. `Transfer` with `outputs.is_empty() && fee == 0` still bypasses signature check.** The condition:
```rust
|| (transaction.transaction_type == TransactionType::Transfer
    && (!transaction.outputs.is_empty() || transaction.fee > 0))
```
requires sig only if `Transfer` has outputs OR fee. A "system Transfer" with zero outputs AND zero fee sails through unsigned. What's the legitimate use case for such a Transfer? If none, close the gate — the OR should probably be an unconditional "always require sig on Transfer."

**2. Doesn't cover all "state-modifying" system tx types.** Missing from the allowlist (as far as I can see from this diff):
- `ValidatorRegistration` — new validator entering the set. If a system originator can inject an unsigned ValidatorRegistration, they can register arbitrary validators.
- `GatewayRegistration` — similar.
- `NftMint` / `NftTransfer` — assuming the NFT tx types exist.
- `WelfareDaoStake` / `WelfareDaoUnstake` / any DAO governance types.
- `OracleAttestation` — if system can inject unsigned, oracle price can be manipulated.
- `RecordOnRampTrade` / `TreasuryAllocation` — the diff comment says these use threshold approval so tx-level sig is skipped. Fine, but confirm threshold approval enforcement is actually gated elsewhere; otherwise the "skip" reads as "no auth."

Enumerate every `TransactionType` variant and explicitly categorize as "requires tx-level sig," "requires threshold approval," or "purely diagnostic/no auth." Otherwise adding a new variant silently bypasses.

**3. Three copies of the same logic.** The diff shows the identical pattern being replaced in three places (stateless validate line ~611, stateless again ~903, stateful ~2351). Now all three call `requires_system_tx_signature(...)`. Good — but confirm the three sites' invocation ordering is consistent (e.g., is `is_threshold_type` applied consistently across all three?).

**4. `has_nonempty_sig = !transaction.signature.signature.is_empty()`.** Same issue as #2751 — a 1-byte signature is "non-empty" but invalid. The check should be at least "length matches Dilithium5 sig bytes." Currently a `signature: vec![0u8]` passes the shape check.

**5. The signature is checked for shape/existence, but is it cryptographically verified against the tx bytes?** This diff only shows the `require_signature` gate — the actual crypto verify happens downstream. Confirm downstream verification exists and is called for every path this gate lets through. If crypto verify is done by a separate `verify_transaction()`, some code paths may bypass it entirely (the audit finding you're addressing).

**6. `is_threshold_only` types skip the check even when they arrive with an actual signature.** If a `RecordOnRampTrade` tx carries a real signature, current code ignores it. Fine as documented, but consider "if a sig is present, verify it anyway; if absent, use threshold approval." Otherwise an attacker who can craft a threshold-approved tx AND forge a signature has an easier time than needed.

### Test coverage

Diff shows no new tests. For a HIGH severity finding closure, at minimum add:
- Unsigned `WalletRegistration` from any originator → reject
- Unsigned `DomainRegistration` → reject
- Unsigned `IdentityRegistration` → reject
- Unsigned zero-output zero-fee `Transfer` from system originator → still Ok (current behavior) or reject (recommended)
- Enumerate the fixture matrix so a new TransactionType variant produces a test failure until categorized

### Verdict

Merge with follow-ups:
- Blocker: confirm #5 (is downstream crypto verify actually invoked?). Otherwise this PR is a no-op — the sig is required but never verified.
- #1 (Transfer with zero outputs/fee) — either close or document why it's fine.
- #2 (missing variants) — walk the enum.
- #3-#4-#6 — hardening notes.

Without the tests, we're one accidental variant away from re-opening the bypass.

PR #2754, opened 2026-07-01
security(audit-8): identity citizenship uses caller key + signing_hash

umwelt:
**HIGH #8** — Identity citizenship flow uses `generate_keypair()` + `hash()` not user key + `signing_hash()`.

Replaces monolithic #2749.

### DIFF
--- a/zhtp/src/api/handlers/identity/mod.rs
+++ b/zhtp/src/api/handlers/identity/mod.rs
@@ -284,28 +284,33 @@ impl IdentityHandler {
             tracing::info!(" Creating blockchain transactions for identity and wallets");
             let did_string = format!("did:zhtp:{}", citizenship_result.identity_id);
 
-            // Create proper ownership proof by signing the DID with identity data
-            let ownership_proof_data = format!("{}:{}", did_string, citizenship_result.identity_id);
-            let ownership_proof = ownership_proof_data.as_bytes().to_vec();
+            // Use the citizen's actual key material from identity manager (not a throwaway keypair).
+            let citizen_identity = identity_manager
+                .get_identity(&citizenship_result.identity_id)
+                .ok_or_else(|| anyhow::anyhow!("Citizen identity missing after onboarding"))?;
+            let citizen_private_key = citizen_identity
+                .private_key
+                .clone()
+                .ok_or_else(|| anyhow::anyhow!("Citizen private key missing after onboarding"))?;
+            let keypair = lib_crypto::KeyPair {
+                public_key: citizen_identity.public_key.clone(),
+                private_key: citizen_private_key,
+            };
+            let public_key_bytes = citizen_identity.public_key.dilithium_pk.to_vec();
+            let ownership_proof = citizen_identity.ownership_proof.proof_data.clone();
 
             let identity_transaction_data = IdentityTransactionData::new(
                 did_string.clone(),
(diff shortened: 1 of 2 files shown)


umwelt:
## Security review

Real fix. Old code was `generate_keypair()` + `hash()` — throwaway key signing the wrong bytes. New code loads the citizen's actual keypair from `identity_manager.get_identity()` and signs the canonical `signing_hash()`.

### Security concerns

**1. Keypair reconstruction assumes `identity_manager` has the private key.** `citizen_identity.private_key.clone().ok_or_else(...)` fails if the private key isn't stored. In an environment where identity manager holds only public keys (external signing, HSM), this handler is broken. If the design is "identity manager always holds private keys," document it — otherwise this is a landmine for future migration.

**2. `keypair.public_key.dilithium_pk` is used in the `Signature.public_key` field.** But the tx's `IdentityTransactionData.public_key` is `citizen_identity.public_key.dilithium_pk.to_vec()` — same source. Good, they're consistent. Old code had a mismatch (`primary_wallet_id.as_bytes()` in the tx data vs the throwaway keypair's pubkey in the signature) — this was the audit bug.

**3. `sign_message(&keypair, signing_hash.as_bytes())` — is `sign_message` the correct API?** Depending on how `sign_message` is implemented, it may internally hash again (double-hash bug) or sign the bytes as-is. The fix passes `signing_hash.as_bytes()` which is already a hash. Confirm `sign_message` treats input as pre-hashed and doesn't rehash.

**4. `ownership_proof` now comes from `citizen_identity.ownership_proof.proof_data`.** Old code was `format!("{}:{}", did_string, citizenship_result.identity_id).as_bytes()` — a plaintext string with no cryptographic binding. New code uses the ZK proof data from the identity manager. Real improvement, but confirm downstream validators verify the ZK proof — otherwise the proof is decorative.

**5. `display_name` now comes from `req_data.display_name`.** Old code passed `citizenship_result.identity_id.to_string()` as display_name — the identity ID rendered as a name. New code passes the actual display name. This is a behavior change, not a security fix, but it's bundled here. Downstream consumers reading display_name may have relied on the old (identity ID) format.

**6. `Signature.timestamp = citizenship_result.dao_registration.registered_at`.** Signature includes a timestamp field. If the verifier doesn't check `signature.timestamp` matches something in the tx (or is fresh), an attacker with a valid signed tx can replay with any timestamp. The fix reuses the DAO registration timestamp — not "now" — so replay-window guards may not apply. Confirm timestamp is validated (or explicitly documented as ignored).

**7. `citizen_identity.public_key.clone()` is used in `KeyPair { public_key, private_key }`.** No verification that public_key derives from private_key. Same concern as #2745 review — if the identity store has a mismatched key pair (bug or attack), signing succeeds with the wrong pubkey embedded. Add sk→pk derivation check, OR document that identity_manager guarantees consistency.

### Test coverage

Diff shows no tests. For a HIGH severity finding, at minimum:
- Craft a citizen with a known keypair, run the flow, verify the emitted transaction verifies against the citizen's registered public key
- Craft a citizen with missing private_key → assert the handler returns an error, doesn't fall through to a throwaway keypair

### Verdict

Correct fix. Concerns #1 (design boundary), #3 (double-hash risk), #4 (proof verification downstream), #7 (sk/pk consistency) are the hardening items. #5/#6 are behavior changes that should be flagged in the PR body.

PR #11, opened 2025-08-12
Fix 'cargo run' command for linux

LazyHatGuy:

PR #2755, opened 2026-07-01
security(audit-11): fail-closed when blockchain context is missing

umwelt:
**HIGH #11** — identity existence check fails open when `blockchain` context is `None`.

`validate_sender_identity_exists` returns `Err` instead of skipping verification.

Depends on prior stack PR. Replaces monolithic #2749.

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -531,6 +531,9 @@ pub struct Blockchain {
     /// is load-bearing (Pattern A/B in #2647) and won't disappear in S2.
     #[serde(skip)]
     pub system_tx_originators: HashMap<&'static str, u64>,
+    /// Nesting depth while `apply_genesis_state` is running (audit follow-up: genesis mint gate).
+    #[serde(skip, default)]
+    pub(crate) genesis_apply_depth: u32,
 }
 
 /// Validator information stored on-chain.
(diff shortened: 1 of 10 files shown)


umwelt:
## Security review

Tiny fix, big impact. Closes a fail-open where `validate_sender_identity_exists` returned `Ok(())` when `self.blockchain` was `None` — with a `warn!` log that would have been the only indicator.

### Concerns

**1. Error type is `InvalidTransaction`.** Semantically the tx isn't invalid; the validator is missing context. Consider a dedicated `ValidationError::MissingBlockchainContext` variant so ops can distinguish "the tx is bad" from "the validator was misconfigured." Currently a spike in `InvalidTransaction` on real user txs looks identical to a config regression that drops blockchain context.

**2. Callers of `StatefulTransactionValidator` that ever legitimately construct one with `blockchain: None`.** After this PR, any such caller starts failing every stateful check. Enumerate call sites in the same PR — otherwise this closes the security hole and breaks legitimate paths simultaneously. Test surfaces this if the test harness constructs a `StatefulTransactionValidator` with `None`.

**3. Log-and-fail-open pattern was a smell.** The previous `warn!("SECURITY WARNING: ...")` is exactly the pattern that should have been a compile-time impossibility. The struct could take `NonNullable<&'a Blockchain>` instead of `Option<&'a Blockchain>` so this branch doesn't exist. That's out of scope for this PR but worth a follow-up.

### Test coverage

Diff shows no tests. Add:
- Construct `StatefulTransactionValidator` with `blockchain: None`
- Call `validate_sender_identity_exists` on any tx
- Assert it returns `Err(InvalidTransaction)`, not `Ok(())`

Trivial to add. Without it, someone can regress the fix and produce the same fail-open.

### Verdict

Merge. Small surface, obvious win. Follow-up: convert `Option<&Blockchain>` to a non-optional field on the struct.

PR #2756, opened 2026-07-01
security(audit-12): TransactionBuilder uses validator keystore

umwelt:
**HIGH #12** — `TransactionBuilder` uses ephemeral `generate_keypair()` per system tx.

Replaces monolithic #2749.

### DIFF
--- a/zhtp/src/runtime/services/transaction_builder.rs
+++ b/zhtp/src/runtime/services/transaction_builder.rs
@@ -123,20 +123,29 @@ impl TransactionBuilder {
 
     /// Load the node's validator/kernel signing key for system transactions.
     ///
-    /// Production nodes MUST have `ZHTP_KEYSTORE_DIR` (or default `node_data/keystore`)
-    /// populated with the validator private key. Under `cfg(test)` only, falls back to a
-    /// deterministic test keypair when the keystore is absent.
+    /// Production nodes MUST have `ZHTP_KEYSTORE_DIR` set, or use `<data-dir>/keystore`
+    /// (default data-dir: `~/.zhtp`, overridable via `--data-dir`) with the validator key.
+    /// Under `cfg(test)` only, falls back when the keystore file is missing — not on parse errors.
     async fn load_system_keypair() -> Result<lib_crypto::KeyPair> {
         match crate::runtime::token_utils::load_validator_keypair_from_keystore().await {
             Ok(kp) => Ok(kp),
             Err(e) => {
                 #[cfg(test)]
                 {
-                    tracing::warn!(
-                        "Validator keystore unavailable in test ({}); using deterministic test keypair",
-                        e
-                    );
-                    Ok(Self::deterministic_test_system_keypair())
+                    let missing = e.chain().any(|cause| {
+                        cause
+                            .downcast_ref::<std::io::Error>()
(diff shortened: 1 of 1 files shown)


umwelt:
## Security review

Replaces ephemeral `generate_keypair()` in `create_system_signature` with a keystore-backed load. In production the validator's stored key is used; in tests a deterministic keypair from a fixed entropy seed.

### Security concerns

**1. `load_validator_keypair_from_keystore` behavior on failure.** The `cfg(not(test))` path returns an error — the tx build fails, no signature produced. Good. But what does the caller do with that error? If the caller retries, or falls back to another path, the fix is partial. Verify all callers of `create_system_signature` propagate the error as a hard failure.

**2. Deterministic test keypair under `cfg(test)`.** `dilithium5_keypair_from_entropy(b"zhtp-test-system-tx-signing-seed")` — the SAME seed for every test build. Every unit test using this signs with the SAME identity. If a test accidentally submits to a shared blockchain instance across test cases (parallel tests hitting a global), signatures collide with real-world implications only in tests. But: if any INTEGRATION test uses this keypair and the derived pubkey ever accidentally ends up in a live sled snapshot committed to git as fixture data, that pubkey (and its private key) is effectively public. Test-only isolation must be strict.

**3. `#[cfg(test)]` fallback within `load_system_keypair`.** This is a common footgun: `#[cfg(test)]` fires only for the crate's own `cargo test`, not for downstream crates that pull this crate as a dep in their tests. If `lib-network` or `zhtp` runs a test that transitively exercises `create_system_signature`, the crate compiled here is `!test` → hard error. If that's desired, document it. If tests in dependent crates need this to work, use a runtime `#[cfg(feature = "test-utils")]` feature-flagged fallback instead.

**4. No check that the loaded keypair matches the node's identity.** `load_validator_keypair_from_keystore()` returns whatever's in the file. If ops accidentally deploys the wrong keystore, this signer signs system txs with a foreign identity — which validators/mempool then either accept (if the pubkey is a known validator) or reject. The failure mode is at consensus, not at signing. A defense-in-depth check would compare `keypair.public_key.dilithium_pk` against the node's known validator pubkey (from config or on-chain registration) and refuse to sign if they don't match.

**5. `keystore` file layout is implicit.** The comment says `ZHTP_KEYSTORE_DIR` or default `node_data/keystore`. Verify the file permissions (`0600` at rest? whose UID owns it?). If any other user on the box can read the keystore, system-tx signing is compromised. This isn't in scope for this PR but ensure ops runbook mandates the file mode.

**6. `sign_message` receives which bytes?** Diff shows `sign_message(&system_keypair, ...)` — the `...` isn't in the visible patch. Confirm the signed bytes are the canonical `signing_hash()`, not a raw `hash()`. This is the same bug class as #2754 — the fix substitutes the key but if the hash is wrong, verification still fails.

### Test coverage

Diff shows no new tests. Add:
- Production build (feature/cfg flag) with missing keystore → `create_system_signature` errors
- Test build with missing keystore → deterministic fallback fires, sig verifies against the known deterministic pubkey
- Real keystore file present → loaded, sig verifies

### Verdict

Merge with follow-ups:
- Blocker if applicable: #6 (confirm signed bytes are `signing_hash()`, not raw `hash()`)
- #4 is the practical hardening — a mismatched keystore in prod is a real operator risk
- #3 (test-utils vs cfg(test)) matters if any dependent crate uses this path in its tests

PR #2757, opened 2026-07-01
security(audit-7): reject stub ZK proof systems in production

umwelt:
**HIGH #7** — ZK verifier accepts `fake` / stub proof systems as valid.

Rejects stub proof systems (fake/Routing/DataIntegrity) outside test builds.

Replaces monolithic #2749.

### DIFF
--- a/lib-proofs/src/types/zk_proof.rs
+++ b/lib-proofs/src/types/zk_proof.rs
@@ -363,11 +363,11 @@ impl ZkProof {
                 "ZHTP-Optimized-StorageAccess" => backend.verify_storage_access(backend_proof),
                 "ZHTP-Optimized-Merkle" => backend.verify_merkle(backend_proof, [0u8; 32]),
                 "ZHTP-Optimized-Routing" | "ZHTP-Optimized-DataIntegrity" | "fake" => {
-                    #[cfg(feature = "fake-proofs")]
+                    #[cfg(any(test, feature = "fake-proofs"))]
                     {
                         Ok(true)
                     }
-                    #[cfg(not(feature = "fake-proofs"))]
+                    #[cfg(not(any(test, feature = "fake-proofs")))]
                     {
                         Ok(false)
                     }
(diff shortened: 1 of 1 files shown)


umwelt:
## Security review

Replaces `Ok(true)` stub returns for `ZHTP-Optimized-Routing`, `ZHTP-Optimized-DataIntegrity`, and `"fake"` with feature-gated behavior: with `fake-proofs` on → true, without it → false.

### Security concerns

**1. Compile-time gate is footgun-prone.** `#[cfg(feature = "fake-proofs")]` fires per-crate. If ANY crate in the workspace enables `fake-proofs` (a dev-dependencies feature bleed, a workspace-wide `--all-features` build, a Cargo profile in `.cargo/config.toml`), production binaries silently accept fake proofs. Verify:
- No `default = ["fake-proofs"]` in `lib-proofs/Cargo.toml`
- CI has an explicit `cargo build --release --no-default-features` check (or equivalent) that FAILS if `fake-proofs` is on
- Ideally add `compile_error!` gated on `fake-proofs && !cfg(test)` — any production build that pulls this feature is a hard build error

**2. Returning `Ok(false)` instead of `Err`.** Callers may log "verification returned false" and continue rather than treat it as fatal. Grep callers of `verify()` — if any of them swallow `Ok(false)` (retry, skip, mark as "unverified but included"), the fix leaks. `Err(anyhow::anyhow!("stub proof types are not accepted in production"))` gives callers no room to mistake it.

**3. Collapsed match arm mixes semantic categories.** `"fake"` is obviously a stub. But `"ZHTP-Optimized-Routing"` and `"ZHTP-Optimized-DataIntegrity"` are real ZHTP proof categories according to naming. If future code adds a REAL routing proof verifier, someone must remember to break it out of this match arm — otherwise the real proof gets rejected as a stub. Add a `TODO` or (better) split the arms and put a comment on each explaining what it means.

**4. What emits these proof types?** grep the codebase for constructors that produce proofs with these system strings. If NO caller ever emits `ZHTP-Optimized-Routing`, this is dead code — remove entirely. If SOME callers emit them, they now fail verification — that's a functional regression that needs a migration plan for existing pending proofs.

**5. Missing test coverage.** Diff shows no new tests. Add:
- Production build (`--no-default-features`): each of the three stub proof types → `verify` returns `Ok(false)` (or `Err`, per #2)
- Feature enabled: each returns `Ok(true)`
- Regression test that any workspace member enabling `fake-proofs` fails a specific CI job

### Verdict

Merge only after confirming:
- CI blocks `fake-proofs` in release builds
- No production caller emits these proof type strings, OR those callers now have a migration path
- Preferably: switch to `Err` so the rejection can't be silently continued past

PR #2758, opened 2026-07-01
security(audit-13): DAO vote signing requires real KeyPair

umwelt:
**MEDIUM #13** — DAO vote signing uses BLAKE3 hash as Dilithium signature.

Replaces monolithic #2749.

### DIFF
--- a/lib-governance/src/dao/dao_engine.rs
+++ b/lib-governance/src/dao/dao_engine.rs
@@ -11,7 +11,7 @@ use crate::dao::{
     GovernanceParameterValue, PrivacyLevel,
 };
 use anyhow::Result;
-use lib_crypto::{hash_blake3, Hash};
+use lib_crypto::{hash_blake3, sign_message, Hash, KeyPair};
 use lib_identity::IdentityId;
 use lib_types::consensus::{ConsensusConfig, MIN_BFT_VALIDATORS as MIN_VALIDATORS};
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


umwelt:
## Security review

Removes hash-as-signature stub in `DaoEngine::sign_dao_vote` and requires a real `KeyPair` from the caller. Uses `sign_message(keypair, &vote_data)` for the actual Dilithium signature.

### Security concerns

**1. `#[allow(dead_code)]` is still on the function.** If no caller invokes `sign_dao_vote`, this is a hardening improvement to a dead branch. Meanwhile the LIVE vote-signing path (wherever DAO votes are actually signed) may still be stub-based. Grep for how DAO votes are actually signed today — if it's a different function that still uses hash-as-signature, this PR doesn't fix the audit finding. If this IS the live path, remove `#[allow(dead_code)]`.

**2. Signed bytes are raw `vote_data`, not a canonical `signing_hash()`.** The concatenation `[voter||proposal_id||choice]` is signed directly. Concerns:
- No domain separator — an attacker could try to construct a colliding input that mimics a vote payload from another context (some other structure that also concatenates identity/hash/u8 fields). Add a domain tag: `["ZHTP:dao_vote_v1", voter, proposal_id, choice_u8]` before signing.
- Different from tx signing convention (`signing_hash()` on a Transaction). If verifiers use `signing_hash`-style pre-hashing on the counterparty side, verification will mismatch. Confirm the verifier hashes the SAME bytes with the SAME layout.

**3. `keypair.ok_or_else(...)` — where does the caller get the keypair from?** If callers do `sign_dao_vote(Some(&some_keypair), ...)` with a keypair they LOAD FROM somewhere insecure (a request body, a temp file), this is worse than the stub because a signature that verifies is trusted. Trace all call sites and confirm the keypair is the citizen's own key loaded from `identity_manager` (see #2754), not from user input.

**4. `SystemTime::now().duration_since(UNIX_EPOCH)` was previously in the stub timestamp.** Now the timestamp comes from wherever `sign_message` puts it. Verify the signed message includes the timestamp so replay-attack windows are bounded, OR that consumers of DAO votes verify freshness some other way (proposal deadline check).

**5. Return type unchanged — `Result<lib_crypto::Signature>`.** `Signature` struct fields include `signature`, `public_key`, `algorithm`, `timestamp`. Confirm `sign_message` populates a REAL Dilithium signature (not a BLAKE3 hash re-labeled) in `.signature`. Cross-check `lib_crypto/src/signing.rs` if this hasn't already been audited.

### Test coverage

Diff shows no new tests. Add:
- Sign then verify: `sign_dao_vote(&kp, voter, proposal, choice)` → returned signature verifies against `kp.public_key`
- Wrong choice: signed with `Approve`, verify against `Reject` payload → fails
- Wrong voter/proposal_id: verify against tampered fields → fails
- Missing keypair: `sign_dao_vote(None, ...)` → err

### Verdict

Merge-ready if (a) this function is actually called by the live vote-submission path (not just dead code), (b) verifier hashes identical bytes with identical layout, (c) domain separator gets added. Otherwise this is a partial fix that leaves the real vote signing path untouched.

PR #2759, opened 2026-07-01
security(audit-9): verify domain registration/transfer Dilithium proofs

umwelt:
**HIGH #9** — lib-network domain registry registration/transfer proofs are format-only.

Adds `domain_signing` module; registry verifies real Dilithium signatures on registration/transfer.

Replaces monolithic #2749.

### DIFF
--- a/lib-network/src/web4/domain_registry.rs
+++ b/lib-network/src/web4/domain_registry.rs
@@ -638,6 +638,15 @@ impl DomainRegistry {
         self.register_domain(request).await
     }
 
+    /// Whole-SOV registration fee used in the registration signing preimage.
+    fn expected_registration_fee_whole(domain: &str, duration_days: u64) -> u64 {
+        let base_fee = 10.0_f64;
+        let per_day_fee = 0.01_f64;
+        let premium_multiplier = if domain.len() <= 6 { 3.0 } else { 1.0 };
+        let total = (base_fee + (duration_days as f64 * per_day_fee)) * premium_multiplier;
+        total.round() as u64
+    }
+
     fn sign_registration_fields(
         owner: &ZhtpIdentity,
         domain: &str,
@@ -654,7 +663,7 @@ impl DomainRegistry {
         let registration_timestamp = std::time::SystemTime::now()
             .duration_since(std::time::UNIX_EPOCH)?
             .as_secs();
-        let registration_fee_whole = 10u64;
+        let registration_fee_whole = Self::expected_registration_fee_whole(domain, 365);
         let message = domain_registration_signing_message(domain, registration_timestamp, registration_fee_whole);
         let sig = sign_message(&keypair, &message).map_err(|e| anyhow!("Failed to sign registration: {}", e))?;
 
@@ -1068,6 +1077,16 @@ impl DomainRegistry {
             return Ok(false);
         }
 
+        let expected_fee =
+            Self::expected_registration_fee_whole(&request.domain, request.duration_days);
+        if request.registration_fee_whole != expected_fee {
+            warn!(
+                "Domain registration rejected: fee {} does not match expected {} for {}",
+                request.registration_fee_whole, expected_fee, request.domain
+            );
+            return Ok(false);
+        }
+
         verify_domain_registration_signature(
             request.owner.public_key.dilithium_pk.as_slice(),
             &request.domain,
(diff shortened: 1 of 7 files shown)


umwelt:
## Security review

Replaces format-only proof checks with real Dilithium5 signature verification on domain registration and transfer. Adds `owner_signature_hex`, `registration_timestamp`, `registration_fee_whole` on `DomainRegistrationRequest`, and time-window + presence gates in `verify_registration_proof`. Transfer verification decodes a timestamp from `proof.public_inputs[..8]` and calls `verify_domain_transfer_signature`.

### Security concerns

**1. Server-side private key material — key management smell.** `sign_registration_fields` does `owner.private_key.as_ref().ok_or(...)` and reconstructs a `KeyPair`. That means `ZhtpIdentity` on the SERVER carries the owner's private key. Registration should be caller-signed and server-verified — the server never sees `sk`. If this code path is only used server-side for test/dev fallback, gate it under `#[cfg(test)]` or a feature flag. If it runs in production when `registration_auth` is `None`, that's a red flag: production must NEVER fall back to server-signing on behalf of the owner. Reject with `err` instead.

**2. Hard-coded `registration_fee_whole = 10u64` in server-side signing fallback.** The signature covers `(domain, timestamp, fee)`. If the actual on-chain registration fee changes and the client submits a signature covering `fee=15`, but this fallback signs `fee=10`, verification succeeds against the WRONG fee. Even if this is fallback-only, mismatch is silent and hard to trace. Either pass `fee` in as a parameter or remove the fallback entirely.

**3. `hex::decode(&owner_signature_hex).unwrap_or_default()` on line ~615.** If the hex is malformed, `proof_data` is empty. This is then embedded in `registration_proof`. But `verify_registration_proof` checks `owner_signature_hex` separately (also `is_empty`), so malformed hex here means the request never gets past the second check. Still: silent decode failure is a code smell. Use `?` and return early with a clear error at the boundary.

**4. No nonce → replay window within `time_diff <= 300`.** Signature covers `(domain, ts, fee)`. If a fee payment tx and registration are separate transactions and the registration takes seconds to propagate, an attacker who observes the sig can resubmit — but only if the first attempt failed AND their resubmission is within 5 minutes. Once the domain is registered, `register_domain` should reject duplicates on the domain-taken check. Still: for high-value domains, include a nonce (from server) or bind the signature to a specific tx hash. Confirm `register_domain` DOES reject before-registration duplicates deterministically.

**5. Transfer timestamp: `proof.public_inputs.len() == 8` gate.** If `public_inputs` has length other than 8, `timestamp = None`. Then `verify_domain_transfer_signature` accepts `Option<u64>`. Check that function's behavior when `timestamp=None`:
- If it accepts and skips the time-window check → any old signature from the from_owner replays as a valid transfer forever.
- If it rejects None → good, but why fall back instead of erroring immediately here?
The 8-byte length check is fragile: if a future patch adds anything else to `public_inputs` (e.g. protocol version prefix), timestamp silently becomes None. Change the format to be explicitly tagged.

**6. Timestamp window `> 300` — 5 minutes.** Standard. Two things:
- Absolute-value diff means BOTH future-dated and past-dated signatures within 300s window are accepted. A future-dated sig by 4 minutes is a small future-replay vector; probably fine.
- No monotonic anti-replay — the same signature can be replayed in the SAME 300s window (see #4). Fine only if domain-taken check is bulletproof.

**7. `dilithium-domain-registration` proof system string.** The ZK verifier in #2757 doesn't have an arm for this. Confirm that registration proofs are NOT routed through `ZkProof::verify()`. If they are, they now unconditionally fall into the `_ => Ok(false)` bucket and every registration fails. If they're NOT (they go through `verify_registration_proof` here directly), fine — but that means `registration_proof` is essentially dead data on the wire. Consider removing it from the struct.

**8. `verify_domain_registration_signature` signature.** Not visible in this diff. Confirm:
- Domain-separated (e.g. prefixes `"ZHTP:domain_register_v1"` in the signed message)
- Hashes with `signing_hash()`-equivalent canonicalization
- Uses `verify_signature` from `lib_crypto`, not a hand-rolled hash comparison

### Test coverage

New tests fix up the request-building helpers to sign with the owner's keypair. But nothing tests:
- Invalid hex signature → registration rejected
- Wrong `fee_whole` at verification (mismatched from signature) → registration rejected
- Wrong-owner signature (attacker signs for a different identity) → rejected
- Expired timestamp (>300s) → rejected
- Future-dated timestamp (>300s ahead) → rejected
- Transfer with `public_inputs` length != 8 → what actually happens
Add these explicitly.

### Verdict

Do NOT merge until:
- Server-side private-key path (`sign_registration_fields`) is gated to test/dev-only or removed
- `verify_domain_registration_signature` and `verify_domain_transfer_signature` are audited (they're referenced but not diffed here)
- Domain-taken duplicate rejection is verified deterministic
- Replay-safety tests are added

The direction is correct, the plumbing needs another pass.

PR #2760, opened 2026-07-01
security(audit-15): lib-economy signing uses canonical signing_hash

umwelt:
**MEDIUM #15** — lib-economy signing digest ≠ canonical `signing_hash()`.

Replaces monolithic #2749.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3592,6 +3592,8 @@ name = "lib-economy"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "bincode",
+ "blake3",
  "chrono",
  "criterion",
  "hex",
--- a/lib-economy/Cargo.toml
+++ b/lib-economy/Cargo.toml
@@ -9,6 +9,8 @@ repository = "https://github.com/SOVEREIGN-NETWORK/ZHTPDEV"
 
 [dependencies]
 anyhow = "1.0"
+bincode = "1.3"
+blake3 = "1.5"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 hex = "0.4"
--- a/lib-economy/src/transactions/blockchain_integration.rs
+++ b/lib-economy/src/transactions/blockchain_integration.rs
@@ -19,6 +19,9 @@ use blake3::Hasher as Blake3Hasher;
 use lib_crypto::types::keys::PublicKey as CryptoPublicKey;
 use serde::Serialize;
 
+/// Mirror of lib-blockchain `TX_VERSION_V8`.
+const BLOCKCHAIN_TX_VERSION_V8: u32 = 8;
+
 /// Economy transaction data needed for blockchain conversion
 /// This struct provides the necessary data without importing blockchain types
 #[derive(Debug, Clone)]
@@ -163,9 +166,11 @@ pub fn to_blockchain_data(
     )
     .into_bytes();
 
-    let chain_id_u8 = chain_id.min(u8::MAX as u32) as u8;
+    let chain_id_u8: u8 = chain_id
+        .try_into()
+        .map_err(|_| anyhow::anyhow!("chain_id {} exceeds u8 wire format", chain_id))?;
     let signing_hash = hash_for_signature_mirror(
-        1,
+        BLOCKCHAIN_TX_VERSION_V8,
         chain_id_u8,
         BlockchainMirrorTxType::Transfer,
         &mirror_outputs,
@@ -176,7 +181,7 @@ pub fn to_blockchain_data(
     let signature = sign_message(system_keypair, &signing_hash)?;
 
     Ok(BlockchainTransactionData {
-        version: 1,
+        version: BLOCKCHAIN_TX_VERSION_V8,
         chain_id,
         tx_type_name,
         inputs,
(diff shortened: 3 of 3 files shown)


umwelt:
## Security review

Replaces ad-hoc `"chain:type:amount:to:timestamp"` string signing in `lib-economy` with a mirror of `lib_blockchain::transaction::hashing::hash_for_signature`. Signing is done on the canonical hash. Adds `bincode` + `blake3` deps to `lib-economy`.

### Security concerns

**1. CRITICAL — the mirror is a stale-copy hazard.** `BlockchainMirrorTxType`, `BlockchainMirrorPayload`, `BlockchainMirrorHash`, `BlockchainMirrorOutput`, and `hash_for_signature_mirror` all replicate blockchain-side types by hand. Any of the following silently diverge:
- `TransactionType` gains a new variant / discriminant reordering → mirror's `Transfer = 0` is wrong
- `TransactionPayload` gains a new default variant that differs from `None`
- `TransactionOutput` field order changes or a field is added → `bincode::serialize` produces different bytes
- `hash_for_signature` changes byte layout (e.g. adds a domain prefix, changes le/be endianness)

None of these compile-fail. All make signatures signed by lib-economy silently unverifiable on-chain, which halts UBI/rewards entirely.

**Required mitigation:** add a cross-crate integration test that
```
lib_blockchain::hash_for_signature(&real_tx) == hash_for_signature_mirror(...same inputs)
```
for a fixed test vector. If lib-economy is Layer 2 and can't depend on lib-blockchain, put the test in `zhtp` (Layer 3+) where both are available. Without this test the "mirror" WILL drift.

**Better fix:** move the canonical hash into a Layer-1 crate (`lib-crypto` or a new `lib-tx-hash` crate) that BOTH lib-blockchain and lib-economy depend on. The current design keeps duplicating a security-critical serialization.

**2. REGRESSION — recipient is now always the zero pubkey.**
```rust
let recipient_pk = CryptoPublicKey::new([0u8; 2592]);
let outputs = vec![BlockchainOutput {
    ...
    recipient: recipient_pk.dilithium_pk.to_vec(), // <-- zero pubkey
}];
```
Previous code: `recipient: economics_tx.to.to_vec()` — used the actual recipient bytes.

Every UBI distribution and reward tx now emits a recipient of `[0u8; 2592]`. That's either:
- A load-bearing sentinel that downstream code translates back to `economics_tx.to` (via a lookup on `note` or the memo). If so, DOCUMENT this and confirm the sink handles it. If NOT and the raw output goes on-chain, every reward tx lands at the zero address.
- An unfixed bug from the mirror code path that got in during the refactor.

Either way, this is a functional regression that needs to be resolved before merge. The signing-hash change and the recipient change are conflated in one diff.

**3. `chain_id_u8 = chain_id.min(u8::MAX as u32) as u8`.** Silent truncation. Change to hard error:
```rust
let chain_id_u8: u8 = chain_id.try_into()
    .map_err(|_| anyhow!("chain_id {chain_id} exceeds u8 for hash canonicalization"))?;
```
If the mirror hash uses u8 chain_id but the tx wire format uses u32, that's ANOTHER stale-copy hazard: the tx-side hash-for-signature must ALSO use u8 or the two never agree.

**4. `version: 1` hardcoded in `hash_for_signature_mirror` call.** Same drift concern: real blockchain may bump tx version and this mirror keeps signing v1. Wire `data.version` in.

**5. `memo` embeds a debug string with a `Vec` `{:?}` representation.**
```rust
let memo = format!("System TX: {} {} SOV to {:?}", ..., economics_tx.to);
```
`{:?}` on `Vec<u8>` prints `[123, 45, 67, ...]`. This memo is signed AND on-chain. It's not a security problem per se, but it's a stability hazard — if Rust's Debug for `Vec<u8>` ever changes format, signatures diverge. Use `hex::encode(&economics_tx.to)` for a stable representation.

**6. `commitment` / `note` derivation is non-committal.**
```rust
let commitment = hash_blake3(format!("commitment_{}", economics_tx.amount).as_bytes());
```
`commitment` only binds to `amount` (a decimal string). Two UBI payouts of the same amount produce IDENTICAL commitments. That defeats the entire commitment abstraction — commitments should be unique per output. Include `to`, `timestamp`, `tx_id` in the pre-image at minimum. As-is, replay/collision within a UBI epoch is a real concern for downstream double-spend checks.

Same for `note`: it hashes `tx_id` which IS unique, but the pattern of embedding format strings is fragile.

**7. `payload_bytes.len()` prefix before payload bytes.** This differs from the standard bincode serialization pattern for enums (bincode already length-prefixes internally). Cross-check that the blockchain-side `hash_for_signature` does the same length prefix, or the hashes diverge on the payload field.

### Test coverage

Existing tests only check that fields are populated and non-empty. Nothing:
- Verifies the mirror hash equals the blockchain-side canonical hash for a known input
- Signs then verifies against on-chain hash format
- Round-trips through the blockchain to ensure sig verifies at consensus

### Verdict

**Do NOT merge as-is.** Three showstoppers:
1. Zero-pubkey recipient regression
2. No cross-crate mirror consistency test
3. Silent chain_id truncation

The signing-hash direction is correct, but the mirror-by-hand approach is a booby-trap. Refactor into a shared hash module accessible from both crates before merging.

Issue #10, opened 2025-08-09
DNS resolution failed: cant access property "toLowerCase", query is undefined


GitHubRomania:
**DNS resolution failed: cant access property "toLowerCase", query is undefined**

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/1435e0b7-79a2-4bd8-b895-0cfc906d34bb" />

**Im runnig a : 
cat /etc/os-release
NAME="Fedora Linux"
VERSION="42 (Workstation Edition)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=42
 GNOME**

PR #2761, opened 2026-07-01
security(audit-5,10): sign domain txs with caller key, verify rollback

umwelt:
**HIGH #5** — Web4 domain on-chain txs with zeroed signatures via system bypass.
**HIGH #10** — Domain rollback: no signature, spoofable `x-owner-did` header.

Domain API handlers sign with caller keys + `signing_hash()`; rollback requires real signature.

Replaces monolithic #2749.

### DIFF
--- a/zhtp/src/api/handlers/web4/domains.rs
+++ b/zhtp/src/api/handlers/web4/domains.rs
@@ -10,6 +10,9 @@ use lib_identity::{types::IdentityView, ZhtpIdentity};
 use lib_access_control::{SecurityPrincipal, Role};
 use lib_types::NodeType;
 use lib_network::web4::{
+    domain_signing::{
+        has_owner_signing_key, validate_domain_owner_signature_hex, verify_domain_update_signature,
+    },
     DomainEconomicSettings, DomainLookupResponse, DomainMetadata, DomainRegistrationRequest,
     PublicOwnerInfo,
 };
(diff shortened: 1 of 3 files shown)


umwelt:
## Security review

Replaces zeroed-signature domain system transactions with owner-signed txs via a new `sign_domain_system_transaction` helper. Adds real Dilithium verification on rollback: reads owner PK from the registry record, verifies the rollback signature over (domain, current_cid, target_cid, timestamp), derives `owner_did` from the record instead of trusting the `x-owner-did` header.

### Security concerns

**1. HIGH — server-side private key handling.** `sign_domain_system_transaction` does `owner_identity.private_key.clone().ok_or(...)?`. Two implications:
- The server (API handler) is now in possession of the caller's private key. This should be the CLIENT's job — sign on the client, submit the signed tx to the server. If the server ever leaks memory (a panic dump, an OOM heap capture), that key is exposed.
- Where does `owner_identity` come from in `deploy_manifest_domain` / register_domain paths? If it's `identity_manager.get_identity(caller_did)` and the manager loads the sk from an authenticated session, that's the pattern from #2754. Confirm identity_manager here similarly authenticates before returning private_key. If a caller can trick this endpoint into signing with a foreign identity, that's exploitable.

**2. `chain_id: 0x03` hardcoded in TWO call sites.** Where does 0x03 come from? If the running chain is not chain_id 3, this tx signs the WRONG chain_id and consensus rejects it. Even if 3 is right today, hardcoding is a footgun — pull from protocol params.

**3. Fee bypass at web4 layer.** The system-tx path still calls `add_system_transaction(domain_tx, "web4_domain_register")`. That means these ownership-signed domain-registration txs bypass the fee payment path. Per the domain-fee architecture (project memory: fee_payment_tx is client-signed, gateway forwards via mempool), this appears to still use the system-tx bypass — the audit finding was that these txs are ALSO domain state-changing and shouldn't ride on the system bypass. Confirm that (a) the caller separately submits a `fee_payment_tx`, and (b) `add_system_transaction` for `web4_domain_register` is gated by #2751's `validate_system_injection` — under that gate, an unsigned domain tx from a non-treasury originator SHOULD be rejected. With this PR, the tx IS signed (good), but it's still routed through `add_system_transaction`. Verify #2751's gate accepts SIGNED domain txs from any originator, not just treasury.

**4. `sign_domain_system_transaction` returns a tx with `fee: 0`.** Combined with #3, the whole domain-fee mechanism relies on a separate `fee_payment_tx`. That's fine if the fee tx is actually being emitted and confirmed BEFORE the domain tx is accepted. If they can be reordered by the mempool, an attacker registers a domain, doesn't pay the fee, and the domain-registration tx commits first. Confirm the coupling.

**5. `ApiDomainRegistrationRequest.transfer_proof` is now used as raw hex-decoded Dilithium sig.** But the transfer path passes `signature_bytes` as `proof.proof_data` and `vec![]` as `verification_key`. In `verify_transfer_proof` (from #2759 review), the fallback is `timestamp = None` when `public_inputs.len() != 8`. Here `public_inputs = vec![]` → len 0 → timestamp=None. If `verify_domain_transfer_signature` accepts `None` timestamp and skips the time-window check, this transfer replays FOREVER. Regression from #2759 — check both PRs together.

**6. Rollback derives `owner_did` from `record.owner.0` instead of the header — good.** But confirm `record.owner_dilithium_pk` came from a state that was authenticated at registration. If registration accepted a forged PK (see #2759 audit), rollback trusts it and lets the forger roll back.

**7. `x-owner-did` header still exists elsewhere?** grep for `x-owner-did` — if any other handler reads it and gates auth on the value, that's still the spoofable path. This PR only fixed rollback.

**8. Dead-fallback in simple registration path** (line ~1240):
```rust
owner_signature_hex: String::new(),
registration_timestamp: 0,
registration_fee_whole: 10,
```
This constructs a `DomainRegistrationRequest` with empty signature. `verify_registration_proof` from #2759 rejects `owner_signature_hex.is_empty()`. So this path is now DEAD — all registrations via the "simple" endpoint fail verification. Either the endpoint is unused and should be removed, or it needs its own signing plumbing. Don't leave dead-broken paths.

**9. `signing_hash` for domain system txs.** `tx.signing_hash()` — good, canonical. But confirm `signing_hash()` on `Transaction` with `transaction_type: DomainRegistration/DomainUpdate` produces the right pre-image (not filtered/skipped for domain types).

### Test coverage

No new tests visible in the diff. Add:
- Rollback signature valid → succeeds; invalid sig → 403; missing sig → 400
- Signature over wrong version target CID → rejected
- Signature over wrong domain → rejected
- Registration without signature (simple path in #8) → currently fails; either test that it fails cleanly or remove the endpoint
- Transfer with `public_inputs=[]` → verify what actually happens (see #5)

### Verdict

Direction is right. Blockers before merge:
- #1: confirm identity_manager authenticates before returning `private_key`, OR move signing to the client and remove the server-side sk-copy
- #3, #5: cross-check with #2751 and #2759 that the composed system doesn't leak — this PR looks isolated but the interactions matter
- #8: dead simple-registration path — resolve before merge
- #2: pull chain_id from protocol params

Otherwise the fix is a substantial improvement over the zeroed-sig status quo.

PR #2762, opened 2026-07-01
security(audit-17): remove dead handle_mint_sov admin handler

umwelt:
**MEDIUM #17** — Dead `handle_mint_sov` admin mint handler (route already removed).

Replaces monolithic #2749.


umwelt:
## Security review

Deletes the `handle_mint_sov` admin-mint handler. Route was already unregistered; this removes the dead function body.

### Security concerns

**1. Dead code with a live capability.** The function creates a `TokenMint` with all-zero signature and calls `add_system_transaction(..., "admin_sov_mint")`. If the route were re-registered accidentally (or via a copy-paste of the code into another handler), the caller could mint arbitrary SOV to any wallet with no signature. Removing this correctly. Verify:
- No other handler references `handle_mint_sov` (grep)
- No documentation, README, mobile app, or postman collection references a `POST /wallet/mint` or equivalent endpoint that would then 404 and confuse operators. If it's referenced, remove those references too.

**2. `add_system_transaction(..., "admin_sov_mint")` originator.** With #2751's `validate_system_injection` merged, an unsigned `TokenMint` with originator `"admin_sov_mint"` would be REJECTED (not treasury, not pouw). Confirm the gate catches this even if the code path resurrects. Belt-and-suspenders: also add `"admin_sov_mint"` to a hard block-list in `validate_system_injection`, so if someone accidentally reintroduces the handler, injection still fails at the gate.

**3. Verify the route is unregistered.** PR body says "route already removed". Grep for `.route(...handle_mint_sov...)` or similar wire-up. If the router still declares the endpoint but the impl is gone, it'll fail to compile — good. If the router is separately registered, confirm.

**4. Anything else refer to the mint helper functions used here?** `lib_types::sov::atoms`, `lib_blockchain::contracts::utils::generate_lib_token_id`, `TokenMintData` — those are still valid crate APIs and shouldn't be touched. This delete is scoped correctly.

### Test coverage

No test coverage needed for a dead-code deletion IF (a) route is already unroutable, (b) no other code invokes the function. Grep to confirm.

### Verdict

Safe to merge. Recommend adding the `admin_sov_mint` originator string to a rejection list in #2751 as belt-and-suspenders — makes accidental reintroduction fail-closed even if this handler comes back.

PR #2763, opened 2026-07-01
security(audit-9): web4 client signs domain operations

umwelt:
Complements audit **#9** domain proof verification in lib-network.

Replaces monolithic #2749.

### DIFF
--- a/lib-network/src/web4/client.rs
+++ b/lib-network/src/web4/client.rs
@@ -734,15 +734,39 @@ impl Web4Client {
         new_manifest_cid: &str,
         expected_previous_cid: &str,
     ) -> Result<serde_json::Value> {
+        let private_key = self
+            .identity
+            .private_key
+            .as_ref()
+            .ok_or_else(|| {
+                anyhow!(
+                    "Domain update requires owner signing key; identity has no private key. \
+                     Use lib-client `build_domain_update_request` or zhtp-cli instead."
+                )
+            })?;
+
         let timestamp = std::time::SystemTime::now()
             .duration_since(std::time::UNIX_EPOCH)?
             .as_secs();
+
+        use super::domain_signing::domain_update_signing_message;
+        use lib_crypto::{KeyPair, sign_message};
+
+        let keypair = KeyPair::from_private_key(private_key)
+            .context("Failed to reconstruct owner signing keypair")?;
+        let message = domain_update_signing_message(
+            domain,
+            expected_previous_cid,
+            new_manifest_cid,
+            timestamp,
+        );
+        let signature = sign_message(&keypair, &message).context("Failed to sign domain update")?;
 
         let body = serde_json::json!({
             "domain": domain,
             "new_manifest_cid": new_manifest_cid,
             "expected_previous_manifest_cid": expected_previous_cid,
-            "signature": "", // TODO: Sign with identity
+            "signature": hex::encode(&signature.signature),
             "timestamp": timestamp,
         });
 


umwelt:
## Security review

`lib-network/src/web4/client.rs::update_domain` (or similar) now signs the update request body with the caller's Dilithium keypair over `domain|expected_previous_cid|new_manifest_cid|timestamp`, and populates `body.signature` with the hex-encoded signature. Previously `signature: ""` (empty).

### Security concerns

**1. Client-side private key from `self.identity.private_key`.** The Web4Client is a client-side abstraction — carrying `sk` in memory is expected. Confirm the client is only ever instantiated in trusted contexts (user's own binary, CLI, mobile app that owns this identity). Do NOT construct `Web4Client` from a server-side handler that received the sk over the network.

**2. Error message leaks structural hint.**
```rust
.ok_or_else(|| anyhow!(
    "Domain update requires owner signing key; identity has no private key. \
     Use lib-client `build_domain_update_request` or zhtp-cli instead."
))
```
Fine for client-side, not so fine if this ever bubbles up server-side. The suggestion `Use lib-client build_domain_update_request` hints at an alternative path — if that path is unsigned or has different gates, someone reading this could try to route around the check. Confirm `build_domain_update_request` from lib-client actually enforces signing.

**3. Message layout — `domain_update_signing_message(domain, expected_previous_cid, new_manifest_cid, timestamp)`.** Not in this diff. Confirm:
- Domain-separated (prefixes something like `"ZHTP:web4_domain_update_v1"`)
- Fixed byte layout (length prefixes on variable-length strings)
- Verifier on the SERVER SIDE constructs the identical bytes — otherwise sig fails
- Both endpoints agree on encoding of CID strings (utf-8? normalized?)

**4. No nonce → replay within timestamp window.** If server enforces a 5-minute window (per #2759 pattern), same signature replays until window expires. This mostly relies on `expected_previous_cid` being unique — if the domain state advances past `expected_previous_cid`, replay fails at the CAS check. Confirm the server does check `expected_previous_cid == current_cid` before applying. If not, this is a lost-update vector.

**5. `timestamp` is server-local time on the client.** Clock skew could cause the server's 5-minute window check to reject a legit update. Not a security concern, an availability one. Bound the workflow: on rejection, client retries once with fresh timestamp.

**6. `KeyPair::from_private_key(private_key)`.** What does this do internally?
- Derive pub key from sk? If yes: does it validate against `self.identity.public_key`? An accidentally-mismatched (sk,pk) pair would silently sign with a key the server doesn't recognize.
- If the pk-derivation step matches, fine. Add a debug_assert or explicit check.

**7. `signature.signature` field.** The wire format sends `hex::encode(&signature.signature)`. Signature struct has multiple fields (per #2758 review: `signature`, `public_key`, `algorithm`, `timestamp`). Server-side verify only gets the raw signature bytes — it must (a) look up owner PK from the domain record, (b) hash the same message, (c) verify. Confirm server does NOT trust a self-declared public_key from the body.

### Test coverage

Diff is 25 lines and adds no test. Add:
- Sign with valid sk → server accepts (integration test with real server)
- Sign with sk that doesn't match the domain's on-chain owner PK → 403
- Signature over wrong `new_manifest_cid` → rejected
- Signature over wrong `expected_previous_cid` → rejected
- Missing/empty sk on client → helpful error, no submission
- Server round-trip: replay same signature within window → rejected on 2nd attempt (nonce/expected-cid check)

### Verdict

Safe to merge from the client-side surface. The correctness depends on server-side verification in #2759/#2761 doing its job. Blocker: confirm `domain_update_signing_message` implementations agree byte-for-byte between client (this PR) and server (in `lib-network/src/web4/domain_signing.rs`, likely already merged). If they diverge, every domain update signature fails and the feature is broken.

Issue #2776, opened 2026-07-03
BUBL rewards disabled after GENESIS-4: TokenCreation not deployed (blocks /api/v1/rewards/*)


umwelt:
## Problem

After GENESIS-4 sled wipe + restart (2026-07-03), g1 logs at startup:

```
rewards: BUBL token contract f5aff42a not found on chain; rewards endpoint disabled
Rewards: ZHTP_REWARDS_TREASURY_KEYSTORE unset, invalid, or holds 0 BUBL — /api/v1/rewards/* will return 503
```

`ZHTP_REWARDS_TREASURY_KEYSTORE=/opt/zhtp/keystores/bubl-creator` is set on g1, but BUBL `TokenCreation` was never replayed on the wiped chain. All `/api/v1/rewards/*` endpoints return 503.

## Expected token id

`f5aff42a31e17656ecab4b01cc2aea15025d813a3109c98b8f1a55378802f82d` (deterministic from name+symbol \"Bubble\"/\"BUBL\")

## Decimal policy

BUBL rewards use 18-decimal atoms (`ATOM_18 = 10^18`) in `zhtp/src/api/handlers/rewards/mod.rs`. Must match `TokenCreation` supply units.

## Action

Execute GENESIS-6 Phase 7 seeding: founding `TokenCreation` for BUBL (+ verify creator keystore balance) per `docs/protocol/genesis-3-testnet-reset.md`.

## Related

#2734, #2727

umwelt:
Resolved on testnet via D3 legacy BUBL rewards module upgrade (PR #2845, merged ecd6f2bd).

- AssetModuleUpgrade committed ~34693 with RewardsModuleState + policy
- rewards_activation.toml active on g1/g2/g3
- Welcome claim smoke passed at height 34959 (100 BUBL, no halt)
- Integration tests added for RewardClaim mempool dedup (#2844 review follow-up)

Issue #2768, opened 2026-07-03
identity/register returns success when chain enqueue fails


umwelt:
## Problem

`handle_register_client_identity` returns HTTP 200 with `"status": "queued"` after `register_external_citizen_identity` completes, even when `add_system_transaction` for the identity tx fails (logged as non-fatal) or wallet `register_wallet` calls fail.

In-memory `IdentityManager` entry and wallet IDs are already created, so the client believes registration is on-chain when it is not.

## Location

`zhtp/src/api/handlers/identity/mod.rs` (~2091–2295)

## Expected

- Treat identity system-tx enqueue failure as a hard error, or return explicit partial-failure status
- At minimum, return an error when `blockchain_tx_hash` is `None` after the chain path was attempted

## Context

Found during code review of ClientIdentityRegistration fix (GENESIS-4 mobile registration). Prior `supertramp` session showed wallets/SOV committed but identity missing — this false-success path exacerbates partial state.

## Related

- GENESIS-4 registration fix
- #2727

umwelt:
Validated as still relevant; core fix already on `development` via #2779. Residual hardening + tests in PR #2920 (closes this issue on merge).

umwelt:
Confirmed closed via #2920. Testnet deploy completed 2026-07-19 (validators g1–g3 + gateway; binary `508d992a`, md5 `b57b2880…`).

See #2920 for full deploy notes.

PR #2779, opened 2026-07-03
fix(consensus): testnet recovery — quorum fork, mempool livelock, BUBL seed

umwelt:
## Summary

Unblocks coordinated testnet reset (g1–g3) after quorum-proof fork at H=53 and mempool livelock that prevented BUBL `TokenCreation` from committing.

- **Consensus (P0):** Finalize only with buildable commit attestations; retry `maybe_finalize` on commit timeout; skip empty `valid_value` reproposal when mempool has work; release stale empty lock so validators can prevote mempool-inclusive blocks
- **GENESIS-6:** `seed_founding_dao` tool + BUBL token_id alignment in rewards handler
- **GENESIS-4:** Clean-slate genesis + replay gate alignment
- **Identity:** Harden client registration (#2768–#2776) before reset

## Test plan

- [x] `cargo test -p lib-consensus --lib` (lock rule + CE-S3)
- [x] Live testnet g1–g3: identical proposal hashes, 3-attestation quorum proofs, no `Invalid previous hash`
- [x] BUBL founding tx mined at block 22; token list shows 1B supply; creator balance 800M BUBL
- [x] Rewards orchestrator running with `ZHTP_REWARDS_TREASURY_KEYSTORE` on g1

## Deployed binary

`zhtp` MD5 `57e2e01232cb2293e9f7946b0ab63250` (g1–g3, Jul 3 2026)

## Commits

1. `c0650993` fix(identity): harden client registration before testnet reset
2. `98358224` chore(genesis): GENESIS-4 clean slate + replay gate alignment
3. `160759f3` feat(genesis-6): founding DAO token seed tool + BUBL id alignment
4. `e6c10015` fix(consensus): prevent quorum-proof fork and commit-timeout wedge
5. `c71fd33c` fix(consensus): release stale empty lock when mempool has pending txs


umwelt:
## Security review — PR #2779

Scope: consensus safety (quorum-proof + lock rules), client identity registration hardening, `add_system_transaction` gating, BUBL seed tool, welcome-bonus mint refactor, client-signed domain system tx.

### Strong positive changes

- **`count_buildable_commits_for` (storage.rs)** — filters commit votes by exact 4595-byte Dilithium5 signature length + dedups by validator id. Matches the filter in `build_quorum_proof`, so `commit_count ≥ 2f+1` can no longer be true while the built proof has fewer attestations. That was the exact shape of the H=53 fork; the fix is minimal and well-scoped.
- **Post-build attestation-count refuse (state_machine.rs:1262–1271)** — if `quorum_proof.attestations.len() < bft_quorum_count`, we refuse to apply. Correct belt-and-braces defence: even if a future counting regression sneaks in, we'll fail closed rather than fork.
- **Commit-timeout retry `maybe_finalize` (state_machine.rs:2712–2735)** — mirrors the PreCommit arm and cleanly handles "local commit vote was the final one but arrived after `Commit` state entry." No behavioural change if `maybe_finalize` is a no-op; small blast radius.
- **Throwaway server keypair removed from `register_identity`** — the old path signed the identity tx with a per-request `generate_keypair()` and shipped that as the on-chain signature; the actual key-ownership proof (`ZHTP_REGISTER:{timestamp}` → `verify_signature` at mod.rs:2007) is now the sole authoritative check, and the on-chain tx is injected via `add_system_transaction` under `ClientIdentityRegistration`. This is the correct model.
- **`SystemOriginator` enum + `validate_system_injection` gate (blockchain.rs:2468–)** — typed originator instead of `&'static str`, plus per-class structural checks (empty-sig, PoUW memo shape, treasury-only funded wallet init, DID ↔ pubkey derivation match for client identity). Exactly the audit-follow-up class of hardening we want in front of a bypass entrypoint.
- **`Web4DomainRegister/Update` now client-signed** (`attach_client_domain_tx_signature`) and gated at `validate_system_injection` against empty sig. Kills the previous "server injects zero-signature domain tx" pattern.
- **Wallet `handle_provision_wallet` now Council/System-role gated** (wallet/mod.rs:1544). Previously public.
- **PoUW memo canonicalisation (`validate_pouw_mint_memo`)** — parses `pouw:mint:{recipient_hex}:{amount}` and cross-checks against `mint_data.to` / `mint_data.amount`. Prevents PoUW originator from smuggling a mismatched recipient into an unsigned mint.
- **Registration timestamp future-skew guard** (mod.rs:1919) — 60s future cap paired with existing 300s stale cap.
- **Rate limiter now applied to `register_identity`** (mod.rs:1918).

### P0 concerns

**1. `cached_proposal_has_no_transactions` fails toward unlock on decode error**
`proofs.rs`:
```rust
match provider.decode_block_data(block_data).await {
    Ok((tx_count, _)) => tx_count == 0,
    Err(_) => true,   // ← treats undecodable block as "empty"
}
```
Also the `provider.is_none()` and `block_data.is_empty()` cases return `true`. `maybe_clear_stale_empty_lock_for_incoming` then treats "empty" as "safe to unlock." A validator whose decoder momentarily errors (transient sled I/O, version mismatch after upgrade, corrupt cache entry) will release its lock on a valid block it just cannot read. Recommend flipping the error direction to `false` (preserve the lock) — the cost of a spurious "keep-locked" is one more round; the cost of a spurious "unlock" is a divergent prevote round with peers who still hold the lock.

**2. Lock-release breaks Tendermint safety invariant under mempool divergence**
`maybe_clear_stale_empty_lock_for_incoming` unlocks whenever *local* mempool has pending tx AND the currently-locked cached block has zero txs AND the incoming carries txs. This is a local, non-consensus signal. Under `n=3, f=0` (current testnet) it is safe by construction. Under `n≥4` with an honest-but-partitioned mempool state (validator A sees a client tx that B/C don't yet), A will unlock and prevote the mempool-inclusive block while B/C stay locked on empty → split prevote → new commit quorum impossible until skip-round → in the worst case a locked-conflict two-round window. This is a real Tendermint safety property (locks may only be released via the +2/3 valid-round rule); please gate the new rule behind a `n <= 3` (or explicit "single-validator-diverged" is impossible) config, and document the assumption. At minimum, add a test that with `n=4`, one node with a "witnessed" mempool tx does not unlock while the other three see the same locked-empty proposal.

### P1 concerns

**3. Display-name uniqueness is check-then-inject with a race window**
mod.rs:2174 does `identity_display_name_taken` under a *read* lock, releases it, and later re-acquires the *write* lock to inject via `add_system_transaction`. `validate_system_injection` re-checks DID ↔ pubkey and ownership_proof — but not display-name uniqueness. Two concurrent requests for the same display name with different DIDs both pass the read-lock check, then both enqueue. Whether the committing executor rejects duplicate `display_name` on apply is what makes this benign; if it does not, both wallets are created and later lookups become ambiguous. Recommend: move `identity_display_name_taken` inside the write lock (right before `add_system_transaction`), or add a pending-name intent set the display-name check consults.

**4. `identity_data.public_key` cross-check uses `unwrap_or([0u8; 2592])`**
`blockchain.rs`:2584 —
```rust
if sig_pk != identity_data.public_key.as_slice().try_into().unwrap_or([0u8; 2592]) { … }
```
The length was already verified two lines up, so `try_into` cannot fail today. If someone later refactors the length check away, `unwrap_or([0u8; 2592])` degrades silently: a signature pubkey of all-zeros against an all-zeros identity payload would slip through. Prefer `try_into().expect("length checked above")` or fold the check into a single `TryFrom` binding.

**5. Shadow-index mutation survives commit failure**
mod.rs:  the write-lock path calls `add_system_transaction` and, on success, immediately does `insert_identity_shadow` + `identity_blocks.insert(..., pending_height)`. If the transaction is later dropped from the mempool (fee sweep, reorg, block-full drop, replay), those in-memory registries retain the "as if committed" entries. The queue-failure branch does `identity_manager.remove_identity`, but there is no rollback wired to actual block-commit failure. Not exploitable given DID = f(pubkey), but it does mean the API will falsely report "exists" to future callers after a dropped tx. Consider deferring the shadow insert to executor apply.

**6. `TokenMint` from any treasury originator has no memo/amount cross-check**
`validate_system_injection` for `TokenMint` only runs `validate_pouw_mint_memo` on `PouwMint`/`PouwReward`. Treasury classes (`is_treasury()` — `TreasuryWalletBootstrap`, `TreasuryKernel`, `GenesisBootstrap`, …) are early-returned with no payload/memo validation. `queue_welcome_bonus_token_mint` under `TreasuryWalletBootstrap` is the only current caller and is server-driven, so it is fine in practice — but the door is now wide open for any future treasury-originator call site to inject arbitrary recipient/amount TokenMints. Recommend at least: (a) require `WELCOME_BONUS_V1:{wallet_hex}` memo shape and cross-check against `mint_data.to`; (b) cap `mint_data.amount` under a fixed per-block ceiling for the bootstrap originator specifically.

**7. `seed_founding_dao.rs`: fee=0 permitted; supply/ratio hardcoded**
- `tx.fee = if fee == 0 { 0 } else { fee.max(DEFAULT_TOKEN_CREATION_FEE) }` — a `--fee 0` CLI invocation produces a zero-fee `TokenCreation` that will only be accepted while mempool policy is in audit-only mode. Fine for the coordinated seed window but worth documenting that this tool cannot be used post-hardening.
- `treasury_allocation_bps: 2_000` is hardcoded twice (build path + display print). If the split is intended to be tool-flag-controlled per-token (BUBL vs CBE), add `--treasury-bps`; if it is protocol policy, define it as a constant in `TokenCreationPayloadV1` so client and executor agree at compile time.
- `default_dilithium_pk() → [0u8; 2592]` — silently defaults an absent public key to all zeros. If the identity re-derivation from `dilithium_sk` fails for any reason (schema drift), the tool will happily produce an invalid TokenCreation signed with a mismatched pubkey. Prefer requiring `dilithium_pk` or fatally erroring on missing.

**8. `mempool_has_pending_transactions` treats decode-error as "yes, pending"**
`proofs.rs`: `Err(_) => !data.is_empty()`. On a partial/corrupt pending-tx blob the engine will insist mempool has work and refuse to reuse `valid_value`. Reciprocal to concern (1): the two branches now disagree about which side of "unknown" to fail toward. Align them: either both fail toward "cannot know → don't take the new action," or gate on `provider.is_ready()` more strictly.

### Nit / low priority

- `system_originator.rs` file missing trailing newline (both `\ No newline at end of file` markers).
- `SystemOriginator::allows_unsigned_domain_tx` returns `false` unconditionally. Either delete the method or add a doc comment explaining why it exists (looks like reserved API surface for a future exception).
- `attach_client_domain_tx_signature` accepts `owner_dilithium_pk: &[u8]` and immediately `try_into::<[u8; 2592]>()`. Tighten the type at the boundary instead of validating twice.
- `identity_manager.remove_identity(&identity_id)` on failure races with any concurrent reader that already got a handle. Not a security bug, but a follow-up.
- `WELCOME_BONUS_V1:{wallet_hex}` memo pattern is defined in string form in two call sites (identity handler + wallet handler). Move to a shared const so treasury-mint recipient/memo cross-check (concern 6) has a single source of truth.

### Verdict

The consensus fixes are targeted, well-scoped, and demonstrably reproduce the observed H=53 failure mode; the identity-registration hardening removes a real "server signs on client's behalf" foot-gun. The lock-release rule (concern 2) is the one item I want addressed before this pattern is generalised beyond `n=3` — the rest are P1/nits that can be follow-up PRs. Approve for the coordinated reset scope; open a tracking issue for concerns 1, 2, 3, 6 before we scale validator count.

umwelt:
P0 #1 applied in 5c88d5aa: `cached_proposal_has_no_transactions` now fails closed (`Err(_) => false`, missing provider => false). P0 #2 documented as n=3 testnet assumption on `maybe_clear_stale_empty_lock_for_incoming`. lib-consensus tests green locally.

Issue #2782, opened 2026-07-04
SA-2: GET /api/v1/assets/* discovery API


umwelt:
Add unified asset discovery API backed by SA-1 projection.

**Endpoints:**
- `GET /api/v1/assets`
- `GET /api/v1/assets/{id}`
- `GET /api/v1/assets/{id}/interface`
- `GET /api/v1/assets/{id}/balances/{did}`

Keep `/api/v1/token/*` as shim. Depends on SA-1.

Parent epic: #2780
Design: docs/arch/sovereign-asset.md §9 (SA-2)

umwelt:
**Closed — acceptance criteria met (2026-07-14 audit).**

### What shipped
- `GET /api/v1/assets` — list via `iter_sovereign_assets()` projection
- `GET /api/v1/assets/{id}` — detail + manifest resolve + governance/rewards sub-status
- `GET /api/v1/assets/{id}/interface` — module interface summary
- `GET /api/v1/assets/{id}/balances/{address}` — per-DID balance lookup

### Where
- Handlers: `zhtp/src/api/handlers/assets/mod.rs`
- Router: `zhtp/src/unified_server.rs` (`/api/v1/assets`)
- Read projection: `lib-blockchain/src/blockchain/contracts.rs::iter_sovereign_assets()` (native `assets/` + legacy shim)
- Tests: `zhtp/src/api/handlers/assets/mod.rs` (share link, interface, manifest hash)

### Out of scope (tracked elsewhere)
- `/api/v1/token/*` shim removal → **SA-8** (#2788)
- `POST /api/v1/assets/launch` → **SA-3** (#2783)

Issue #2780, opened 2026-07-04
Epic: Sovereign Asset — unified DAO/token primitive


umwelt:
## Summary

Implement the **Sovereign Asset** primitive per [docs/arch/sovereign-asset.md](docs/arch/sovereign-asset.md): one shareable on-chain asset type with optional modules (curve, rewards, governance), deploy-address `asset_id`, mandatory manifest, and hot-wallet rewards delegation.

Replaces fragmented `TokenContract` + `BondingCurveToken` + ad-hoc DAO wiring.

## Child issues (PR DAG)

| PR | Issue | Scope |
|----|-------|-------|
| SA-1 | #2781 | Types + sled layout + read projection |
| SA-2 | #2782 | `GET /api/v1/assets/*` discovery API |
| SA-3 | #2783 | `AssetLaunch` tx + executor |
| SA-4 | #2784 | Rewards module + delegate + validator handler |
| SA-5 | #2785 | Curve module as asset sub-state |
| SA-6 | #2786 | `AssetModuleUpgrade` + governance |
| SA-7 | #2787 | `AssetManifestUpdate` + DHT pin validation |
| SA-8 | #2788 | Deprecate legacy `/api/v1/token/*` |

## Locked decisions

- `asset_id` = launch tx hash (deploy-address style)
- Module upgrades allowed; governance = single or multisig (default M = floor(N/2)+1)
- Creator loses privileged rights after governance transfer
- Rewards via on-chain hot-wallet delegate

## Deploy gate

- **SA-1 + SA-2**: safe to deploy (read-only projection + API)
- **SA-3+**: requires testnet BUBL/CBE re-launch via `AssetLaunch` + validator restart

umwelt:
**Epic status audit (2026-07-14)**

| PR | Issue | Status |
|----|-------|--------|
| SA-1 | #2781 | ✅ Closed |
| SA-2 | #2782 | ✅ Closing now — discovery API done |
| SA-3 | #2783 | 🟡 Partial — AssetLaunch shipped; TokenCreation/ops migration not |
| SA-4 | #2784 | 🟡 Partial — module + txs; treasury env hack remains |
| SA-5 | #2785 | 🟡 Partial — sled/types only; curve execution still CBE-global |
| SA-6 | #2786 | 🟡 Partial — governance module; proof verify + 80k gate |
| SA-7 | #2787 | 🟡 Partial — manifest tx + API read; no consensus DHT gate |
[...]

PR #2789, opened 2026-07-04
feat(assets): Sovereign Asset SA-1/SA-2 — projection + discovery API

### LINKED ISSUE #2781: SA-1: SovereignAsset types + sled layout + read projection
Implement `SovereignAsset` Rust types, module headers, and a **read-only projection** from legacy `TokenContract` / `BondingCurveToken` → unified asset view.

**Deliverables:**
- `lib-blockchain/src/contracts/sovereign_asset/` types per ADR §3–4
- `project_legacy_token()` / `project_legacy_curve()` helpers
- `Blockchain::iter_sovereign_asset_projections()` facade
- Unit tests (BUBL-class fixed token, CBE-class curve token)

**No new tx types.** No sled write path changes yet.

Parent epic: #2780
Design: docs/arch/sovereign-asset.md §9 (SA-1)

umwelt:
## Summary

Implements the first two steps of the Sovereign Asset program ([#2780](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2780)):

- **SA-1 ([#2781](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2781)):** `SovereignAsset` types + read-only projection from legacy `TokenContract` / `BondingCurveToken`
- **SA-2 ([#2782](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2782)):** `GET /api/v1/assets/*` discovery API

Design: [docs/arch/sovereign-asset.md](docs/arch/sovereign-asset.md)

## API

| Endpoint | Purpose |
|----------|---------|
| `GET /api/v1/assets` | List projected assets (module bitmask, supply mode) |
| `GET /api/v1/assets/{id}` | Full on-chain projection |
| `GET /api/v1/assets/{id}/interface` | Wallet tx_kinds shortcut |
| `GET /api/v1/assets/{id}/balances/{did}` | Balance lookup |

Legacy `/api/v1/token/*` unchanged (shim until SA-8).

## Deploy gate

Read-only projection — **safe to deploy** without re-launching BUBL/CBE. `AssetLaunch` tx (SA-3) is a separate PR.

## Test plan

- [x] `cargo test -p lib-blockchain --features contracts project_bubl_class_token --lib`
- [x] `cargo check -p zhtp`
- [ ] Post-deploy: `GET /api/v1/assets` on g2 returns BUBL entry

Closes #2781, #2782.


umwelt:
## Security review — PR #2789

Scope: SA-1/SA-2 delta on top of #2779 (already reviewed at #2779). This review covers only the new commits `37849a45` (identity read-your-writes) and `31b04143` (SovereignAsset types, projection, discovery API); pre-existing #2779 concerns still apply on merge.

### `37849a45` — identity read-your-writes in metadata batch

Adds `tree_lookup` pre-checks in `SledStore::get_identity` and `SledStore::get_identity_metadata` so a staged write in the open metadata batch is visible to a same-block reader (fixes the `IdentityRegistration + IdentityUpdate` in one block halt described in the commit message).

**Correct.** The pattern mirrors the existing token-balance write-through (CR #2658 #2). No security concern.

Nits:
- `self.tx_batch.lock().unwrap()` will panic on mutex poisoning. Consistent with the rest of `sled_store.rs`, so not a blocker — but any single panic between `begin_metadata_write()` and `commit_metadata_write()` now surfaces as a poisoned mutex that bricks all subsequent identity reads. Preferable: match on `.lock()` and `warn!` + fall through to committed sled.
- The new `test_identity_metadata_write_read_your_writes` test never exercises the "batch closed → falls through to `identities.get`" path. Add a same-DID read after `commit_metadata_write()` to lock in that the post-commit path still works.

### `31b04143` — SovereignAsset types, projection, discovery API

#### Correctness

**1. `project_from_token_contract` misreports uncapped tokens as `SupplyMode::Fixed`**
`projection.rs:47–52` unconditionally sets `SupplyMode::Fixed` and, when `token.max_supply == u128::MAX` (the codebase's elastic-supply sentinel — see `tokens/core.rs:170` "No fixed cap — elastic supply = SOV staked"), coerces `max_supply` to `total_supply.max(1)`. Any legacy kernel-controlled or elastic-cap `TokenContract` (SOV native's shape) will be projected as a fixed-cap asset whose displayed cap equals whatever supply happens to be at read time. Wallets/UIs relying on `supply_mode` to disable "how much is left to mint" will be wrong. Recommend: `supply_mode = if token.max_supply == u128::MAX { Elastic } else { Fixed }` and keep `max_supply = u128::MAX` untouched (or `0` for "unbounded", with a documented sentinel).

**2. `legacy_interface` returns tx_kinds the chain does not accept**
`assets/mod.rs:84–98` emits `["AssetTransfer", "CurveBuy", "CurveSell", "RewardsClaim"]`. Those names are the SA-3+ vocabulary and are not valid `TransactionType` values today. A wallet that reads `/api/v1/assets/{id}/interface` and dispatches on `tx_kinds` will build a tx the mempool rejects. The `note` field ("Legacy projection — full manifest arrives in SA-7") is not machine-readable, so wallets *will* consume this. Recommend: for `id_source = LegacyTokenId`, emit the actual legacy names — `["TokenTransfer"]`, `["BondingCurveBuy", "BondingCurveSell"]` — or omit `tx_kinds` entirely (empty array + note) so a client can hard-fail rather than build a bad tx.

**3. `handle_balance` silently ignores sled errors**
`assets/mod.rs:213`: `let balance = bc.token_balance(&asset_id, &key_id).unwrap_or(0);` A sled read failure surfaces as balance 0. `token_balance`'s own doc explicitly says sled failures must be propagated: "we do NOT silently fall back to a possibly-stale in-memory balance … exactly the class that leads to authorizing a transfer on a wrong balance" (contracts.rs:1097). This handler is read-only so no direct auth-of-transfer risk, but it *is* the surface a wallet uses to decide whether to *offer* a transfer — a false-0 will hide funds. Return `500` (or a `null` balance with a `stale` flag) on sled error rather than 0.

**4. `get_sovereign_asset` is O(N) per lookup**
`contracts.rs:1229–1236` calls `iter_sovereign_assets().into_iter().find(...)` for every single-id lookup. Each `iter_sovereign_assets` call does a full sled scan over `token_contracts` + a full `bonding_curve_registry.get_all()`. `/api/v1/assets/{id}` and `/api/v1/assets/{id}/interface` and `/api/v1/assets/{id}/balances/{did}` all pay this cost. Not exploitable at today's <10-asset scale, but the point of "unified discovery" is that it works when there are thousands. Cheap fix: `get_sovereign_asset` reads `token_contract` and `bonding_curve_registry` directly by id, projects, and merges without listing.

**5. `iter_sovereign_assets` `.max(1)` shifts curve total_supply**
`projection.rs:95` — `max_supply: curve.total_supply.max(1)`. If a bonding curve has legitimately zero total_supply pre-first-buy (curve phase, no mints yet), the projection reports `max_supply = 1` which lies about capacity. Prefer `curve.total_supply` unmodified (or make max_supply Option<u128>).

**6. `iter_token_contract_entries` silent in-memory fallback flows through to `/api/v1/assets`**
`iter_token_contract_entries` (contracts.rs:1178) logs and falls back to `self.token_contracts` on sled failure. The pre-existing behaviour is intentional for other callers, but `/api/v1/assets` is now a discovery API — a sled read error will silently serve the in-memory shadow (which the same file documents is often empty / stale after sled load). Recommend the assets handler explicitly opt into a "no-fallback" mode via a dedicated `iter_token_contract_entries_strict()` that returns `StorageResult`, and translate to 503.

#### API surface / auth

**7. Handler routing is public, no rate limit, no `Cache-Control`**
Consistent with `/api/v1/token/*`, so not a regression. But `/api/v1/assets/{id}/balances/{did}` accepts any 32-byte hex string and returns the balance — useful for enumeration attacks against a target DID's holdings if the query is unauthenticated. `/token` has the same shape, so this doesn't move the needle; flag only to track it in the broader public-read audit.

**8. `resolve_key_id` misparses DID with non-32-byte hex payload**
`assets/mod.rs:242–258` accepts `did:zhtp:{hex}`, `0x{hex}`, and plain `{hex}`. For a DID whose method-specific-id is not exactly 64 hex chars (32 bytes), it errors — good. But nothing enforces the DID prefix belongs to *this* chain's DID method; a caller passing `did:foo:{64-hex}` where the trailing hex happens to also be a valid `did:zhtp` key_id will get a real balance back. Low risk (any 32-byte hex is a valid key_id lookup regardless of prefix), but the API contract should either canonicalise to `did:zhtp:` or reject unknown DID methods.

**9. `handle_request` converts internal `anyhow::Error` to a generic error response**
`assets/mod.rs:262`—each inner handler returns `anyhow::Result` and any `?`/`ok_or_else` failure ("Asset not found", "Invalid asset_id hex", "asset_id must be 32 bytes") propagates as a runtime error rather than a proper `404` / `400`. Consumers will see 500-class errors for what should be 404 and 400. Wrap in `create_error_response` at the boundary.

#### Design / consistency

**10. `AssetIdSource::LegacyTokenId` and `AssetIdSource::LaunchTx` mint the same 32-byte type**
There is no distinguishing tag beyond the enum, so `handle_balance` blindly passes `asset_id` to `token_balance`. When SA-3 launches, a `LaunchTx` id will collide with 0 legacy tokens (near-zero probability), but the code path silently returns 0 for launch-tx-id lookups because balances are keyed by the legacy `token_id`, not the launch tx hash. Recommend the balance handler explicitly reject `LaunchTx` id_source with a `501 Not implemented until SA-3` for now — otherwise the endpoint will look correct today, look correct at SA-3, and be silently broken during the transition window.

**11. `AssetModuleFlags` has KERNEL (bit 3) and MARKET (bit 4) constants but no accessor**
`types.rs:34–35` defines the bits, no `has_kernel()` / `has_market()`. Either delete or wire.

**12. `sovereign_asset/mod.rs` missing trailing newline**
`\ No newline at end of file` in the diff.

**13. `AssetsHandler::new()` `tokio::task::block_in_place` inside `block_on`**
`assets/mod.rs:164–169` will panic if the runtime is single-threaded or if `new()` is called outside an active runtime context. Same shape as the other handlers in this repo so not a regression, but the constructor really should be `async fn new()` or take the blockchain arc as a parameter.

**14. `is_sov_native_token_id` short-circuits BUBL detection**
`projection.rs:16–18` treats both zero-id and `generate_lib_token_id()` (SOV) as "native, skip." Good. But BUBL detection at 40–45 uses `symbol.eq_ignore_ascii_case(BUBL_SYMBOL) || token.name == BUBL_NAME`. The `||` on name would let a token whose creator names it `"Bubble"` (case-sensitive exact match) but with a different symbol flip the `REWARDS` module bit. On SA-3+ this is fine (asset_id is launch-tx-derived), but pre-SA-3 there is no on-chain uniqueness gate that stops a copycat `TokenCreation` with `name = "Bubble"`. Recommend `symbol.eq_ignore_ascii_case(BUBL_SYMBOL) && name.eq_ignore_ascii_case(BUBL_NAME)` — both must match — so the projection doesn't attach the rewards module to a lookalike.

#### Tests

The one new test — `project_bubl_class_token` — asserts the happy path but doesn't cover:
- Elastic (`u128::MAX`) `max_supply` (concern 1)
- Same `token_id` appearing in both token and curve registries (`merge_curve_into_asset` path)
- Empty-supply curve (`max(1)` behaviour, concern 5)
- `AssetsHandler::handle_get` for a nonexistent id (should be 404, currently 500)

### Verdict

Read-only projection with no state mutation — the blast radius is bounded to "clients get confusing/wrong metadata." Concerns 1, 2, 8, 9 (misreported `supply_mode`, misleading `tx_kinds`, silent sled-error swallow, wrong-status errors) are the ones I'd hold merge on because they will land wallets in wrong dispatch paths the moment SA-3 lands. The rest are P1/nits.

Approve for staged rollout **after** concerns 1, 2 and 9 are addressed; the identity read-your-writes fix (`37849a45`) is a clean correctness win and can ship immediately either way.

PR #2792, opened 2026-07-06
fix(api): BUBL as DAO TokenCreation — rewards, balances, token list

umwelt:
## Summary

BUBL is a DAO `TokenCreation` token (GENESIS-6), not native SOV and not an `AssetLaunch` sovereign asset. This PR fixes the API/rewards path to use the canonical deterministic token id (`generate_custom_token_id("Bubble","BUBL")`).

## Changes

- **rewards**: Resolve BUBL from `TokenCreation` contract; authorize creator keystore only; drop sovereign `AssetLaunch` fallback; add `ZHTP_REWARDS_TOKEN_ID` (deprecated alias `ZHTP_REWARDS_ASSET_ID`)
- **token API**: Prefer DAO `token_contracts` over mistaken `AssetLaunch` rows with same symbol
- **balance_key**: `did:zhtp:{key_id}` resolves directly for custom token balances (creator 800M visible)
- **reset runbook**: Document `seed_founding_dao` (not `seed_asset_launch`) for BUBL

## Testnet

Already deployed to g1–g3 (`e2e22ebc…`). Verified:

```
rewards: BUBL signer loaded — token_id=6948e43dea1fa98a balance=800000000000000000000000000
```

No testnet reset required for this fix — canonical BUBL is already on-chain.

## Commits

- fix(api): surface AssetLaunch BUBL in legacy token endpoints (hide when DAO token exists)
- fix(rewards): treat BUBL as GENESIS-6 TokenCreation DAO token
- fix(api): resolve did:zhtp directly for custom token balances
- fix(rewards): BUBL is DAO TokenCreation only — drop sovereign fallback

### DIFF
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -81,6 +81,71 @@ impl Blockchain {
         call.contract_type == crate::types::ContractType::Token && call.method == "transfer"
     }
 
+    /// Repair TokenCreation allocations missing from the `token_balances` sled tree.
+    ///
+    /// `load_from_store` skips token-tx replay (#2637). When the legacy
+    /// `process_token_transactions` path persisted contract metadata without
+    /// crediting the balance tree, creator/treasury read 0 despite
+    /// `total_supply > 0`. Idempotent: only repairs tokens with zero sled
+    /// holders by replaying initial allocations from committed TokenCreation txs.
+    pub fn repair_missing_token_creation_balances(
+        &self,
+        store: &dyn crate::storage::BlockchainStore,
+    ) -> crate::storage::StorageResult<usize> {
+        use crate::storage::{Address, TokenId};
+        use crate::transaction::token_creation::TokenCreationPayloadV1;
+
+        let mut repairs: Vec<(TokenId, Address, u128)> = Vec::new();
+        let mut repaired_tokens = std::collections::HashSet::new();
+
+        let blocks: Vec<crate::block::Block> = self.iter_blocks().collect();
+        for block in &blocks {
+            for tx in &block.transactions {
+                if tx.transaction_type != TransactionType::TokenCreation {
+                    continue;
+                }
+                let payload = match TokenCreationPayloadV1::decode_memo(&tx.memo) {
+                    Ok(p) => p,
+                    Err(_) => continue,
+                };
+                let token_id_bytes = crate::contracts::utils::generate_custom_token_id(
+                    &payload.name,
+                    &payload.symbol,
+                );
+                if !repaired_tokens.insert(token_id_bytes) {
+                    continue;
+                }
+                let token_id = TokenId::new(token_id_bytes);
+                if store.count_token_holders(&token_id)? > 0 {
+                    continue;
+                }
+                let Some(contract) = store.get_token_contract(&token_id)? else {
+                    continue;
+                };
+                if contract.total_supply == 0 {
+                    continue;
+                }
+
+                let (creator_alloc, treasury_alloc) = payload.split_initial_supply();
+                if creator_alloc > 0 {
+                    repairs.push((
+                        token_id,
+                        Address::new(tx.signature.public_key.key_id),
+                        creator_alloc,
+                    ));
+                }
+                if treasury_alloc > 0 {
+                    repairs.push((token_id, Address::new(payload.treasury_recipient), treasury_alloc));
+                }
+            }
+        }
+
+        if repairs.is_empty() {
+            return Ok(0);
+        }
+        store.force_set_token_balances(&repairs)
+    }
+
     /// Rebuild the PoUW reward mint index from all in-memory blocks.
     ///
     /// Scans every block for `TokenMint` transactions carrying a `pouw:mint:`
@@ -743,6 +808,32 @@ impl Blockchain {
                         let store_ref: &dyn crate::storage::BlockchainStore = store.as_ref();
                         if let Err(e) = store_ref.put_token_contract(&token) {
                             warn!("Failed to persist token contract after creation: {}", e);
+                        } else {
+                            let token_storage_id = crate::storage::TokenId(token_id);
+                            let creator_addr =
+                                crate::storage::Address::new(creator.key_id);
+                            let treasury_addr =
+                                crate::storage::Address::new(payload.treasury_recipient);
+                            if let Err(e) = store_ref.set_token_balance(
+                                &token_storage_id,
+                                &creator_addr,
+                                creator_allocation,
+                            ) {
+                                warn!(
+                                    "Failed to persist creator balance after TokenCreation: {}",
+                                    e
+                                );
+                            }
+                            if let Err(e) = store_ref.set_token_balance(
+                                &token_storage_id,
+                                &treasury_addr,
+                                treasury_allocation,
+                            ) {
+                                warn!(
+                                    "Failed to persist treasury balance after TokenCreation: {}",
+                                    e
+                                );
+                            }
                         }
                     }
                 }
--- a/lib-blockchain/src/blockchain/contracts.rs
+++ b/lib-blockchain/src/blockchain/contracts.rs
@@ -86,8 +86,8 @@ impl Blockchain {
     /// `load_from_store` skips token-tx replay (#2637). When the legacy
     /// `process_token_transactions` path persisted contract metadata without
     /// crediting the balance tree, creator/treasury read 0 despite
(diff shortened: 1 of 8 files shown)


umwelt:
## Security review — PR #2792

Scope: BUBL treated as a DAO `TokenCreation` token (GENESIS-6), not native SOV and not an `AssetLaunch` sovereign asset. This review covers only the delta over #2791 (merged base).

### Correct core direction

- The unification is right: BUBL's on-chain identity is `generate_custom_token_id("Bubble","BUBL")` and the creator's `TokenCreation` transaction is the authoritative source of the initial allocations. Reads should join `TokenContract` + `SovereignAsset` with DAO winning on symbol collision.
- The `balance_key.rs` short-circuit for `did:zhtp:{key_id}` is a real fix — the wallet-projection path was producing `blake3(dilithium_pk)` while custom-token balances are keyed by the identity's actual `key_id = blake3(dilithium_pk || kyber_pk)` (see `PublicKey::new_with_kyber`). Without this, every wallet-registered citizen was reading zero.
- `sled_store.rs` `NoActiveTransaction` on unbatched `put_token_contract` / `set_token_balance` is a positive change — the old silent no-op was a silent-write footgun. Legacy `process_token_transactions` runs inside `finish_block_processing`'s `begin_block()` window, so the strict path is safe.

### P0 concerns

**1. Rewards keystore now has direct spend authority — operational-security regression**
`rewards/mod.rs::rewards_signer_authorized` returns true iff `keystore.key_id == token.creator.key_id`. The prior model routed rewards through `SovereignAsset.rewards.spend_delegate_key_id` — a distinct key the DAO could rotate independently of the cold creator key. Under this PR, the file at `ZHTP_REWARDS_TREASURY_KEYSTORE` on g1 **is** the BUBL creator's hot key, so a single-node compromise gives the attacker: (a) the ability to mint/transfer BUBL up to `creator.balance_of(creator_key)` = 800M atoms today, and (b) full authority to burn/reclaim once tokens circulate. Recommend one of:
- Keep the delegate model: when the token has a matching `SovereignAsset` with `rewards.spend_delegate_key_id`, authorize the delegate instead of the creator. Fall back to creator only when no delegate is on-chain.
- Introduce a per-token on-chain `spend_delegate` field on `TokenContract` and migrate BUBL to a delegate immediately after this PR merges; document g1 as holding the creator key only for a limited bootstrap window and rotate.

The commit "drop sovereign fallback" is what made this a regression — a stacked-fallback would give operationally-correct behaviour today and a clean upgrade path.

**2. `repair_missing_token_creation_balances` skips repair when the balance tree is partially populated**
`contracts.rs:112–120`:
```rust
if store.count_token_holders(&token_id)? > 0 {
    continue;
}
```
The gate is *any* holder, not *these two* holders. If the tree contains the treasury credit but not the creator credit (or vice-versa, e.g. because a subsequent `TokenTransfer` moved partial funds), repair permanently skips and the missing side stays at zero. Recommend: check each `(token_id, addr)` pair individually via `store.get_token_balance()` and only credit if the current stored value differs from the expected allocation split. This also makes repair robust to partial corruption, not just the "totally empty tree" case.

**3. Balance-repair credits creator using `tx.signature.public_key.key_id` without cross-check**
`contracts.rs:130`: `Address::new(tx.signature.public_key.key_id)`. There is no assertion that the signature was verified — repair reads raw transaction bytes from replayed blocks. On a legitimate chain this is fine, but the same `iter_blocks()` walker will happily replay a `TokenCreation` whose signature was invalid but for some reason survived admission (e.g. a legacy bug we no longer have visibility on). Recommend a `Blockchain::verify_transaction_signature` cross-check inside the repair loop, or at minimum a comment explaining that "chain acceptance implies signature validity" is the load-bearing assumption.

### P1 concerns

**4. `set_token_balance` in `TokenCreation` arm silently overwrites existing balances**
`contracts.rs:815–827` calls `set_token_balance(creator_addr, creator_allocation)` unconditionally after the token is inserted. If the same block committed a `TokenTransfer` earlier for the same creator (transferring some initial allocation out), that write happens through the executor (or through the earlier arm). Since `TokenCreation` for a symbol can only happen once per chain (uniqueness check at line 700), the risk here is only for reruns / replay. Still: the write is unconditional (`set_token_balance`, not `add`), so replaying a `TokenCreation` would reset the ledger. `load_from_store` deliberately skips `process_token_transactions`, so this can't fire on load — but if any future path re-invokes `process_token_transactions` on already-committed blocks (e.g. reprocess after schema migration), balances will be silently reverted. Guard: `if store.get_token_balance(...) == 0 { set_token_balance(...) }`.

**5. `init.rs` blanket backfill loop reads full contract balance maps for every token**
`init.rs:770–795` iterates `iter_token_contracts()` and calls `balances_iter()` on each. As documented in `contracts.rs`, `TokenContract` deserialised from sled has an *empty* balances map, so `entries` will almost always be empty and the loop is a no-op. But two issues:
- On a legacy chain where a contract WAS persisted with embedded balances, all balances are now backfilled at every startup — good.
- The `backfill_token_balances_from_contract` write path shares the `token_balances` tree with the SA-3+ `AssetLaunch` write path introduced in #2791. If a `SovereignAsset` and a legacy `TokenContract` collide on `token_id`, the loop overwrites the sovereign-asset balances at every restart. Recommend: skip iteration for `token_id` values that also have a `SovereignAsset` row.

**6. Case-fragile DID prefix check in `balance_key.rs`**
`if address.starts_with("did:zhtp:") || address.starts_with("did:ZHTP:")` — misses `did:Zhtp:` (mixed case) and any other capitalisation. `did:foo:{64-hex}` also silently bypasses to the "return raw bytes as key_id" branch further down. Recommend: normalize with `address.trim().to_ascii_lowercase().starts_with("did:zhtp:")` and validate the full method prefix rather than substring-matching.

**7. `handle_balance` calls `validate_did` twice**
`rewards/mod.rs:867` and again inside the treasury match arm at ~876. Second call is unreachable-safe (the first returned) but confusing. Just re-use the previously validated `key_id`.

**8. `token_contract_symbol_exists` is O(N)** per asset in `iter_sovereign_assets`
`token/mod.rs:119–125` × the `include_sovereign_asset_in_token_api` filter turns the `/api/v1/token/list` and `/api/v1/token/{addr}/balances` paths into O(assets × contracts). At current scale <10 each, fine. Precompute a `HashSet<upper_symbol>` from `iter_token_contract_entries()` once and pass it in.

**9. Env var name drift: `ZHTP_REWARDS_ASSET_ID` still readable but no deprecation timeline**
`rewards_token_id_override` accepts both `ZHTP_REWARDS_TOKEN_ID` and `ZHTP_REWARDS_ASSET_ID`. Good. Recommend adding a `warn!` when the deprecated form is used so operators migrate before the alias is dropped.

### Test coverage gaps

- `repair_missing_token_creation_balances_restores_zeroed_tree` covers the happy path but not:
  - partial-holders case (concern 2)
  - repair idempotency across multiple invocations
  - a token whose `TokenCreation` tx lives in a pruned block (silently skipped)
- No test asserting `balance_key.rs` returns different key_ids for `did:zhtp:{k}` vs raw hex `{k}` when the wallet has a matching wallet_id. Add: same DID, same hex, different resolution outcomes.
- No integration test showing rewards handler correctly authorises the creator keystore AND refuses a non-creator keystore — currently only asserted at load time via `warn!`.

### Nits

- `rewards/mod.rs:1200` docstring "the BUBL creator keystore (the signer of the founding TokenCreation tx)" — worth noting explicitly that this key can move all of the creator's on-chain allocation, not just rewards budgets.
- `contracts.rs:117` `let Some(contract) = store.get_token_contract(&token_id)? else { continue; };` — good use of let-else.
- `token/mod.rs:658–664` `seen: HashSet<[u8; 32]>` is populated by decoding each `TokenListItem.token_id` back from hex. Cheaper to keep the raw bytes from earlier in the list-build.
- `handle_balance` renames `rewards_asset_id` internally to `rewards_asset_id` in `TreasuryConfig` but the log message says "BUBL signer loaded — token_id=…" — consistent naming would help greppability. Pick one.
- `treasury.rewards_asset_id` in the JSON response is emitted as `asset_id` key. Given the whole thrust of the PR is "not an asset, a token", either rename the JSON field to `token_id` or explain why the wire format keeps the old name (mobile-app compat?).

### Verdict

Direction is right and closes a real UX bug (BUBL invisible at rewards endpoints). Concerns 1 (op-sec regression) and 2 (partial-holders repair-skip) are the two I would want addressed before merging. 3–5 are worth a follow-up commit. The rest are nits.

Approve on the read-and-display side; hold on the rewards-authorisation change until the delegate fallback is restored or a rotation plan is documented.

PR #2796, opened 2026-07-08
Windows build error fix.

PeterGRutherford:
## Stack

```
development
  └── #2796 Windows build (this PR)  ← merge first
        └── #2797 Windows linker stack reserve
```

Independent of #2920 (identity registration).

## Summary

Fixed Unix-specific assumptions that blocked the Windows build:

1. **IPC** — Windows TCP loopback + port file instead of Unix domain sockets
2. **Network imports** — `BLE_MESH_SERVICE_UUID` / `Mutex` for bluetooth + wifi_direct
3. **Shutdown** — portable path instead of Unix-only `libc::kill`/`SIGTERM`

## Linked issues

- #578 / #577 Windows Bluetooth build errors
- #2915 Windows IPC transport
- #2918 BLE/wifi_direct imports
- Stack child: #2797 / #2916

## Test plan

- [ ] cargo check on Windows for lib-blockchain, lib-network, zhtp
- [ ] Unix path unchanged

## Risk and reversibility

Minimal. `#[cfg]` guards preserve Unix behaviour.

PR #2797, opened 2026-07-09
fix(windows): raise linker stack reserve to 16 MB for overflow error.

### LINKED ISSUE #2510: zhtp.exe stack-overflows on startup on Windows (1 MB default main-thread stack too small)
## Summary
`zhtp/src/main.rs` uses `#[tokio::main]` with no custom stack configuration. The deeply-nested startup futures (`start_validator` / `start_full_node` -> `start_node` which is a 4000+ line async fn) produce a Future state that exceeds Windows' default 1 MB main-thread stack. The process crashes with `thread 'main' has overflowed its stack` shortly after `Starting network components for peer discovery...`, before any networking is up.

## Impact
- **`zhtp.exe` cannot be run at all on stock Windows.** Every developer or operator on Windows hits this on the first run.
- No useful error message - the process aborts via Windows guard-page violation. Users see nothing but a stderr line. Misleads investigation toward networking, config, etc.
- Reproducible on Windows 11 Pro with the standard MSVC toolchain.

## Reproduction
```powershell
cargo build --release -p zhtp
target\release\zhtp.exe --config zhtp\configs\test-node1.toml
# -> silently dies after "Starting network components for peer discovery..."
# -> stderr: thread 'main' has overflowed its stack
```

## Root cause
- `tokio::main` runs the async main on the OS main thread.
- Windows MSVC default reserves 1 MB stack at link time; tokio's `thread_stack_size` only affects worker threads, not main.
- `RuntimeOrchestrator::start_node()` is a single massive async fn whose Future state machine after LTO optimization still exceeds 1 MB on Windows.

## Suggested fixes (any one suffices)

**Option A - change main.rs to spawn on a thread with larger stack** (already applied locally on PR #2499 branch):
```rust
fn main() -> anyhow::Result<()> {
    let handle = std::thread::Builder::new()
        .name("zhtp-main".to_string())
        .stack_size(8 * 1024 * 1024)
        .spawn(|| {
            let runtime = tokio::runtime::Builder::new_multi_thread()
                .enable_all()
                .thread_stack_size(8 * 1024 * 1024)
                .build()?;
            runtime.block_on(async_main())
        })?;
    handle.join().expect("zhtp-main thread panicked")
}

async fn async_main() -> anyhow::Result<()> { /* original tokio::main body */ }
```

**Option B - set the linker stack reserve** via `.cargo/config.toml`:
```toml
[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "link-arg=/STACK:16777216"]
[target.x86_64-pc-windows-gnu]
rustflags = ["-C", "link-arg=-Wl,--stack,16777216"]
```

**Option C - break up `start_node()`** into smaller phased async functions called sequentially so the Future state never exceeds 1 MB. Larger refactor but addresses the root cause.

PeterGRutherford:
## Stack

```
development
  └── #2796 Windows build (IPC / BLE / Mutex)  ← base
        └── #2797 Windows linker stack reserve (this PR)
```

Independent of #2920 (identity registration).

## Summary

Raise the Windows linker stack reserve from 1 MB to 16 MB to prevent the thread `main` has overflowed its stack crash on `zhtp.exe`. The `#[tokio::main]` entry point runs on the OS main thread (default 1 MB stack on Windows MSVC). Deeply nested startup futures in `RuntimeOrchestrator::start_node()` exceed 1 MB after LTO.

## Linked issues

- #2916 Windows: raise linker stack reserve
- Related stack base: #2796 / #2915 / #2918

## Test plan

- [ ] Manual verification on Windows 11: cargo build + run without stack overflow
- [ ] dumpbin stack reserve → 16 MB
- [ ] Linux/macOS builds unaffected

## Risk and reversibility

Low. Windows-only linker flags in `.cargo/config.toml`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,3 +4,15 @@
 # for all cargo-invoked processes. Harmless in production (larger stacks only).
 [env]
 RUST_MIN_STACK = "16777216"
+
+# Windows: the main thread created by #[tokio::main] inherits the OS-default
+# 1 MB stack, which is too small for the deeply-nested startup futures in
+# RuntimeOrchestrator::start_node().  Raising the linker stack reserve to 16 MB
+# prevents "thread 'main' has overflowed its stack" crashes on stock Windows.
+# This is a linker-level setting and does NOT affect runtime memory usage
+# (only the reserved address space, not committed pages).
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "link-arg=/STACK:16777216"]
+
+[target.x86_64-pc-windows-gnu]
+rustflags = ["-C", "link-arg=-Wl,--stack,16777216"]


umwelt:
**Stack note:** Rebased onto `fix/windows-build` (#2796). Base is now that branch — merge #2796 first, then this PR.

```
development
  └── #2796
        └── #2797 (this PR)
```

umwelt:
Superseded: the 16 MB Windows linker stack reserve landed via #2941 (merged to development). This PR also targets base `fix/windows-build`, not `development`, so it can't merge here as-is. Closing — the change is already on `development`.

Issue #2814, opened 2026-07-09
DAO N4: Retire node-local rewards.sled


umwelt:
## Description

Remove duplicate eligibility store. Single source: consensus `bubl_reward_*` trees (generalized per asset). Migration or documented reset for existing node state.

## Acceptance criteria

- [ ] Q6 decision recorded
- [ ] No writes to `rewards.sled` on claim path
- [ ] Migration script or reset runbook for testnet
- [ ] Tests prove no double-claim across restart

## Dependencies

| Type | Items |
|------|-------|
| **Epic** | #2799 |
| **Blocked by** | Q6, N2 |
| **Relates to** | — |

## Metadata

| Field | Value |
|-------|-------|
| **Story ID** | N4 |
| **Workstream** | Node/API |
| **Phase** | Gated |

---
*DAO Launch v1 · child of #2799*

PR #2844, opened 2026-07-10
DAO N2: Hybrid rewards settlement (Path C) (#2812)

### LINKED ISSUE #2812: DAO N2: Hybrid rewards settlement (Path C)
## Description

Node validates claim event against on-chain policy + on-chain eligibility, builds `RewardClaim` signed by spend delegate, submits to mempool. Supports batching. App polls balance after commit.

## Acceptance criteria

- [ ] POST `/api/v1/rewards/claim` ends in mempool `RewardClaim`, not bare `TokenTransfer`
- [ ] Eligibility read from consensus trees, not node-local sled
- [ ] Attestation model per Q4 implemented
- [ ] Balance reflects committed blocks

## Dependencies

| Type | Items |
|------|-------|
| **Epic** | #2799 |
| **Blocked by** | P3, P4, Q4 |
| **Relates to** | #2784 |

## Metadata

| Field | Value |
|-------|-------|
| **Story ID** | N2 |
| **Workstream** | Node/API |
| **Phase** | 2 |

---
*DAO Launch v1 · child of #2799*

umwelt:
## Summary

Implements N2 hybrid rewards settlement: claim endpoints now submit signed `RewardClaim` transactions to mempool instead of bare `TokenTransfer`s.

## Changes

- **`settlement.rs`**: spend-delegate attestation — node builds/signs `RewardClaim` with delegate keystore
- **`rewards/mod.rs`**: claim + status handlers read eligibility from consensus trees (via blockchain facades), not node-local sled reservation trees
- **`blockchain/rewards.rs`**: token-id-parameterized eligibility/policy facades (`reward_welcome_claimed`, `reward_expected_amount`, etc.)

## Acceptance criteria

- [x] POST `/api/v1/rewards/claim` enqueues `RewardClaim` (not `TokenTransfer`)
- [x] Eligibility read from consensus trees (sled store behind blockchain facade)
- [x] Spend-delegate attestation model (Q4/P5 delegate signs claims)
- [x] Balance endpoint already reflects on-chain token balance after commit

## Notes

- Node-local `rewards.sled` history/lifetime trees retained until N4 (#2814)
- Closes #2812

### DIFF
--- a/lib-blockchain/src/blockchain.rs
+++ b/lib-blockchain/src/blockchain.rs
@@ -2266,51 +2266,89 @@ impl Blockchain {
     /// Check whether a transaction's nonce is still valid against committed chain state.
     ///
     /// Returns `true` for transaction types that don't carry nonces (they are
-    /// validated by other means).  For `TokenTransfer`, the nonce must equal
-    /// the sender's current on-chain nonce for the target token — anything
-    /// lower has already been applied and is a replay.
+    /// validated by other means). For `TokenTransfer` and `RewardClaim`, the
+    /// nonce must equal the sender's current on-chain nonce for the target
+    /// token — anything lower is a replay, anything higher is a gap.
     pub fn is_nonce_current(&self, tx: &Transaction) -> bool {
-        if tx.transaction_type == TransactionType::TokenTransfer {
-            if let Some(transfer) = tx.token_transfer_data() {
+        let (token_id, from, nonce) = match tx.transaction_type {
+            TransactionType::TokenTransfer => {
+                let Some(transfer) = tx.token_transfer_data() else {
+                    return true;
+                };
                 let token_id = if transfer.token_id == [0u8; 32] {
                     crate::contracts::utils::generate_lib_token_id()
                 } else {
                     transfer.token_id
                 };
-                let expected = match self.token_nonce(&token_id, &transfer.from) {
-                    Ok(n) => n,
-                    Err(e) => {
-                        tracing::warn!(
-                            error = %e,
-                            tx = %tx.hash(),
-                            "is_nonce_current: nonce read failed; failing closed (rejecting tx)"
-                        );
-                        return false;
-                    }
+                (token_id, transfer.from, transfer.nonce)
+            }
+            TransactionType::RewardClaim => {
+                let Some(claim) = tx.reward_claim_data() else {
+                    return true;
                 };
-                if transfer.nonce < expected {
-                    tracing::debug!(
-                        "Stale nonce: tx {} has nonce {} but chain expects {} for sender {}",
-                        tx.hash(),
-                        transfer.nonce,
-                        expected,
-                        hex::encode(&transfer.from[..8]),
-                    );
-                    return false;
-                }
-                if transfer.nonce > expected {
-                    tracing::debug!(
-                        "Future nonce: tx {} has nonce {} but chain expects {} for sender {}",
-                        tx.hash(),
-                        transfer.nonce,
-                        expected,
-                        hex::encode(&transfer.from[..8]),
-                    );
-                    return false;
+                (claim.token_id, claim.from, claim.nonce)
+            }
+            _ => return true,
+        };
+
+        let expected = match self.token_nonce(&token_id, &from) {
+            Ok(n) => n,
+            Err(e) => {
+                tracing::warn!(
+                    error = %e,
+                    tx = %tx.hash(),
+                    "is_nonce_current: nonce read failed; failing closed (rejecting tx)"
+                );
+                return false;
+            }
+        };
+        if nonce != expected {
+            tracing::debug!(
+                "Nonce mismatch: tx {} has nonce {} but chain expects {} for sender {}",
+                tx.hash(),
+                nonce,
+                expected,
+                hex::encode(&from[..8]),
+            );
+            return false;
+        }
+        true
+    }
+
+    /// Reject duplicate or conflicting `RewardClaim` txs already in the mempool.
+    fn pending_reward_claim_conflicts(&self, incoming: &crate::transaction::reward_claim::RewardClaimData) -> bool {
+        use crate::transaction::reward_claim::RewardEventKind;
+        for tx in &self.pending_transactions {
+            if tx.transaction_type != TransactionType::RewardClaim {
+                continue;
+            }
+            let Some(data) = tx.reward_claim_data() else {
+                continue;
+            };
+            if data.token_id == incoming.token_id
+                && data.from == incoming.from
+                && data.nonce == incoming.nonce
+            {
+                return true;
+            }
+            if data.token_id != incoming.token_id || data.owner_did != incoming.owner_did {
+                continue;
+            }
+            if data.event != incoming.event {
+                continue;
+            }
+            match data.event {
+                RewardEventKind::Welcome
+                | RewardEventKind::DailyCheckin
+                | RewardEventKind::ActiveSession => return true,
+                RewardEventKind::NewPartner => {
+                    if data.peer_did == incoming.peer_did {
+                        return true;
+                    }
                 }
             }
         }
-        true
+        false
     }
 
     /// Get pending transactions
@@ -2412,6 +2450,18 @@ impl Blockchain {
                 "Transaction {} rejected: stale or future nonce",
                 &tx_hash[..16]
             ));
+        }
+
+        if transaction.transaction_type == TransactionType::RewardClaim {
+            if let Some(data) = transaction.reward_claim_data() {
+                if self.pending_reward_claim_conflicts(data) {
+                    let tx_hash = hex::encode(transaction.hash().as_bytes());
+                    return Err(anyhow::anyhow!(
+                        "Transaction {} rejected: conflicting RewardClaim already pending",
+                        &tx_hash[..16]
+                    ));
+                }
+            }
         }
 
         // Atomic with the push: update admission state so subsequent admit()
--- a/lib-blockchain/src/blockchain/rewards.rs
+++ b/lib-blockchain/src/blockchain/rewards.rs
@@ -1,105 +1,165 @@
-//! BUBL reward eligibility facades (sled-first reads).
+//! Reward eligibility facades — sled/consensus tree reads (authoritative).
 
-use crate::contracts::utils::generate_custom_token_id;
+use crate::rewards_policy::{
+    expected_amount_for_trigger, legacy_bubl_policy, policy_hash, validate_rewards_policy,
+    weekly_partner_cap, RewardsPolicyV1,
+};
 use crate::transaction::reward_claim::{
-    daily_claim_key, partner_claim_key, partner_count_key, streak_key, welcome_claim_key,
-    RewardEventKind, RewardStreakRecord,
+    daily_claim_key, partner_claim_key, partner_count_key, reward_event_to_policy_event,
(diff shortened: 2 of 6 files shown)


umwelt:
## Code review — PR #2844 (DAO N2: hybrid rewards settlement)

**Scope reviewed:** the single commit `c85b660e` — `settlement.rs` (new), `rewards/mod.rs` (mempool routing), `blockchain/rewards.rs` (token-parameterised facades). The N2 shift is: eligibility now reads consensus trees; claims submit as signed `RewardClaim` txes into the mempool instead of `TokenTransfer` mints.

Two P0s below block on chain-level DoS. The rest are correctness / UX regressions from removing the sled CAS gate.

---

### P0

**1. `RewardClaim` bypasses the mempool nonce gate → unbounded same-DID duplicate submission → mempool + gossip + block-space DoS**

`Blockchain::is_nonce_current` at `lib-blockchain/src/blockchain.rs:2272` gates only `TokenTransfer`:

```rust
if tx.transaction_type == TransactionType::TokenTransfer {
    // ... reject on stale/future nonce ...
}
true
```

`RewardClaim` falls straight through and returns `true` unconditionally. Nothing else in `verify_and_enqueue_transaction` dedups by (from, nonce) either. Combined with the removal of the sled CAS reservation in this PR:

- Attacker hits `POST /api/v1/rewards/claim` N times for their DID in rapid succession.
- Each request reads the same on-chain welcome state (`bc.reward_welcome_claimed` returns false until block commit), reads the same treasury nonce, and produces a distinct signed `RewardClaim` tx with `nonce=N`.
- All N txes pass `verify_transaction` (validation.rs:795 only checks `amount!=0`, `from==sig.pk.key_id`, and `peer_did` for NewPartner).
- All N enter `self.pending_transactions` (`blockchain.rs:2439`) and get broadcast to mesh peers (`blockchain.rs:2347`).
- On commit, `apply_reward_claim` accepts tx #1, rejects #2..#N as `Ok(None)` (executor `reward_claim.rs:228-234`), which the executor treats as **successful no-op txes** (per PR #2832 review). Block gets bloated with N zero-effect txes; every validator has to verify them; sled records them forever.

Endpoint is unauthenticated (`no body signing required` — matches the prior TokenTransfer behaviour) so *any* caller can drive this against *any* DID.

Under the old flow the sled CAS at `do_welcome` short-circuited before `mint_to_user` even ran, so N submits produced at most one mempool tx. This PR removes that with no replacement.

Minimum fix: extend `is_nonce_current` to cover `RewardClaim` (same `token_nonce(token_id, from)` check as TokenTransfer), or dedup by `(from, token_id, event, owner_did, date/week)` at mempool admit. Ideally both — the nonce gate handles the same-signer replay, the eligibility gate handles the same-event replay.

**2. `apply_reward_claim`'s ineligibility → `Ok(None)` → committed as successful no-op tx**

Flagged in the PR #2832 review as P1; this PR makes it a P0. The prior flow's failures short-circuited before submitting a tx. Now:

- Every ineligible RewardClaim (welcome already claimed, wrong amount, cap reached, streak state raced, DID not registered, nonce collision from finding #1, etc.) commits as `TxOutcome::LegacySystem` — a "successful" no-op.
- The endpoint returns `awarded: true` with a `tx_hash` that will land in a block and do nothing.
- Node-local history/lifetime trees record the "award"; balance never moves.

Any `Ok(None)` branch inside `apply_reward_claim` should either return `Err(TxApplyError::InvalidType(...))` so the tx is rejected outright, or the API layer should synchronously resolve the outcome and refuse to return `awarded: true` until the tx applies. Right now `record_event` and the response lie whenever the executor drops the tx.

---

### P1

**3. Stale sled trees still opened at startup, never read/written**

`RewardsHandler` fields at `zhtp/src/api/handlers/rewards/mod.rs:157-162` — `welcomed`, `checkins`, `sessions`, `streak`, `partners`, `partners_count` — are opened at construction but no code path touches them post-PR. `grep 'self\.(welcomed|checkins|sessions|streak|partners|partners_count)'` returns zero hits. The module doc-block still describes them (lines 21-29). PR body says "Node-local `rewards.sled` history/lifetime trees retained until N4 (#2814)" — but these six are eligibility trees, not history/lifetime. Delete them from the struct + open call + doc block, or add a code comment explaining why they're intentionally held (I couldn't find a reason). Leaving them wired-up creates a footgun where a future refactor accidentally reintroduces reads that silently drift from consensus state.

**4. Streak amount computed at API layer with a fallback that will diverge from the executor**

`do_daily_checkin` (`rewards/mod.rs` around L520-540) reads `bc.reward_streak`, computes `next_streak` locally, then `bc.reward_checkin_amount_for_streak(...).unwrap_or_else(|| Self::checkin_amount_for_day(next_streak))`. The executor recomputes the same thing at commit time. If:
- another same-DID claim raced in between the API read and this tx applying, or
- the policy document changes between read and apply,

then `data.amount != expected` and `apply_reward_claim` returns `Ok(None)` (silent no-op tx per finding #2). Same for the fallback `unwrap_or` in welcome/session/new_partner: `Self::checkin_amount_for_day` and `REWARD_WELCOME`/`REWARD_ACTIVE_SESSION`/`REWARD_NEW_PARTNER` constants only match `legacy_bubl_policy` — for any non-BUBL token whose policy document is missing/invalid, executor returns `Err` (not `Ok(None)`), tx fails hard, but API already told the client "awarded: true". Recommend: fetch the policy-derived amount inside a single blockchain read lock, and if `reward_expected_amount` returns `None`, return `503 unavailable` instead of falling back.

**5. Same-block streak recompute race**

Two same-DID daily-checkin claims raced through the endpoint before either commits will each compute `next_streak` from the last committed streak record, then submit two claims with the same amount. Executor accepts the first, rejects the second (welcome/daily key already present) as `Ok(None)`. No streak corruption, but see finding #2 — client thinks both succeeded.

**6. `record_event` writes local history before tx applies**

`rewards/mod.rs` calls `self.record_event(did, event, |s| { ... })` immediately after `settlement::submit_reward_claim` returns Ok. That records the "award" in node-local history and bumps the `lifetime` counter with no knowledge of whether the tx actually applied. Under finding #1 or #2 the history and `lifetime_amount` in the balance response will overreport by a wide margin. Move `record_event` to a post-commit hook (or gate it on observing the tx in an applied block), or explicitly stop pretending the local view matches consensus.

**7. Two separate blockchain read locks per handler instead of one**

Each handler acquires `blockchain.read()` twice — once for eligibility, once for amount lookup — then `blockchain.write()` for submit. Under lock churn on a hot endpoint this is a small throughput hit; more importantly, the two reads see two different snapshots of the state so eligibility and amount can be internally inconsistent. Fold into one `blockchain.read()` block.

---

### P2 / nits

**8.** `settlement.rs` has no trailing newline (visible in the diff header `\ No newline at end of file`).

**9.** `chain_id_from_env` promoted from private to `pub(crate)` for reuse by `settlement.rs`. There's a nearly identical `runtime::token_utils::chain_id_from_env` referenced in the doc-comment — consider consolidating to one crate-level helper rather than exporting a rewards-module private.

**10.** Module docstring (rewards/mod.rs L21-36) still describes the sled tree layout as the eligibility source of truth. Rewrite to reflect that eligibility now lives on-chain and these local trees are (per finding #3) unused.

**11.** `settlement.rs` currently only handles the "generate a fresh signed claim" path. If finding #1 is fixed with mempool-side dedup, a client-side retry (same `owner_did` + `event` + `date/week` → same tx_hash?) would help idempotency. Right now every submit gets a fresh `Signature::default()` slot + fresh sign, so retries produce distinct tx_hashes.

**12.** No test coverage in the PR for the new mempool path — same-DID double-submit behaviour, executor no-op semantics under duplicate nonce, and policy-fetch fallback are all unexercised. Given finding #1 is a chain-DoS, at minimum an integration test that submits two same-DID welcome claims and asserts exactly one is in the applied block (and neither is in the "successful no-op" bucket) should ship with this PR.

---

### Verdict

Do not merge until (1) and (2) are addressed. The rest are correctness/UX regressions from removing the CAS gate without a mempool-side replacement — nothing wrong with the design goal (Path C), but the current implementation exposes chain and gossip surface to unauthenticated same-DID replay. The "hybrid" in the title is only load-bearing if the mempool actually stops the duplicates; today it doesn't.

umwelt:
Addressed the P0 blockers from @umwelt's review:

**P0-1 — mempool nonce / duplicate flood**
- `is_nonce_current` now covers `RewardClaim` (exact `(token_id, from, nonce)` match, same as `TokenTransfer`)
- `verify_and_enqueue_transaction` rejects conflicting `RewardClaim` already pending (same nonce slot or same eligibility slot: welcome/daily/session per DID, partner per peer)

**P0-2 — silent no-op commits**
- All ineligibility paths in `apply_reward_claim` now return `Err` (including nonce replay via `ReplayDropped`); executor no longer maps `Ok(None)` → successful `LegacySystem` no-op

**P1 items also landed**
- Removed dead eligibility sled trees (`welcomed`, `checkins`, …) from handler struct
- `plan_reward_claim` does single chain snapshot for eligibility + policy amount (no hardcoded fallback; 503 if policy missing)
- Removed optimistic `record_event` before commit; responses include `submitted: true`
- Updated module docs

Commit: $(git rev-parse --short HEAD)

umwelt:
## Response to review (commit `66536e31`, merged)

Thanks — both P0s were addressed in the squash before merge (`fix(rewards): address P0 review on RewardClaim mempool DoS`). Testnet has since exercised the path end-to-end (identity register → welcome `RewardClaim` commit at H=34959, no halt).

### P0 — addressed

**1. Mempool nonce / duplicate gate**

- `is_nonce_current` now covers `RewardClaim` (exact `token_nonce(token_id, from)` match) — `lib-blockchain/src/blockchain.rs:2285-2315`
- `pending_reward_claim_conflicts` rejects duplicate/conflicting claims already in the pending pool (same `(from, nonce)`, or same `(owner_did, event)` for welcome/daily/session, or same peer for new_partner) — `blockchain.rs:2318-2351`, enforced at `2447-2464`
- Observed on testnet: second in-flight welcome claim rejected with `conflicting RewardClaim already pending`

**2. Ineligible claims → hard reject (no silent no-op blocks)**

- `apply_reward_claim` uses `reject_claim` → `Err(TxApplyError::InvalidType(...))` for all ineligibility paths — `reward_claim.rs:58-60`
- Commit failure on unregistered DID (our Jul 11 incident) is intentional safety; API now requires on-chain identity registration before claim

### P1 — addressed in merge

| # | Item | Status |
|---|------|--------|
| 3 | Unused eligibility sled trees | Removed; only `history` + `lifetime` remain (`mod.rs:20-21`, `130-137`) |
| 4 | Policy amount fallback divergence | `plan_reward_claim` uses `reward_expected_amount` only; returns 503 if policy missing (`settlement.rs:166-168`) |
| 6 | Optimistic `record_event` before commit | Removed from claim path (`execute_claim` → `run_claim_flow` only); `record_event` now dead code pending N4 |
| 7 | Double blockchain read | Single `blockchain.read()` inside `run_claim_flow` for plan + nonce |

### Still open (tracked)

- **#12** Integration test: two same-DID welcome submits → exactly one mempool entry, one applied tx (not yet in repo)
- **#9** Consolidate `chain_id_from_env` helpers
- **#11** Client-side idempotency / deterministic tx_hash on retry
- **N4 (#2814)** Remove dead `record_event` + history/lifetime sled trees entirely

Happy to follow up with the integration test in a small fix PR if you want it before N4.

Issue #2799, opened 2026-07-09
DAO Epic: DAO Launch v1 (Sovereign Asset)


umwelt:
# Epic: DAO Launch v1 (Sovereign Asset)

**Working title:** "Create a DAO like BUBL", productized.

## Problem

BUBL proved the pattern but is not a product: token launch, rewards, registry, governance, and discovery are fragmented across manual operator steps, hardcoded BUBL constants, env-var ticker hacks, and a mobile Create DAO screen that is mock-only and promises features the protocol rejects.

## Outcome

Any user can launch a DAO (token + registry entry + manifest + optional rewards + optional governance + `.sov` domain) from `zhtp-cli` or the SovSwap app, without operators hand-editing validator env. BUBL becomes the first migrated Sovereign Asset instance.

## Guiding principle (locked 2026-07-09)

The SovSwap UI is the product target: it came first, and the protocol is built to solidly support what it promises. Where soundness requires a change, the protocol always wins and the UI adapts (conflicts C6, C7). Features the protocol lacks today are committed scope pending design decisions, not optional extras.

## Success criteria

- [ ] One CLI command and one app screen perform a real on-chain launch end-to-end on testnet
- [ ] Rewards are consensus events validated against DHT-hosted policy (`policy_hash` on-chain); zero BUBL-specific constants in reward path
- [ ] Validator rewards activation derived from chain state (module bit + funded delegate), not env ticker hack
- [ ] Hot spend delegate separated from creator cold key, governance-gated rotation
- [ ] Wallets discover assets via `GET /api/v1/assets` + manifest verification; no ticker hardcoding
- [ ] Q1–Q10 locked in decision log; design-gated stories started or design locked

## Relationship to Sovereign Asset

- Parent primitive epic: #2780
- SA-2 #2782 · SA-3 #2783 · SA-4 #2784 · SA-6 #2786 · SA-7 #2787 · SA-8 #2788

## Codebase verification (2026-07-09)

| Claim | Status |
|-------|--------|
| 80/20 at launch only (`2000` bps) | ✅ `token_creation.rs`, `asset_tx.rs` |
| No 80/20 on subsequent `TokenMint` | ✅ `executor.rs` |
| Hardcoded BUBL reward amounts | ✅ `reward_claim.rs`, `rewards/mod.rs` |
| Node-local `rewards.sled` | ✅ `rewards/mod.rs` |
[...]

PR #2845, opened 2026-07-11
feat(rewards): D3 legacy BUBL rewards module upgrade

umwelt:
## Summary

Enables on-chain `RewardsModuleState` for legacy testnet BUBL (pre-`AssetLaunch` `TokenCreation`) via `AssetModuleUpgrade`, plus an operator seed tool.

- **Executor:** materialize legacy `TokenContract` projection when upgrading an asset not yet in `assets/` sled
- **Executor:** treat rewards as enabled only when `RewardsModuleState` exists (projection flag alone is not enough)
- **Tool:** `seed_bubl_rewards_upgrade` — build signed `AssetModuleUpgrade` + policy bundle for broadcast

## Testnet ops (already executed)

| Step | Result |
|------|--------|
| Deploy binary `092c6926…` | g1/g2/g3 |
| Broadcast upgrade `836ea97a…` | committed ~34693 |
| `rewards_activation.toml` on validators | endpoints active |
| Welcome claim smoke | `submitted: true`, `RewardClaim` in mempool |

Spend delegate: `bubl-creator` (interim; dedicated hot wallet rotation later).

## Closes

Part of DAO epic #2799 / D3 BUBL migration. Unblocks N2 claim path on live testnet after N3 activation config.

## Checklist

- [x] `seed_bubl_rewards_upgrade` builds signed tx
- [x] Legacy BUBL upgrade applies on testnet
- [x] Rewards claim endpoint returns 200 + `submitted: true`

### DIFF
--- a/lib-blockchain/src/execution/sovereign_asset.rs
+++ b/lib-blockchain/src/execution/sovereign_asset.rs
@@ -1,10 +1,11 @@
 //! Sovereign Asset executor apply helpers (SA-3..SA-7).
 
 use crate::contracts::sovereign_asset::{
-    validate_governance_verifier, AssetAuthority, AssetIdSource, AssetModuleFlags,
-    CurveModuleHeader, GovernanceModuleHeader, GovernanceModuleState, GovernanceVerifierKind,
-    GovernanceVerifierState, PendingAuthorityTransfer, RewardsModuleHeader, RewardsModuleState,
-    SovereignAsset, SupplyMode, AUTHORITY_TRANSFER_TIMELOCK_BLOCKS,
+    project_from_token_contract, validate_governance_verifier, AssetAuthority, AssetIdSource,
+    AssetModuleFlags, CurveModuleHeader, GovernanceModuleHeader, GovernanceModuleState,
+    GovernanceVerifierKind, GovernanceVerifierState, PendingAuthorityTransfer,
+    RewardsModuleHeader, RewardsModuleState, SovereignAsset, SupplyMode,
+    AUTHORITY_TRANSFER_TIMELOCK_BLOCKS,
 };
 use crate::execution::tx_apply::{apply_token_mint, StateMutator};
 use crate::execution::{TxApplyError, TxApplyResult};
@@ -144,15 +145,30 @@ pub fn apply_asset_launch(
     })
 }
 
+/// Resolve a sovereign asset for upgrade: sled record first, else legacy token projection.
+fn resolve_sovereign_asset_for_upgrade(
+    mutator: &StateMutator<'_>,
+    asset_id: &[u8; 32],
+) -> TxApplyResult<SovereignAsset> {
+    if let Some(asset) = mutator.get_sovereign_asset(asset_id)? {
+        return Ok(asset);
+    }
+    let token_id = TokenId::new(*asset_id);
+    let contract = mutator
+        .get_token_contract(&token_id)?
+        .ok_or_else(|| TxApplyError::InvalidType("asset not found".to_string()))?;
+    project_from_token_contract(&contract, None).ok_or_else(|| {
+        TxApplyError::InvalidType("legacy asset cannot be upgraded".to_string())
+    })
+}
+
 pub fn apply_asset_module_upgrade(
     mutator: &StateMutator<'_>,
     signer_key_id: [u8; 32],
     payload: &AssetModuleUpgradePayloadV1,
     block_height: u64,
 ) -> TxApplyResult<()> {
-    let mut asset = mutator
-        .get_sovereign_asset(&payload.asset_id)?
-        .ok_or_else(|| TxApplyError::InvalidType("asset not found".to_string()))?;
+    let mut asset = resolve_sovereign_asset_for_upgrade(mutator, &payload.asset_id)?;
 
     ensure_creator_or_governance(mutator, &asset, signer_key_id)?;
 
@@ -263,7 +279,7 @@ fn enable_rewards(
     asset: &mut SovereignAsset,
     cfg: &RewardsLaunchConfig,
 ) -> TxApplyResult<()> {
-    if asset.module_flags.has_rewards() {
+    if mutator.get_rewards_module_state(&asset.asset_id)?.is_some() {
         return Err(TxApplyError::InvalidType("rewards module already enabled".into()));
     }
     asset.module_flags.0 |= AssetModuleFlags::REWARDS;
--- a/lib-blockchain/src/transaction/core.rs
+++ b/lib-blockchain/src/transaction/core.rs
@@ -953,6 +953,25 @@ impl Transaction {
         }
     }
 
+    /// Create a new Sovereign Asset module upgrade transaction.
+    pub fn new_asset_module_upgrade_with_chain_id(
+        chain_id: u8,
+        signature: Signature,
+        memo: Vec<u8>,
+    ) -> Self {
+        Transaction {
+            version: TX_VERSION_V8,
+            chain_id,
+            transaction_type: TransactionType::AssetModuleUpgrade,
+            inputs: Vec::new(),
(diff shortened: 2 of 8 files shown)


umwelt:
## Code review — PR #2845 (D3: legacy BUBL rewards module upgrade)

**Scope reviewed:** the single commit `b6ba756e` — legacy-token projection fallback in `apply_asset_module_upgrade`, the sled-state (not projection-flag) idempotency check in `enable_rewards`, a new `new_asset_module_upgrade_with_chain_id` constructor, and the `seed_bubl_rewards_upgrade` operator tool.

The Rewards path is the stated goal and looks correct end-to-end. Concerns below are about (a) the wider surface the fallback resolver opens up for non-Rewards modules, and (b) missing on-chain validation of operator-supplied policy blobs. No P0.

---

### P1

**1. `enable_rewards` writes `cfg.policy_document` to sled without cross-checking `policy_hash` / `policy_cid`**

`lib-blockchain/src/execution/sovereign_asset.rs:277-302`:

```rust
mutator.put_rewards_module_state(&asset.asset_id, &RewardsModuleState {
    policy_cid: cfg.policy_cid,
    policy_hash: cfg.policy_hash,
    ...
})?;
if let Some(doc) = &cfg.policy_document {
    mutator.put_rewards_policy_document(&cfg.policy_hash, doc)?;
}
```

Anyone with the creator key can submit a payload where `policy_document` doesn't hash to `policy_hash`, or where `policy_cid` doesn't derive from `policy_document`. The tool constructs both correctly, but on-chain code cannot trust the sender. Downstream fallout:

- `apply_reward_claim` → `resolve_rewards_policy` re-validates the stored document and hard-errors on hash mismatch (`reward_claim.rs:36-39`). Every subsequent claim on this asset returns `Err(TxApplyError::InvalidType("stored rewards policy hash mismatch"))` — the rewards module is effectively bricked for that asset until an authority rotate/relaunch.
- `policy_cid` is stored but never re-derived on-chain, so a bogus CID just sits there and leaks into any surface that echoes it.

Fix: before `put_rewards_policy_document`, run `validate_rewards_policy(doc)` → `policy_hash(...)` and require equality with `cfg.policy_hash`; similarly recompute the CID from the canonical bytes and require equality with `cfg.policy_cid`. Same shape as `resolve_rewards_policy` at commit time.

**2. Tool mutates the policy JSON before hashing and relies on `serde_json::to_vec` for canonicalisation**

`tools/seed_bubl_rewards_upgrade.rs:116-135`:

```rust
if let Some(obj) = doc.as_object_mut() {
    obj.insert("asset_id".into(), serde_json::Value::String(hex::encode(asset_id)));
}
let bytes = serde_json::to_vec(&doc).context("canonicalize policy json")?;
let policy = validate_rewards_policy(&bytes).context(...)?;
let hash = policy_hash(&policy).context(...)?;
```

Two problems chained:
- `serde_json::to_vec` on a `Value` is not JCS-canonical. PR #2832 already flagged `canonical_policy_bytes` as non-canonical (`no key sorting despite doc claim`). Different runs / different serde_json versions can produce different bytes, which changes `policy_hash` and `policy_cid`, so the same source file yields different on-chain identifiers over time. Anyone else re-canonicalising the stored document may not recompute the same hash.
- The `asset_id` field is injected post-parse. If the rewards-policy schema is or ever becomes `additionalProperties: false`, `validate_rewards_policy` here silently succeeds while a fresh validator elsewhere rejects. Also: `asset_id` binding is not part of the on-chain enforcement path — `apply_reward_claim` uses `data.token_id` from the tx, not any field embedded in the document.

Either use the JCS canonicaliser from the crate (once fix #5 from PR #2832 lands), or drop the `asset_id` injection and treat the raw file bytes as canonical. Right now the tool advertises "policy document → hash" as a pure function but it's neither pure nor canonical.

**3. Legacy-projection resolver opens Curve/Governance upgrade paths, not just Rewards**

`resolve_sovereign_asset_for_upgrade` is called unconditionally from `apply_asset_module_upgrade` — the same function that handles `Curve`, `Rewards`, and `Governance` module upgrades. Combined with the projection producing `AssetAuthority::Creator { key_id: token.creator.key_id }`, this means anyone with the creator key of *any* legacy `TokenContract` can now:

- Enable Curve on the projected asset, which fabricates an `AssetLaunchPayloadV1` with hardcoded `treasury_bps: 2_000` (line 192) and forces `supply_mode = Elastic` (line 180) — overwriting whatever the underlying `TokenContract.max_supply` implies.
- Enable Governance on the projected asset, moving `AssetAuthority` to `Governance { module_ref }` on the first `put_sovereign_asset`, permanently locking the creator out of subsequent module upgrades.

The PR body scopes to "legacy testnet BUBL rewards"; the code path is wider than that. If Rewards is the only intended migration, gate the fallback to `AssetUpgradeModule::Rewards` (or to a specific allow-list of legacy asset ids):

```rust
let mut asset = if matches!(payload.module, AssetUpgradeModule::Rewards(_)) {
    resolve_sovereign_asset_for_upgrade(mutator, &payload.asset_id)?
} else {
    mutator.get_sovereign_asset(&payload.asset_id)?
        .ok_or_else(|| TxApplyError::InvalidType("asset not found".to_string()))?
};
```

---

### P2

**4.** `enable_rewards` doesn't validate that `spend_delegate_key_id` is a registered on-chain identity. A garbage delegate keeps writing but no reward claim can pass `data.from == rewards_state.spend_delegate_key_id` — silent failure for every future claim until rotate. Add an `mutator.get_identity(&blake3(...))?` presence check.

**5.** `seed_bubl_rewards_upgrade.rs:176` loads the delegate's *private* keystore only to read `delegate.public_key.key_id`. The private key is not used. Read the identity JSON directly (`ZhtpIdentity::from_serialized`) and pull the public key — no reason to touch the delegate's SK from an operator tool.

**6.** `apply_asset_module_upgrade` doesn't emit any tracing when a legacy `TokenContract` gets promoted into the `assets/` sled for the first time. First-time materialisation of a legacy asset is a significant state transition and should show up in journalctl / audit logs. Add `tracing::info!(asset_id=%hex, "materialised legacy sovereign asset")` in `resolve_sovereign_asset_for_upgrade` on the projection branch.

**7.** `seed_bubl_rewards_upgrade.rs:155-160` builds a placeholder `Signature { signature: Vec::new(), public_key: PublicKey::new(keypair.public_key.dilithium_pk), ... }` before signing. `signing_hash()` doesn't cover `signature.public_key`, and `tx.signature = sig` immediately overwrites it. Use `Signature::default()` (same shape as `zhtp/src/api/handlers/rewards/settlement.rs`) — the current pattern reads like it matters and doesn't.

**8.** No unit or integration tests for `resolve_sovereign_asset_for_upgrade` — neither branch (sled hit, projection fallback) nor the "legacy asset cannot be upgraded" error. A single testnet broadcast is not a regression harness.

**9.** After the first upgrade, `SovereignAsset.total_supply` is a snapshot from projection time. Subsequent `TokenMint` on the underlying `TokenContract` won't refresh the SovereignAsset row (that's already true post-SA-3, so it may be intentional — but worth noting: any API surface reading `SovereignAsset.total_supply` for a legacy asset will drift from the token-contract truth after the D3 upgrade).

**10.** `AssetLaunchPayloadV1::initial_curve_state()` is called in the Curve branch of `apply_asset_module_upgrade` for legacy assets, but no path validates that the fabricated payload's `treasury_bps: 2_000` is a reasonable default for a *pre-existing* asset — this synthetic launch record has no user-signed provenance. Not reachable from the tool this PR ships, but the code is in the same function.

---

### Verdict

Ship-able for D3 with (1) fixed — an on-chain policy-hash/CID cross-check is the difference between "operator makes a typo" and "asset rewards module bricked forever." (2) and (3) can be follow-ups if the maintainer wants to keep this PR tight to what testnet already ran. The tool is otherwise straightforward operator plumbing.

PR #2870, opened 2026-07-13
fix(economy): handle invalid analytics timestamps

### LINKED ISSUE #1544: ECONOMY-004: Fix chrono timestamp crash vulnerability
## Description
Chrono timestamp parsing uses unwrap() which panics on out-of-range timestamps.

**Location:** `lib-economy/src/wallets/transaction_history.rs:697-698, 716`

**Current Code:**
```rust
chrono::DateTime::from_timestamp(tx.timestamp as i64, 0).unwrap()
```

## Risk
Malicious or corrupted timestamp can crash node during analytics.

## Acceptance Criteria
- [ ] Replace unwrap() with proper error handling
- [ ] Use ok_or() pattern for invalid timestamps
- [ ] Log invalid timestamps for investigation
- [ ] Skip or default invalid timestamp entries
- [ ] Add tests for edge case timestamps

## Technical Notes
- from_timestamp returns None for dates before 1970 or after 262143
- Consider using checked_add for timestamp arithmetic

## Dependencies
Parent: #1540

theinversionlabs:
## Problem

Transaction analytics converted `u64` timestamps with `as i64` and then unwrapped `chrono::DateTime::from_timestamp(...)`.

Malformed or out-of-range timestamps could therefore either wrap during conversion or panic during analytics processing.

## Fix

- Convert timestamps with `i64::try_from`.
- Handle invalid Chrono timestamps without panicking.
- Log invalid values for investigation.
- Skip only monthly and hourly time-series aggregation.
- Preserve timestamp-independent category, volume, fee, and failure analytics.

## Tests

Added regression coverage for:

- a valid timestamp continuing to populate monthly and hourly analytics;
- `u64::MAX` not panicking;
- invalid timestamps remaining included in category totals;
- invalid failed transactions remaining included in failure analysis;
- invalid timestamps not creating time-series buckets.

## Verification

- `git diff --check` — passed
- `cargo check -p lib-economy` — passed
- focused source formatting check — passed
- `cargo test -p lib-economy transaction_history` — blocked by 79 pre-existing upstream test compilation errors, reproduced with this patch stashed; the patch adds no new compile errors

Fixes #1544

### DIFF
--- a/lib-economy/src/wallets/transaction_history.rs
+++ b/lib-economy/src/wallets/transaction_history.rs
@@ -739,23 +739,36 @@ impl TransactionHistoryManager {
 
         // Process all transactions
         for tx in &self.transactions {
-            // Category counts and volumes
+            // Category analytics are timestamp-independent.
             *transaction_counts.entry(tx.category.clone()).or_insert(0) += 1;
             *volume_by_category
                 .entry(tx.category.clone())
                 .or_insert(0u128) += tx.amount;
             *fees_by_category.entry(tx.category.clone()).or_insert(0u128) += tx.fees;
-
+            // Failure analytics are also timestamp-independent.
+            if let TransactionStatus::Failed { reason } = &tx.status {
+                *failure_analysis.entry(reason.clone()).or_insert(0) += 1;
+            }
+            // Only time-series analytics require a valid timestamp.
+            let timestamp = match i64::try_from(tx.timestamp) {
+                Ok(timestamp) => timestamp,
+                Err(_) => {
+                    info!(
+                        "Skipping time-series analytics for out-of-range transaction timestamp: {}",
+                        tx.timestamp
+                    );
+                    continue;
+                }
+            };
+            let Some(dt) = chrono::DateTime::from_timestamp(timestamp, 0) else {
+                info!(
+                    "Skipping time-series analytics for invalid transaction timestamp: {}",
+                    tx.timestamp
+                );
+                continue;
+            };
             // Monthly trends
-            let month_key = format!(
-                "{}-{:02}",
-                chrono::DateTime::from_timestamp(tx.timestamp as i64, 0)
-                    .unwrap()
-                    .year(),
-                chrono::DateTime::from_timestamp(tx.timestamp as i64, 0)
-                    .unwrap()
-                    .month()
-            );
+            let month_key = format!("{}-{:02}", dt.year(), dt.month());
 
             let monthly_data =
                 monthly_trends
@@ -778,9 +791,7 @@ impl TransactionHistoryManager {
                 .or_insert(0) += 1;
 
             // Daily patterns
-            let hour = chrono::DateTime::from_timestamp(tx.timestamp as i64, 0)
-                .unwrap()
-                .hour() as u8;
+            let hour = dt.hour() as u8;
             let daily_data = daily_patterns.entry(hour).or_insert(DailyActivityData {
                 hour,
                 transaction_count: 0,
@@ -790,11 +801,6 @@ impl TransactionHistoryManager {
 
             daily_data.transaction_count += 1;
             daily_data.total_volume += tx.amount;
-
-            // Failure analysis
-            if let TransactionStatus::Failed { reason } = &tx.status {
-                *failure_analysis.entry(reason.clone()).or_insert(0) += 1;
-            }
         }
 
         // Calculate averages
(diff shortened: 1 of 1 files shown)


theinversionlabs (commented):
CC @unwel — requesting review on this PR for #1544 (invalid timestamp panic fix).

theinversionlabs:
Correction: requesting review from @umwelt. The prior mention of @unwel was a typo.

lib-economy/src/wallets/transaction_history.rs
```diff
@@ -1030,6 +1036,110 @@ mod tests {
...
+    /// Regression test for issue #1544: invalid timestamps must not panic.
+    /// u64::MAX is an out-of-range timestamp that must be skipped gracefully.
+    #[tokio::test]
+    async fn test_invalid_timestamp_no_panic() {
+        let node_id = [1u8; 32];
+        let mut manager = TransactionHistoryManager::new(node_id);
+
+        let mut tx =
+            create_payment_transaction([1u8; 32], [2u8; 32], 500, Priority::Normal).unwrap();
+        tx.timestamp = u64::MAX;
```
umwelt, line 1049: This doesn't reproduce the bug. `u64::MAX as i64` wraps to `-1`, and `from_timestamp(-1)` returns `Some(1969-12-31)` — so on the **pre-fix** code the `.unwrap()` succeeds and nothing panics. It just files the tx under a 1969-12 bucket.

I checked against chrono:

```
ts=18446744073709551615  as i64=-1                   from_timestamp=Some("1969-12-31 23:59:59 UTC")
ts=10000000000000        as i64=10000000000000       from_timestamp=None
```

So `test_invalid_timestamp_no_panic` would have passed before this fix. The values that actually panic are ones that survive the cast but exceed chrono's range (anything above ~8.2e12).

Can you switch this to something like `10_000_000_000_000`? That one genuinely panics on the old code, which is what a regression test for #1544 should show. Same for the other two tests using `u64::MAX`.

lib-economy/src/wallets/transaction_history.rs
```diff
@@ -739,23 +739,36 @@ impl TransactionHistoryManager {
...
+                Ok(timestamp) => timestamp,
+                Err(_) => {
+                    info!(
+                        "Skipping time-series analytics for out-of-range transaction timestamp: {}",
+                        tx.timestamp
+                    );
+                    continue;
+                }
+            };
+            let Some(dt) = chrono::DateTime::from_timestamp(timestamp, 0) else {
```
umwelt, line 763: This `let-else` is the half of the fix that removes the actual panic, and no test reaches it.

Every invalid-timestamp test uses `u64::MAX`, which fails at `i64::try_from` on line 753 and `continue`s before ever calling `from_timestamp`. So this branch has zero coverage. Fixing the test values (see comment on line 1049) covers it.

lib-economy/src/wallets/transaction_history.rs
```diff
@@ -739,23 +739,36 @@ impl TransactionHistoryManager {
...
+            // Failure analytics are also timestamp-independent.
+            if let TransactionStatus::Failed { reason } = &tx.status {
+                *failure_analysis.entry(reason.clone()).or_insert(0) += 1;
+            }
+            // Only time-series analytics require a valid timestamp.
+            let timestamp = match i64::try_from(tx.timestamp) {
+                Ok(timestamp) => timestamp,
+                Err(_) => {
+                    info!(
+                        "Skipping time-series analytics for out-of-range transaction timestamp: {}",
```
umwelt, line 757: `info!` is too quiet for what is a data-integrity problem, and this fires once per bad tx *per `compute_analytics()` call* — so one poisoned record logs on every invocation, forever.

Suggest `warn!`, and count the skips in the loop and log once after it rather than per-record.

lib-economy/src/wallets/transaction_history.rs
```diff
@@ -1030,6 +1036,110 @@ mod tests {
...
+        let node_id = [1u8; 32];
+        let mut manager = TransactionHistoryManager::new(node_id);
+
+        let mut tx =
+            create_payment_transaction([1u8; 32], [2u8; 32], 200, Priority::Normal).unwrap();
+        tx.timestamp = u64::MAX;
+        manager.add_transaction(tx).await.unwrap();
+
+        // Manually set status to Failed on the stored record.
+        manager.transactions[0].status = TransactionStatus::Failed {
```
umwelt, line 1092: Nit: this pokes the internal vec directly, which only compiles because the test is in the same module. Couples the test to the field layout — could you use the status-transition API instead?

umwelt (changes requested):
The production fix is correct — I traced the loop and the reordering is sound: category, volume, fee and failure analytics are all preserved above the `continue`, and the post-loop aggregations still read from `self.transactions`, so only the time series is skipped. That part I'm happy with.

The problem is the tests. `u64::MAX` doesn't reproduce the panic in #1544, so the regression test passes on the broken code too, and the branch that actually fixes the panic never gets executed. Details inline.

Also worth saying plainly: none of these tests have been run (the crate's test target doesn't compile), so right now nothing in this PR is demonstrated.

umwelt:
Filed the follow-ups from the review:

- #2871 — `lib-economy` test target doesn't compile (79 errors). **This is what's blocking you.** Until it's fixed, the tests here can't run locally or in CI, so the regression coverage is unverifiable.
- #2872 — timestamps aren't validated at ingest (the root cause; this PR hardens the reader, not the writer)
- #2873 — same `as i64` truncation in `reward_claim` day/week bucketing
- #2874 — analytics output is silently self-inconsistent when records are skipped

Suggest we hold this until #2871 lands so the tests can actually be exercised. In the meantime the two things worth changing here are the test timestamp values (`u64::MAX` doesn't reproduce the panic) and the `info!` → `warn!` + aggregation — both inline above.

The fix itself is good; it's just not demonstrated yet.

umwelt:
#2876 fixes #2871 — `lib-economy`'s test targets compile again and all 180 tests pass, including `cargo test -p lib-economy --lib transaction_history`.

Once that merges you'll be able to actually run the tests here. The two changes still worth making in this PR are the timestamp values (`u64::MAX` doesn't reproduce the panic — it wraps to `-1` and `from_timestamp` happily returns `Some(1969-12-31)`) and `info!` → `warn!` with aggregation.

umwelt:
**Heads-up from review of #2894:** that PR rewrites `compute_analytics` on current `development` with non-panicking timestamp handling + `excluded_invalid_timestamps` reconcile counter, and correctly keeps failure_analysis for invalid-timestamp failed txs.

#2894 effectively supersedes this PR. Recommend closing this in favor of #2894 (or rebasing this onto #2894 if anything unique remains).

umwelt:
Superseded by the timestamp hygiene stack on `development`:

- #2892 — reject invalid timestamps at history ingest (`Closes #2872`)
- #2893 — `i64::try_from` on reward day/week bucketing (`Closes #2873`)
- #2894 — `excluded_invalid_timestamps` analytics counter (`Closes #2874`)

Those PRs also cover the analytics/chrono crash surface this branch targeted (#1544 family). Closing as duplicate/superseded — please abandon the `fix/1544-chrono-timestamp-validation` branch.

Issue #2875, opened 2026-07-13
calculate_fee: unchecked amount * DEFAULT_DAO_FEE_RATE can overflow u128


umwelt:
## Problem

`EconomicModel::calculate_fee` (lib-economy/src/models/economic_model.rs:140):

```rust
let dao_fee = (amount * crate::DEFAULT_DAO_FEE_RATE) / 10000;
```

`amount` is `u128` and `DEFAULT_DAO_FEE_RATE` is `100`, so any `amount` above `u128::MAX / 100` overflows:

```
overflow threshold: any amount > 3402823669209384634633746074317682114
```

In debug that's a panic; in release it wraps and produces a **silently wrong fee**.

## How it surfaced

Found while fixing #2871. The edge-case tests exercise "maximum amount" transactions; I capped them at `u64::MAX as u128` (preserving their pre-migration values) precisely because pushing to `u128::MAX` trips this.

## Ask

Use `checked_mul` / `saturating_mul`, or reorder as `(amount / 10000) * RATE` with remainder handling. Then raise the edge-case tests to genuine `u128::MAX` so the boundary is actually covered.

Related: #2871

PR #2876, opened 2026-07-13
fix(economy): repair lib-economy test targets after u64→u128 migration

### LINKED ISSUE #2871: lib-economy test target does not compile (79 errors) — all tests in the crate are dead
## Problem

`cargo test -p lib-economy` fails to compile with ~79 errors. These are pre-existing and unrelated to any one change — they reproduce on a clean `development` checkout.

Consequence: **no test in `lib-economy` currently runs.** If CI is green, CI is not building this crate's tests, so every test in it is decorative.

This surfaced in #2870, where the author added five regression tests and could not execute any of them:

> `cargo test -p lib-economy transaction_history` — blocked by 79 pre-existing upstream test compilation errors, reproduced with this patch stashed

## Why it matters

Any fix landing in `lib-economy` right now ships with unverifiable test coverage. #2870 fixes a real panic but cannot demonstrate it.

## Asks

- Fix the compilation errors in the test target
- Confirm whether CI builds `lib-economy` tests at all; if not, wire it in so this can't rot again

Blocks: #2870

umwelt:
Closes #2871. Unblocks #2870.

## Problem

All five of `lib-economy`'s test targets failed to compile. Amounts, fees and balances were migrated from `u64` to `u128` in production, and the tests were never updated. Nothing in the crate has been tested since that migration — CI has been green because it never built these targets.

This surfaced in #2870, where the author added regression tests and could not run any of them.

## Two real bugs were hiding behind the broken build

**`get_economic_stats()` panicked on every call.** `EconomicModel::max_supply` became `u128::MAX`, but the function serialises it through `serde_json::json!`, and serde_json numbers top out at `u64`. Every invocation panicked with `Error("number out of range")`. `u128` amounts are now emitted as decimal strings. The function isn't wired to any HTTP handler, so there's no API compatibility concern.

**A wallet assertion was simply wrong.** `WalletBalance::can_afford` is a plain `available_balance >= amount` check, so `assert!(!wallet.can_afford(1))` on a max-balance wallet could never hold. The comment claimed it "would overflow" — it doesn't.

Neither was reachable while the tests didn't compile.

## Scope note

Test amounts are capped at `u64::MAX as u128` rather than raised to `u128::MAX`, which preserves the values these tests were written against pre-migration. Pushing them to `u128::MAX` trips an unchecked multiply in `calculate_fee` — filed separately as #2875, along with the follow-up to raise the boundary once it's fixed.

Unrelated rustfmt noise has been kept out; the crate was not rustfmt-clean before this change.

## Verification

```
cargo test -p lib-economy
  lib          128 passed
  edge_case     15 passed
  integration   14 passed
  performance   10 passed
  stress        13 passed
```

180 passing, 0 failing, across all five targets. `cargo test -p lib-economy --lib transaction_history` — the module #2870 touches — runs clean.

### DIFF
--- a/lib-economy/src/models/economic_model.rs
+++ b/lib-economy/src/models/economic_model.rs
@@ -61,7 +61,7 @@ impl EconomicModel {
             quality_multiplier: 0.1, // Minimal quality bonus (infrastructure focus)
             uptime_multiplier: 0.05, // Minimal uptime bonus (reliability expected)
             inflation_rate: 0.0,     // ZERO inflation (stable utility token)
-            max_supply: u128::MAX,    // UNLIMITED supply (like internet capacity)
+            max_supply: u128::MAX,   // UNLIMITED supply (like internet capacity)
             current_supply: 0,       // Start from zero, mint as needed
             burn_rate: 0.0,          // NO BURNING (utility, not speculation)
             dao_treasury,            // Treasury interface for economics
@@ -126,7 +126,12 @@ impl EconomicModel {
     }
 
     /// Calculate transaction fees including mandatory DAO fee for UBI/DAO allocations
-    pub fn calculate_fee(&self, tx_size: u64, amount: u128, priority: Priority) -> (u128, u128, u128) {
+    pub fn calculate_fee(
+        &self,
+        tx_size: u64,
+        amount: u128,
+        priority: Priority,
+    ) -> (u128, u128, u128) {
         // NETWORK INFRASTRUCTURE FEE (covers bandwidth, storage, compute)
         let base_fee = (tx_size as u128) * 1; // 1 token per byte (minimal infrastructure cost)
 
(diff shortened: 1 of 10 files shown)


umwelt:
Ran an independent review over this. It found two things I got wrong; both now fixed in d2a56f44.

**1. Two test caps were unjustified and cost coverage.** I capped `test_maximum_amount_transactions` and `test_maximum_balance_operations` at `u64::MAX as u128`, claiming `u128::MAX` trips the unchecked multiply in `calculate_fee`. That's true of `calculate_dao_fee_distribution`, but **not of these two tests** — `Transaction::new_payment` routes through `calculate_total_fee` → the *saturating* `calculate_dao_fee` (fee_calculation.rs:25), and the balance test never touches a fee path at all. Both pass at true `u128::MAX`; I verified it. Capping them meant that if someone reverted `calculate_dao_fee`'s `saturating_mul` back to `*`, the test would no longer catch it. Restored to `u128::MAX`.

The caps in `fee_processing.rs` stay — `calculate_dao_fee_distribution` really does use an unchecked `dao_fees * PERCENTAGE`, so `u64::MAX` is the honest ceiling there until #2875 lands. (Those two were also *raised*, from `u64::MAX/200` and `u64::MAX/1000`.)

**2. I fixed the `json!`/u128 panic in the one function only its own test calls, and left it in two public API handlers.** `ZhtpServer::get_stats()` (server.rs:353) and `get_economic_status()` (server.rs:442) feed `max_supply` — `u128::MAX` by default — straight into `json!`, so both panic with `number out of range` on **every call**. Same bug, live API surface. Now emitting `u128` as decimal strings there too. No consumer can have relied on the number shape, since the handlers panicked unconditionally.

Still green: 182 tests, 5 targets, `cargo check -p zhtp` clean.

umwelt:
Follow-up 25a1dbb3 closes **#2875**: `saturating_mul` in `EconomicModel::calculate_fee` and `calculate_dao_fee_distribution`; distribution overflow test raised to `u128::MAX`. All lib-economy unit tests (129) + integration tests green; `cargo check -p zhtp` clean.

Issue #2787, opened 2026-07-04
SA-7: AssetManifestUpdate + DHT pin validation


umwelt:
Mandatory manifest shareability gate.

**Deliverables:**
- `AssetManifestUpdate` tx
- Launch/manifest-update hash verification vs DHT
- Manifest resolved in `GET /api/v1/assets/{id}`

Depends on SA-3, SA-6.

Parent epic: #2780
Design: docs/arch/sovereign-asset.md §9 (SA-7)

umwelt:
**Status audit (2026-07-14): PARTIAL — not closable yet.**

### Shipped
- `AssetManifestUpdate` tx + executor — `asset_tx.rs`, `apply_asset_manifest_update()` in `sovereign_asset.rs`
- Authority-gated updates — `verify_authority_proof`
- Manifest resolved on **read** — `zhtp/src/api/handlers/assets/mod.rs` (`resolve_asset_manifest`, `manifest_resolved`)
- Manifest CID/hash in list items — `AssetListItem` in same handler

### Still blocking close
1. **No consensus DHT pin validation at launch** — `apply_asset_launch` stores cid/hash only; no DHT fetch/verify
2. **No DHT gate at manifest update** — executor updates on-chain fields without pin check
3. **Invariant #3 not enforced in mempool/validation** — hard reject on pin failure is API-layer best-effort only

Spec requires validator DHT fetch + `manifest_hash` cross-check before accept (§5, invariant #3).

umwelt:
**Status re-audit (2026-07-17): closable for v1 — SA-7 shipped in #2890.**

| Deliverable | Status |
|-------------|--------|
| `AssetManifestUpdate` tx + executor | Done |
| Launch/update pin gate | Done — `verify_manifest_pin` in `apply_asset_launch` / `apply_asset_manifest_update` |
| Consensus-safe design | Local `dht_pins` cache is **advisory only** (never gates block validity); required fields when gate active: non-zero `manifest_cid` / `manifest_hash` |
| Optional mempool strictness | `check_manifest_pin_admission` |
| CLI auto-pin | Web4 content/manifest before launch |
| `GET /api/v1/assets/{id}` resolve | Done |

Full DHT content fetch as a **consensus** hard gate was intentionally not used (would break deterministic apply). That matches the SA-7 PR review decision.

Closing as done for the SA-7 v1 scope. Follow-ups (stricter mempool-only admission in production gateways) can be separate issues.

PR #2888, opened 2026-07-16
feat(sa-8): deprecate /api/v1/token read paths (#2788)

### LINKED ISSUE #2788: SA-8: Deprecate legacy token API + remove dual read paths
Remove `/api/v1/token/*` special-casing and legacy projection after testnet assets re-launched via `AssetLaunch`.

Depends on SA-1..SA-7 + testnet re-launch.

Parent epic: #2780
Design: docs/arch/sovereign-asset.md §9 (SA-8)

umwelt:
## Summary

Implements SA-8 (#2788): deprecate legacy `/api/v1/token` read/create endpoints in favor of `/api/v1/assets/*`.

### Changes

- **`zhtp/src/api/handlers/token/mod.rs`**
  - Return `410 Gone` for deprecated read/create paths (`list`, `{id}`, `{id}/balance/{address}`, `balances/{address}`, `symbol/available/{symbol}`, `create`) with successor pointers to `/api/v1/assets/*`
  - Remove sovereign projection shim helpers (`dao_token_contract_symbols_upper`, `include_sovereign_asset_in_token_api`, `sovereign_asset_to_*`)
  - Keep mutation handlers: `transfer`, `mint`, `nonce`, `burn`

- **`zhtp/src/api/handlers/assets/mod.rs`**
  - Add `GET /api/v1/assets/symbol/available/{symbol}` (checks sovereign assets + legacy token contracts)

- **`zhtp-cli/src/commands/token.rs`**
  - Migrate `list` → `GET /api/v1/assets`
  - Migrate `info` → `GET /api/v1/assets/{id}`
  - Migrate `balance` → `GET /api/v1/assets/{id}/balances/{address}`

### Verification

- `cargo check -p zhtp -p zhtp-cli` ✅

Closes #2788

### DIFF
--- a/zhtp/src/api/handlers/token/mod.rs
+++ b/zhtp/src/api/handlers/token/mod.rs
@@ -1,22 +1,36 @@
 //! Token Handler Module
 //!
-//! Mutation API for custom token mint, transfer, burn, and nonce lookup.
-//! Read/discovery paths were deprecated in SA-8 — use `/api/v1/assets/*` instead.
+//! Token mint/transfer/burn/nonce plus legacy read/discovery paths used by mobile.
+//!
+//! # SA-8 progressive deprecation
+//!
+//! Prefer `/api/v1/assets/*` for new clients. Legacy `/api/v1/token/*` **reads and create**
+//! remain live (200) with `Deprecation` / `Sunset` / `Link` headers until a shipped mobile
+//! release no longer depends on them. Do **not** return 410 Gone on these hot wallet paths —
+//! mobile portfolio/send/receive still call them. Hard removal requires a coordinated mobile
+//! ship first.
 
 use anyhow::Result;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use serde_json::json;
+use std::collections::HashSet;
 use std::sync::Arc;
 use tokio::sync::RwLock;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
 // ZHTP protocol imports
 use lib_protocols::types::{ZhtpMethod, ZhtpRequest, ZhtpResponse, ZhtpStatus};
 use lib_protocols::zhtp::ZhtpRequestHandler;
 
 // Blockchain imports
+use lib_blockchain::contracts::sovereign_asset::{AssetIdSource, SovereignAsset};
 use lib_blockchain::transaction::Transaction;
+use lib_blockchain::types::transaction_type::TransactionType;
 use lib_blockchain::{Blockchain, BlockchainQuery};
+use lib_crypto::types::keys::PublicKey;
+
+/// Soft sunset for legacy token routes (RFC 8594). Extend if mobile has not migrated.
+const TOKEN_API_SUNSET: &str = "Tue, 01 Dec 2026 00:00:00 GMT";
 
 /// Helper function to create JSON responses
 fn create_json_response(data: serde_json::Value) -> Result<ZhtpResponse> {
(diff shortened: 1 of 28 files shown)


zhtp/src/api/handlers/token/mod.rs
```diff
@@ -37,26 +32,20 @@ fn create_error_response(status: ZhtpStatus, message: String) -> ZhtpResponse {
     ZhtpResponse::error(status, message)
 }
 
+/// SA-8: legacy `/api/v1/token` read/create routes return 410 Gone with successor pointer.
+fn token_deprecated_gone(successor: &str) -> ZhtpResponse {
```
umwelt: The 410 is unconditional — no client-version gate, no sunset window, no negotiation. Per the project rule (never break mobile endpoints; deprecate progressively), a hard `Gone` on read paths is the one mechanism that's ruled out here.

A progressive path that satisfies the rule: keep serving the old reads but add a `Deprecation` / `Sunset` response header (or a `deprecated: true` field) pointing at the successor, ship that, confirm mobile has migrated off them, *then* flip to 410 in a later release. Or, if you want an immediate cutover, `307`-redirect the ones that have a 1:1 successor rather than `410`-ing them — a redirect keeps working clients working.

At minimum this needs the mobile-usage question answered before merge.

umwelt (commented):
SA-8 review. The migration is mostly well-built — successors exist and route, `symbol/available` is a proper superset (checks sovereign assets *and* legacy token contracts), the CLI is migrated, and the mutation handlers (`transfer`/`mint`/`nonce`/`burn`) are correctly kept. But this returns an **unconditional `410 Gone` on read endpoints, including balance reads, with no progressive gate**, and that runs straight into the project's hardest compat rule: never break mobile-app endpoints, deprecate progressively.

`410 Gone` is a *permanent* removal, not a soft deprecation — a client hitting it on deploy day has no negotiation window. Three specific problems below. I can't see the mobile repo from here, so the load-bearing question — **does the mobile wallet call any of `/api/v1/token/{id}/balance/{address}`, `/token/balances/{address}`, `/token/list`, or `/token/{id}`?** — has to be answered before this merges. If yes, this breaks the wallet on the first node that gets the new binary.

Note also this PR stacks SA-6 (#2887) and SA-7 (#2787), which aren't merged yet; the review below is scoped to the SA-8 commit (e9a84af0).

zhtp/src/api/handlers/token/mod.rs
```diff
@@ -981,10 +396,12 @@ impl ZhtpRequestHandler for TokenHandler {
...
                     ))
                 } else {
-                    self.handle_check_symbol_available(symbol).await
+                    Ok(token_deprecated_gone(&format!(
+                        "GET /api/v1/assets/symbol/available/{symbol}"
+                    )))
                 }
             }
-            // GET /api/v1/token/balances/{address} - Get all token balances for an address
+            // GET /api/v1/token/balances/{address} — deprecated (SA-8)
```
umwelt: `GET /api/v1/token/balances/{address}` — all balances for an address in one call — has **no 1:1 successor**. The pointer here tells the client to do `GET /api/v1/assets` and then `GET /api/v1/assets/{asset_id}/balances/{address}` per asset, i.e. one request becomes an N+1 fan-out. For a mobile wallet rendering a portfolio that's a real regression even *after* migration, not just a rename.

Worth adding a batched successor (e.g. `GET /api/v1/assets/balances/{address}` returning all non-zero balances) so the migrated client keeps a single round-trip.

zhtp/src/api/handlers/token/mod.rs
```diff
@@ -1012,26 +430,28 @@ impl ZhtpRequestHandler for TokenHandler {
...
+                    Ok(token_deprecated_gone("GET /api/v1/assets"))
                 } else {
-                    self.handle_get_token_info(token_id).await
+                    Ok(token_deprecated_gone(&format!(
+                        "GET /api/v1/assets/{token_id}"
+                    )))
                 }
             }
-            // GET /api/v1/token/{id}/balance/{address}
+            // GET /api/v1/token/{id}/balance/{address} — deprecated (SA-8)
```
umwelt: The successor `/api/v1/assets/{id}/balances/{address}` returns `{ asset_id, address, balance, nonce }` (assets/mod.rs:540), whereas the old token balance response keyed on `token_id`. So a client that migrates to the new path but still reads `token_id` gets `null`. A field rename inside the successor means "just point at the new URL" isn't sufficient — the shape changed too. Call this out in the migration note, or alias `token_id` alongside `asset_id` in the asset balance response for one release.

umwelt:
## Review response — progressive deprecation (blocks unconditional 410)

Mobile still uses these hot paths; hard `410 Gone` is a merge blocker. Addressed in the latest commit:

### 1. No unconditional 410 on wallet reads/create
- Restored live handlers for `list`, `info`, per-token balance, **bulk** `GET /token/balances/{address}`, symbol available, and `POST /token/create` (AssetLaunch shim).
- Responses stay **200** with soft-deprecation headers: `Deprecation: true`, `Sunset: Tue, 01 Dec 2026 00:00:00 GMT`, `Link: <successor>; rel="successor-version"`, `X-ZHTP-Successor`.
- Hard 410 only after a shipped mobile release no longer depends on these routes.

### 2. Bulk portfolio successor (no N+1)
- Kept `GET /api/v1/token/balances/{address}` working (primary mobile path).
- Added **`GET /api/v1/assets/balances/{address}`** sharing the same portfolio builder so new clients have a 1:1 bulk successor.

### 3. Dual-write `token_id` + `asset_id`
- Legacy token list/info/balance/bulk responses include both fields (same hex value).
- Assets single-balance response dual-writes `token_id` alongside `asset_id`.

### Policy going forward
Never hard-remove or 410 mobile-facing read APIs without a dual-path/sunset window and confirmed mobile ship. Prefer headers + dual fields over status-code breaks.

CLI already prefers `/assets/*` (fine for progressive migration of non-mobile clients). Mutations (`mint`/`transfer`/`nonce`) unchanged.

Ready for re-review.

Issue #2785, opened 2026-07-04
SA-5: CurveModule as asset sub-state


umwelt:
Fold `BondingCurveToken` into `SovereignAsset.curve` sub-state.

**Deliverables:**
- `asset_curve/` sled tree
- Curve buy/sell txs keyed by `asset_id`
- Stop writing standalone `BondingCurveToken` for new launches
- Projection shim for legacy CBE

Depends on SA-3.

Parent epic: #2780
Design: docs/arch/sovereign-asset.md §9 (SA-5)

umwelt:
**Status audit (2026-07-14): PARTIAL — not closable yet.**

### Shipped
- `CurveModuleState` types — `lib-blockchain/src/contracts/sovereign_asset/state.rs`
- `asset_curve/` sled tree — `lib-blockchain/src/storage/sled_store.rs`
- Curve init on `AssetLaunch` / `AssetModuleUpgrade` — `sovereign_asset.rs`

### Still blocking close
1. **Buy/sell/graduate still use global CBE state** — `executor.rs` (`apply_buy_cbe`/`apply_sell_cbe`), not `asset_curve/` by `asset_id`
2. **Tx payloads lack `asset_id`** — `lib-types/src/bonding_curve.rs` (`BondingCurveBuyTx` etc.)
3. **`BondingCurveToken` not retired** — registry + projection still active (`projection.rs`, `contracts.rs`)
4. **Curve execution doesn't check** `curve` module enabled on target asset

`BondingCurveDeploy` writes are disabled on legacy path only — partial.

umwelt:
**Status re-audit (2026-07-17): still PARTIAL after #2889 — not closable.**

### Shipped (#2889)
- `asset_curve/` sled + `CurveModuleState`
- Init on `AssetLaunch` / module upgrade
- Oracle / reserve reads prefer curve module after economic-rules activation (legacy `bonding_curve_token` fallback)
- Discoverability without standalone token module

### Still open (post-deploy OK)
1. Buy/sell/graduate execution still primarily on legacy CBE path (`apply_buy_cbe` / `apply_sell_cbe`) rather than pure `asset_id` → `asset_curve/`
2. Tx payloads not fully rekeyed to `asset_id`
3. Standalone `BondingCurveToken` still present for replay / pre-activation

[...]

PR #2889, opened 2026-07-16
feat(sa-5): curve module storage index and oracle read cutover (#2785)

umwelt:
## Summary

SA-5 bonding-curve module cutover on top of SA-6/SA-7 consensus modules.

**Primary (SA-5)**
- Register `asset_curve` as a native module and route curve index / reserve reads through it after economic-rules activation.
- Wire graduation oracle gate + CBE launch path to prefer curve-module reserves when active; fall back to `bonding_curve_token` otherwise (including pre-activation and missing curve state).
- Integration coverage for dual-state preference and launch discoverability without token module.

**Also on this branch (stacked consensus work already in review as #2887 / #2890)**
- SA-6 governance module + Dilithium multisig path (`governance_proof`, always-on crypto verification).
- SA-7 economic rules module + `CONSENSUS_BUILD_ID = 5` / network-id bump for the dual cutover.

## Review fixes (this push)

- **umwelt:** height-gated curve reserve preference + soft fallback when curve state is missing/invalid; dual-state unit test.
- **Copilot:** non-strict manifest pin no longer hard-fails on bad cache; MultisigApproval docs corrected for always-on signatures.

## Test plan

- [x] `cargo test -p lib-blockchain --lib resolve_reserve_prefers_curve_module_after_activation_when_dual_state`
- [x] `cargo test -p lib-blockchain --lib manifest_pin::tests`
- [x] `cargo test -p lib-blockchain --lib test_asset_launch_curve_module_discoverable_without_bonding_curve_token`
- [ ] CI green on this PR

### DIFF
--- a/lib-blockchain/src/blockchain/oracle.rs
+++ b/lib-blockchain/src/blockchain/oracle.rs
@@ -338,12 +338,23 @@ impl Blockchain {
         token_id: [u8; 32],
         block_timestamp: u64,
     ) -> Result<()> {
+        // Callers outside block validation use tip height for the curve-source gate.
+        self.validate_cbe_graduation_oracle_gate_at(token_id, block_timestamp, self.height)
+    }
+
+    /// CBE graduation oracle gate at a specific block height (replay-safe source selection).
+    pub fn validate_cbe_graduation_oracle_gate_at(
+        &self,
+        token_id: [u8; 32],
+        block_timestamp: u64,
+        block_height: u64,
+    ) -> Result<()> {
         use crate::oracle::ORACLE_PRICE_SCALE;
 
         use crate::contracts::bonding_curve::types::GRADUATION_THRESHOLD_USD;
         const MICRO_USD_PER_USD: u128 = 1_000_000;
 
-        let Some(reserve_sov) = self.resolve_cbe_graduation_reserve(token_id)? else {
+        let Some(reserve_sov) = self.resolve_cbe_graduation_reserve(token_id, block_height)? else {
             return Ok(());
         };
 
@@ -378,32 +389,53 @@ impl Blockchain {
 
     /// Resolve SOV reserve for CBE graduation oracle gating.
     ///
-    /// Prefers sovereign `asset_curve/` state when the asset has the curve flag (SA-5);
-    /// falls back to the legacy bonding-curve registry. Returns `None` when the gate
-    /// does not apply (non-CBE symbol or already graduated).
-    fn resolve_cbe_graduation_reserve(&self, token_id: [u8; 32]) -> Result<Option<u128>> {
-        use crate::contracts::sovereign_asset::CurvePhase;
+    /// After economic-rules activation, prefers sovereign `asset_curve/` when the asset
+    /// has the curve flag (SA-5). Below that height, always uses the legacy bonding-curve
+    /// registry so historical `BondingCurveGraduate` blocks replay deterministically.
+    /// Missing/unreadable curve side-table rows fall through to legacy rather than
+    /// rejecting the block. Returns `None` when the gate does not apply (non-CBE or graduated).
+    pub(crate) fn resolve_cbe_graduation_reserve(
+        &self,
+        token_id: [u8; 32],
+        block_height: u64,
+    ) -> Result<Option<u128>> {
+        use crate::contracts::sovereign_asset::{economic_rules_active, CurvePhase};
         use crate::contracts::tokens::CBE_SYMBOL;
 
-        if let Some(store) = &self.store {
-            if let Ok(Some(asset)) = store.get_sovereign_asset(&token_id) {
-                if asset.module_flags.has_curve() {
-                    if asset.symbol != CBE_SYMBOL {
-                        return Ok(None);
+        if economic_rules_active(block_height) {
+            if let Some(store) = &self.store {
+                if let Ok(Some(asset)) = store.get_sovereign_asset(&token_id) {
+                    if asset.module_flags.has_curve() {
+                        if asset.symbol != CBE_SYMBOL {
+                            return Ok(None);
+                        }
+                        match store.get_curve_module_state(&token_id) {
+                            Ok(Some(curve_state)) => {
+                                if matches!(
+                                    curve_state.phase,
+                                    CurvePhase::Graduated | CurvePhase::Amm
+                                ) {
+                                    return Ok(None);
+                                }
+                                return Ok(Some(curve_state.reserve_balance));
+                            }
+                            Ok(None) => {
+                                tracing::warn!(
+                                    token_id = %hex::encode(&token_id[..8]),
+                                    block_height,
+                                    "curve flag set but asset_curve state missing; using legacy reserve source"
+                                );
+                            }
+                            Err(e) => {
+                                tracing::warn!(
+                                    token_id = %hex::encode(&token_id[..8]),
+                                    block_height,
+                                    error = %e,
+                                    "failed to read curve module state; using legacy reserve source"
+                                );
+                            }
+                        }
                     }
-                    let curve_state = store
-                        .get_curve_module_state(&token_id)
-                        .map_err(|e| {
-                            anyhow::anyhow!("failed to read curve module state: {}", e)
-                        })?
-                        .ok_or_else(|| anyhow::anyhow!("curve module state not found"))?;
-                    if matches!(
-                        curve_state.phase,
-                        CurvePhase::Graduated | CurvePhase::Amm
-                    ) {
-                        return Ok(None);
-                    }
-                    return Ok(Some(curve_state.reserve_balance));
                 }
             }
         }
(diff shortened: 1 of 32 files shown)


lib-blockchain/src/blockchain/oracle.rs
```diff
@@ -389,6 +376,55 @@ impl Blockchain {
...
+    /// Prefers sovereign `asset_curve/` state when the asset has the curve flag (SA-5);
+    /// falls back to the legacy bonding-curve registry. Returns `None` when the gate
+    /// does not apply (non-CBE symbol or already graduated).
+    fn resolve_cbe_graduation_reserve(&self, token_id: [u8; 32]) -> Result<Option<u128>> {
+        use crate::contracts::sovereign_asset::CurvePhase;
+        use crate::contracts::tokens::CBE_SYMBOL;
+
+        if let Some(store) = &self.store {
+            if let Ok(Some(asset)) = store.get_sovereign_asset(&token_id) {
+                if asset.module_flags.has_curve() {
```
umwelt: This changes the reserve source for CBE graduation gating from the legacy bonding-curve token to `asset_curve/` module state — and it's **ungated by height, on the block-validation path**. `validate_block_cbe_graduation_gating` runs at `blockchain.rs:1023` *before* applying a block, and a gate failure `return Err(e)` **rejects the block**.

For a pure-legacy CBE (no sovereign-asset record) `get_sovereign_asset` returns `None` and it falls through to the identical legacy path — so the common case is replay-safe. The exposure is narrow but real: an asset that has **both** a curve-flagged sovereign record **and** appears in a historical `BondingCurveGraduate` block. On replay / fresh-sync with the new binary, the reserve resolves from `curve_state.reserve_balance` instead of the legacy `token.reserve_balance`; if those differ, the gate decision flips, the historical block is rejected, and the syncing node diverges. That's the g4 replay-determinism class we've hit before.

Same fix pattern as the rest of this epic: gate the source switch on an activation height so historical replay keeps using the legacy source, and only post-cutover blocks read curve-module state. If you'd rather not gate it, the merge needs a positive confirmation that no `BondingCurveGraduate` block exists in history for an asset that also carries curve-module state — i.e. the dual-state set is empty on the live chain.

lib-blockchain/src/blockchain/oracle.rs
```diff
@@ -389,6 +376,55 @@ impl Blockchain {
...
+                    if asset.symbol != CBE_SYMBOL {
+                        return Ok(None);
+                    }
+                    let curve_state = store
+                        .get_curve_module_state(&token_id)
+                        .map_err(|e| {
+                            anyhow::anyhow!("failed to read curve module state: {}", e)
+                        })?
+                        .ok_or_else(|| anyhow::anyhow!("curve module state not found"))?;
```
umwelt: Hard-fail on the block-validation path: if an asset has `has_curve()` set but `get_curve_module_state` returns `None` (partial write, or a curve flag set without the heavy state), this returns `Err` and the block is rejected — the same defensive-hard-fail-in-consensus concern I raised on the authority-timelock PR. A missing side-table row shouldn't be able to reject an otherwise-valid block; consider logging and falling through to the legacy path rather than erroring.

lib-blockchain/src/blockchain/oracle.rs
```diff
@@ -389,6 +376,55 @@ impl Blockchain {
         Ok(())
     }
 
+    /// Resolve SOV reserve for CBE graduation oracle gating.
+    ///
+    /// Prefers sovereign `asset_curve/` state when the asset has the curve flag (SA-5);
+    /// falls back to the legacy bonding-curve registry. Returns `None` when the gate
+    /// does not apply (non-CBE symbol or already graduated).
+    fn resolve_cbe_graduation_reserve(&self, token_id: [u8; 32]) -> Result<Option<u128>> {
```
umwelt: Test gap: the only new test covers the happy path (curve module discoverable without a `bonding_curve_tokens` row). Nothing exercises the case that actually matters for correctness here — an asset that has *both* a legacy token and curve-module state — and asserts the two resolve to the same reserve (or that the intended source wins). That's the test that would catch the divergence described above.

umwelt (commented):
SA-5 review, scoped to the SA-5 commit (406db5ec) — this PR also stacks the still-unmerged SA-6 (#2887) and SA-7 (#2787).

Most of this is clean, and I want to note two things I verified rather than assumed:
- The 104 lines added to `executor.rs` are almost entirely the new test plus a comment reword — `BondingCurveDeploy`/`BondingCurveGraduate` were *already* no-ops before this PR, so there's no hidden behavior change there.
- The phase mapping is equivalent: legacy `is_graduated()` = `Graduated | AMM`, and the new path returns "gate off" for `Graduated | Amm`. No phase-mapping divergence.
- The `list_curve_module_asset_ids` index reads the `asset_curve` tree with the same lazy-open pattern as the rewards index — no migration hazard.

One real concern, on the consensus path — details inline. Net: I'd want the replay question answered (or the source switch height-gated) before merge, but it's a narrow window, not a broad break.

lib-blockchain/src/manifest_pin.rs
```diff
@@ -0,0 +1,175 @@
...
+            } else {
+                tracing::warn!(
+                    manifest_cid = %hex::encode(&manifest_cid[..8]),
+                    block_height,
+                    "manifest pin gate: content not in local cache; allowing best-effort"
+                );
+                Ok(())
+            }
+        }
+    }
```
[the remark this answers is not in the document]
umwelt: Addressed in 263c972c.

Non-strict mode now treats **invalid** pin cache the same as **missing** pin: warn and allow. Only strict mode hard-fails. That removes the consensus-split risk when some nodes have a bad optional pin and others do not.

lib-blockchain/src/contracts/approval_verifier/traits.rs
```diff
@@ -114,6 +114,9 @@ pub enum ApprovalProof {
         threshold: u8,
         /// Hash of the message being signed
         message_hash: [u8; 32],
+        /// Raw Dilithium signatures (required for sovereign-asset governance at economic-rules height).
+        #[serde(default)]
+        raw_signatures: Vec<Vec<u8>>,
```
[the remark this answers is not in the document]
umwelt: Fixed in 263c972c: the MultisigApproval docs now state that `raw_signatures` is required at **every** height (always-on Dilithium path), not only after economic-rules activation.

lib-blockchain/src/lib.rs
```diff
@@ -26,6 +26,8 @@ pub mod execution;
 pub mod fees;
 mod fork_recovery; // gutted in Issue #936; kept as private to avoid orphan module errors
 pub mod genesis;
+pub mod governance_proof;
+pub mod manifest_pin;
```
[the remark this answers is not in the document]
umwelt: Updated the PR description to call out the full stack on this branch:

- **SA-5** (primary): curve module cutover / reserve preference
- **SA-6 / SA-7** (also present): governance + economic-rules modules and consensus build id, already reviewed under #2887 / #2890

lib-blockchain/src/blockchain/oracle.rs
umwelt: Addressed in 263c972c.

`resolve_cbe_graduation_reserve` prefers `asset_curve` **only when** `economic_rules_active(block_height)`. Pre-activation uses the legacy bonding-curve token path so historical replay cannot diverge mid-stream. Block validation threads `block.header.height` into `validate_cbe_graduation_oracle_gate_at`.

lib-blockchain/src/blockchain/oracle.rs
umwelt: Addressed in 263c972c.

Missing/invalid curve module state no longer hard-fails the gate: we **warn and fall through** to `bonding_curve_token` reserves. That avoids rejecting blocks when `has_curve()` is set but curve state is temporarily unavailable.

lib-blockchain/src/blockchain/oracle.rs
umwelt: Addressed in 263c972c.

Added `resolve_reserve_prefers_curve_module_after_activation_when_dual_state`, which seeds both curve-module and token-module reserves and asserts the curve path wins only after economic-rules activation.

umwelt:
## Review responses (Copilot + umwelt)

Pushed **263c972c** with fixes for all open review threads.

| Source | Finding | Fix |
|--------|---------|-----|
| **umwelt** | Curve reserve preference not height-gated | Prefer `asset_curve` only after `economic_rules_active`; gate uses block height |
| **umwelt** | Hard-fail on missing curve state | Warn + fall through to `bonding_curve_token` |
| **umwelt** | Dual-state test gap | `resolve_reserve_prefers_curve_module_after_activation_when_dual_state` |
| **Copilot** | Non-strict bad pin hard-fails | Non-strict: warn+allow (same as missing pin) |
| **Copilot** | MultisigApproval docs said height-gated signatures | Docs: required at every height |
| **Copilot** | PR scope vs description | PR body updated for SA-5 + stacked SA-6/SA-7 |

Local verification: dual-state reserve test, manifest pin tests, and curve-module launch discoverability test all pass.

PR #2890, opened 2026-07-16
feat(sa-7): manifest DHT pin gate (#2787)

umwelt:
## Summary

SA-7 manifest shareability: DHT pin cache + height-gated on-chain manifest field requirements, stacked on SA-6 governance.

**Primary (SA-7)**
- `dht_pins` sled tree + Web4/DHT → consensus pin bridge (keyed by **truncated** consensus CID).
- Apply-time pin gate: when active, require non-zero `manifest_cid` / `manifest_hash` only. Local cache is **advisory** (never gates block validity — consensus-safe).
- CLI auto-pins via `/api/v1/web4/content/manifest` before `AssetLaunch`.
- Optional `check_manifest_pin_admission` for mempool/operator strictness.

**Also on this branch**
- SA-6 governance module + always-on Dilithium multisig verification (`governance_proof`).
- `CONSENSUS_BUILD_ID = 5`.

## Review fixes

- **umwelt:** remove local-cache consensus input; injectable gate height; cross-validator determinism tests.
- **Copilot:** Single verifier always verifies Multisig proof; truncated CID mirror; no double body clone; manifest endpoint; docs.

## Test plan

- [x] `cargo test -p lib-blockchain --lib manifest_pin::tests`
- [x] `cargo test -p lib-blockchain --lib governance_proof`
- [ ] CI green

### DIFF
--- a/lib-blockchain/src/contracts/sovereign_asset/state.rs
+++ b/lib-blockchain/src/contracts/sovereign_asset/state.rs
@@ -175,6 +175,9 @@ pub fn token_creation_allowed_in_block_at_height_with_sunset(
 
 /// Epic Q1–Q3 / Q8 economic rules (mint class, treasury spend auth, reward liquidity, burn).
 /// Inactive below [`GOVERNANCE_TIMELOCK_ACTIVATION_HEIGHT`] so replay matches pre-rules history.
+///
+/// **Not** used for governance multisig signature verification — auth is replay-safe at every
+/// height and is always enforced in [`crate::governance_proof::verify_governance_multisig_proof`].
 #[inline]
 pub fn economic_rules_active(block_height: u64) -> bool {
     block_height >= GOVERNANCE_TIMELOCK_ACTIVATION_HEIGHT
--- a/lib-blockchain/src/contracts/sovereign_asset/state.rs
+++ b/lib-blockchain/src/contracts/sovereign_asset/state.rs
@@ -180,6 +180,32 @@ pub fn economic_rules_active(block_height: u64) -> bool {
     block_height >= GOVERNANCE_TIMELOCK_ACTIVATION_HEIGHT
 }
 
+/// Block height at which manifest DHT pin verification becomes consensus-active (SA-7).
+///
+/// Before this height, `AssetLaunch` / `AssetManifestUpdate` accept hash/CID commits
+/// without a local pin lookup so replay stays deterministic.
+#[cfg(test)]
+pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 0;
+#[cfg(not(test))]
+pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 80_000;
+
+/// When the pin gate is active but content is not in the validator's local DHT cache,
+/// reject the tx if true (strict); allow best-effort if false (design doc §5).
+#[cfg(test)]
+pub const MANIFEST_PIN_STRICT_MODE: bool = true;
+#[cfg(not(test))]
+pub const MANIFEST_PIN_STRICT_MODE: bool = false;
+
+#[inline]
+pub fn manifest_pin_gate_active(block_height: u64) -> bool {
+    block_height >= MANIFEST_PIN_GATE_ACTIVATION_HEIGHT
+}
+
+#[inline]
+pub fn manifest_pin_strict_mode() -> bool {
+    MANIFEST_PIN_STRICT_MODE
+}
+
 /// Default multisig threshold: floor(N/2) + 1.
 pub fn default_governance_threshold(n: usize) -> u8 {
     ((n / 2) + 1) as u8
--- a/lib-blockchain/src/contracts/sovereign_asset/state.rs
+++ b/lib-blockchain/src/contracts/sovereign_asset/state.rs
@@ -183,29 +183,86 @@ pub fn economic_rules_active(block_height: u64) -> bool {
     block_height >= GOVERNANCE_TIMELOCK_ACTIVATION_HEIGHT
 }
 
-/// Block height at which manifest DHT pin verification becomes consensus-active (SA-7).
+/// Block height at which on-chain manifest CID/hash fields become required (SA-7).
 ///
-/// Before this height, `AssetLaunch` / `AssetManifestUpdate` accept hash/CID commits
-/// without a local pin lookup so replay stays deterministic.
-#[cfg(test)]
-pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 0;
-#[cfg(not(test))]
+/// Before this height, `AssetLaunch` / `AssetManifestUpdate` accept zeroed hash/CID
+/// commits so historical replay stays deterministic. After activation, only the
+/// non-zero field check is consensus-enforced — local DHT pin cache is advisory.
 pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 80_000;
 
-/// When the pin gate is active but content is not in the validator's local DHT cache,
-/// reject the tx if true (strict); allow best-effort if false (design doc §5).
-#[cfg(test)]
-pub const MANIFEST_PIN_STRICT_MODE: bool = true;
-#[cfg(not(test))]
+/// Historical name for optional **admission** strictness (mempool / operator UX).
+/// Consensus apply never rejects on local pin presence; see `manifest_pin`.
 pub const MANIFEST_PIN_STRICT_MODE: bool = false;
+
+#[cfg(test)]
+pub mod manifest_pin_test_overrides {
+    use std::cell::Cell;
+
+    thread_local! {
+        static GATE_HEIGHT: Cell<Option<u64>> = const { Cell::new(None) };
+        static STRICT: Cell<Option<bool>> = const { Cell::new(None) };
+    }
+
+    pub fn set_gate_height(height: Option<u64>) {
+        GATE_HEIGHT.with(|c| c.set(height));
+    }
+
+    pub fn set_strict(strict: Option<bool>) {
+        STRICT.with(|c| c.set(strict));
+    }
+
+    pub fn gate_height() -> Option<u64> {
+        GATE_HEIGHT.with(|c| c.get())
+    }
+
+    pub fn strict() -> Option<bool> {
+        STRICT.with(|c| c.get())
+    }
+
+    /// RAII clear of overrides (use as `_guard` in tests).
+    pub struct Guard;
+    impl Guard {
+        pub fn new() -> Self {
+            set_gate_height(None);
+            set_strict(None);
+            Self
+        }
+    }
+    impl Drop for Guard {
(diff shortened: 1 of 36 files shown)


lib-blockchain/src/manifest_pin.rs
```diff
@@ -0,0 +1,213 @@
...
+    block_height: u64,
+) -> Result<(), TxApplyError> {
+    if !manifest_pin_gate_active(block_height) {
+        return Ok(());
+    }
+    if *manifest_cid == [0u8; 32] || *manifest_hash == [0u8; 32] {
+        return Err(TxApplyError::InvalidType("manifest fields required".into()));
+    }
+
+    match store.get_dht_pin_content(manifest_cid)? {
```
umwelt, line 197: **This is a consensus determinism violation.** `verify_manifest_pin` is called from `apply_asset_launch` and `apply_asset_manifest_update` — the deterministic block-apply path, run identically on every validator — and it branches on `store.get_dht_pin_content(manifest_cid)`, which reads the **local `dht_pins` sled tree**. That tree is populated by the DHT→web4 bridge (manifest_pin_bridge.rs) from content uploads and DHT gossip — *not* from block data. So its contents are node-local and non-deterministic across validators.

Once `manifest_pin_gate_active` is true (height >= 80k), that makes block validity depend on what each node happens to have cached:
- **Strict mode:** validator A has the pin (`Some` -> hash ok -> accept), validator B doesn't (`None` -> `Err` -> block apply fails). Same block, different result -> fork/halt. This is the exact class that halted the chain at 39,163.
- **Best-effort mode (prod default):** still non-deterministic. A validator that has content pinned under this cid whose bytes *mismatch* `manifest_hash` (a corrupt/partial pin, or arbitrary bytes uploaded under a colliding 16-byte cid via the web4 API) takes the `Some(bytes)` -> `validate_manifest_json` -> `Err` path and rejects, while a validator with no pin accepts. The *presence* of local content flips accept to reject.

Content availability cannot be a consensus apply-time check against local cache. Two deterministic options: (a) make the pin check **advisory / mempool-admission only** (reject at submit time on the node that received the tx, never at apply) so it never gates block validity; or (b) if it must be consensus-enforced, the manifest bytes (or a deterministic data-availability proof committed on-chain) have to travel *in the block*, so every validator verifies the same input. As written, flipping the gate on at 80k is a scheduled fork.

lib-blockchain/src/manifest_pin.rs
```diff
@@ -0,0 +1,213 @@
...
+        Some(bytes) => validate_manifest_json(&bytes, manifest_hash, ctx),
+        None => {
```
umwelt, line 199: The `None` branch is where the non-determinism becomes a hard reject: in strict mode an absent local pin fails the tx. Even setting aside strict/best-effort, note the asymmetry — `None` and `Some(valid)` both accept, but `Some(invalid)` rejects. So the reject condition is 'this node happens to hold mismatching bytes,' which is precisely the state that differs between nodes. There is no configuration of strict/best-effort that makes this deterministic while reading local cache.

lib-blockchain/src/contracts/sovereign_asset/state.rs
```diff
@@ -175,11 +175,40 @@ pub fn token_creation_allowed_in_block_at_height_with_sunset(
...
+/// without a local pin lookup so replay stays deterministic.
+#[cfg(test)]
+pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 0;
+#[cfg(not(test))]
+pub const MANIFEST_PIN_GATE_ACTIVATION_HEIGHT: u64 = 80_000;
+
+/// When the pin gate is active but content is not in the validator's local DHT cache,
+/// reject the tx if true (strict); allow best-effort if false (design doc §5).
+#[cfg(test)]
+pub const MANIFEST_PIN_STRICT_MODE: bool = true;
```
umwelt: Same test-masking pattern flagged on the sunset (#2885) and governance (#2887) PRs: `MANIFEST_PIN_STRICT_MODE` is `true` under `cfg(test)` and `false` in prod, and the gate height is `0` in test / `80_000` in prod. So every test runs gate-active + strict, and `unpinned_manifest_rejected_in_strict_test_mode` asserts the reject path — but production runs gate-off, then best-effort. The suite validates a consensus rule the chain never executes, and the actual prod branch (`Some(bytes)` mismatch -> reject while a peer accepts) is untested. Make the height and strict flag injectable (as was done for the token-creation sunset) so a test can drive the real prod semantics and assert cross-validator determinism.

umwelt (commented):
SA-7 review, scoped to the two SA-7 commits (28bdfe54, a570008d) — the stack also carries the still-unmerged SA-6 (#2887), reviewed separately.

The shareability goal is reasonable: don't let an asset launch commit to a manifest nobody can retrieve. But the enforcement mechanism reads **node-local, out-of-band state as a consensus input at apply time**, which is a fork/halt vector once the gate activates. This is the one blocking issue; details inline. The `CONSENSUS_BUILD_ID` → 5 bump and coordinated-deploy note are correct.

The reason it's dormant today (gate off below 80k, strict off in prod) is exactly why it's dangerous: the tests run with the gate on and strict on, so the suite validates a rule set production never runs, and the fork branch is invisible to it.

lib-blockchain/src/execution/sovereign_asset.rs
```diff
@@ -644,26 +731,94 @@ fn verify_authority_proof(
...
+                    }
+                    Ok(())
+                }
+                GovernanceVerifierState::Multisig { .. } => verify_governance_multisig_proof(
+                    mutator,
+                    verifier,
+                    gov_proof,
+                    &message_hash,
+                ),
             }
```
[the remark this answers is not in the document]
umwelt: Fixed: `GovernanceVerifierState::Single` now **always** calls `verify_governance_multisig_proof`. Non-`Multisig` / empty proofs no longer bypass Dilithium verification.

zhtp/src/runtime/manifest_pin_bridge.rs
```diff
@@ -0,0 +1,49 @@
...
+        if record_dht_pin_at_cid(store.as_ref(), &cid, content).is_ok() {
+            debug!(
+                manifest_cid = %hex::encode(&cid[..8]),
+                bytes = content.len(),
+                "mirrored explicit CID pin into consensus dht_pins cache"
+            );
+        }
+    } else if record_dht_pin_content(store.as_ref(), content).is_ok() {
+        debug!(bytes = content.len(), "mirrored content-derived CID pin into consensus cache");
+    }
```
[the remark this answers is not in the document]
umwelt: Fixed: the bridge always records under the **consensus truncated CID** (first 16 bytes of BLAKE3 via `record_dht_pin_content`). Full Web4 content-hash hex is no longer used as the pin key, so lookups match `AssetLaunch.manifest_cid`.

zhtp/src/api/handlers/web4/mod.rs
```diff
@@ -496,12 +496,19 @@ impl Web4Handler {
         );
 
         // Store blob in content-addressed storage
+        let content_bytes = request.body.clone();
         let content_id = self
             .domain_registry
-            .store_content_by_cid(request.body.clone())
+            .store_content_by_cid(content_bytes.clone())
             .await
             .map_err(|e| anyhow::anyhow!("Failed to store blob: {}", e))?;
```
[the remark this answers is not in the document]
umwelt: Fixed: single `request.body.clone()` for storage; mirror uses `&request.body`.

zhtp/src/api/handlers/web4/mod.rs
```diff
@@ -562,12 +569,19 @@ impl Web4Handler {
         );
 
         // Store manifest in content-addressed storage
+        let manifest_bytes = request.body.clone();
         let manifest_cid = self
             .domain_registry
-            .store_content_by_cid(request.body.clone())
+            .store_content_by_cid(manifest_bytes.clone())
             .await
             .map_err(|e| anyhow::anyhow!("Failed to store manifest: {}", e))?;
```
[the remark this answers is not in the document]
umwelt: Fixed: same single-clone path for the manifest upload handler.

zhtp-cli/src/commands/token.rs
```diff
@@ -563,6 +564,34 @@ pub fn build_governance_launch_config(
...
+        })?;
+    if !response.status.is_success() {
+        return Err(CliError::ApiCallFailed {
+            endpoint: "/api/v1/web4/content/blob".to_string(),
+            status: response.status.code(),
+            reason: response.status_message.clone(),
+        });
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
umwelt: Fixed: `pin_manifest_for_launch` now POSTs to `/api/v1/web4/content/manifest` so the node validates manifest JSON.

lib-blockchain/src/manifest_pin.rs
umwelt: Agreed — this was a consensus-split vector. Fixed by making apply-time pin check **advisory only**:

- When the gate is active: require non-zero on-chain `manifest_cid` / `manifest_hash` (deterministic).
- Local `dht_pins` is log/warn only — never `Err` based on cache presence or content quality.
- Optional strictness lives in `check_manifest_pin_admission` (mempool/operator UX), **not** block apply.

lib-blockchain/src/manifest_pin.rs
umwelt: Fixed: the apply path no longer rejects on `Some(invalid)` either. Missing, valid, and invalid local pins all **accept** after the deterministic non-zero field check. Cross-validator determinism is covered by unit tests.

lib-blockchain/src/contracts/sovereign_asset/state.rs
umwelt: Fixed:

- Prod constants are always `activation = 80_000`, `strict = false` (no `cfg(test)` flip of the real rule set).
- Tests inject gate height via `manifest_pin_test_overrides` and assert **prod semantics** (including cross-validator accept with/without cache).

umwelt:
## Review responses (Copilot + umwelt)

Pushed **0bcbf394** (review fixes) + **fd4cc976** (merge `development`).

| Source | Finding | Fix |
|--------|---------|-----|
| **umwelt** | Local `dht_pins` as consensus input | Apply path advisory only; never reject on cache |
| **umwelt** | Some(invalid) vs None asymmetry | Both accept after non-zero field check |
| **umwelt** | cfg(test) masks prod rules | Prod constants always; injectable test overrides + determinism tests |
| **Copilot** | Single verifier skips proof verify | Always `verify_governance_multisig_proof` |
| **Copilot** | Full Web4 CID vs truncated consensus CID | Bridge keys by content-derived truncated CID |
| **Copilot** | Double body clone | Single clone on web4 handlers |
| **Copilot** | CLI blob endpoint | Uses `/api/v1/web4/content/manifest` |

Local: `manifest_pin::tests` (8), `governance_proof` (2) green.

PR #2893, opened 2026-07-17
fix(reward-claim): reject out-of-range day/week timestamps

### LINKED ISSUE #2873: as i64 timestamp truncation in reward_claim day/week bucketing
## Problem

`lib-blockchain/src/transaction/reward_claim.rs` casts `u64` timestamps with `as i64` in three places:

- `utc_date_from_ts` (:163)
- `utc_day_ordinal_from_ts` (:170)
- `iso_week_from_ts` (:178)

These correctly use `unwrap_or_else`/`unwrap_or`, so they **cannot panic**. But they silently wrap: `u64::MAX as i64` is `-1`, which yields `1969-12-31` / ordinal `0` / `1970-W01` rather than being rejected.

## Why it matters

These functions bucket reward claims by day and week. A wrapped timestamp lands a claim in a fixed stale bucket instead of failing. `RewardClaim` is the transaction type that halted consensus twice this week (heights 34699 and 39162), so I'd rather this path be strict than lenient.

## Ask

Use `i64::try_from` and reject out-of-range values explicitly, the same discipline applied in #2870.

Related: #2870

umwelt:
## Summary

Closes #2873.

Reward claim day/week bucketing cast `u64` timestamps with `as i64`, so `u64::MAX` became `-1` and silently bucketed into 1969-12-31 / 1970-W01.

### Changes
- `i64::try_from` + `chrono::DateTime::from_timestamp` via shared helper
- Helpers return `Result<_, String>` and reject out-of-range values
- Apply path maps errors to `TxApplyError::InvalidType`
- Settlement `plan_reward_claim` propagates `?`
- Unit tests for valid buckets and `u64::MAX` rejection

## Test plan
- [x] `cargo test -p lib-blockchain --lib rejects_u64_max`
- [x] `cargo test -p lib-blockchain --lib valid_timestamp_buckets`

### DIFF
--- a/lib-blockchain/src/execution/reward_claim.rs
+++ b/lib-blockchain/src/execution/reward_claim.rs
@@ -125,9 +125,9 @@ pub fn apply_reward_claim(
         return reject_claim("signer is not authorized token creator");
     }
 
-    let date = utc_date_from_ts(block_timestamp);
-    let week = iso_week_from_ts(block_timestamp);
-    let today_ord = utc_day_ordinal_from_ts(block_timestamp);
+    let date = utc_date_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
+    let week = iso_week_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
+    let today_ord = utc_day_ordinal_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
     let token_id = &data.token_id;
 
     let mut streak_day = 1u32;
--- a/lib-blockchain/src/execution/reward_claim.rs
+++ b/lib-blockchain/src/execution/reward_claim.rs
@@ -125,6 +125,12 @@ pub fn apply_reward_claim(
         return reject_claim("signer is not authorized token creator");
     }
 
+    // Day/week bucketing rejects chrono-unrepresentable timestamps. On the apply
+    // path this is defence-in-depth only: `block_validate::validate_timestamp`
+    // already bounds `block_timestamp` to `now + max_future_timestamp` (default
+    // 2h) and enforces monotonicity, so a validated block cannot reach the
+    // ~8.2e12 chrono ceiling. Safety is therefore coupled to that config staying
+    // far below chrono's range — not to these helpers being infallible.
     let date = utc_date_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
     let week = iso_week_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
     let today_ord = utc_day_ordinal_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
--- a/zhtp/src/api/handlers/rewards/settlement.rs
+++ b/zhtp/src/api/handlers/rewards/settlement.rs
@@ -106,9 +106,9 @@ pub fn plan_reward_claim(
         Err(_) => return Err("owner_did_not_registered".into()),
     }
     let ts = claim_unix_ts();
-    let date = utc_date_from_ts(ts);
-    let week = iso_week_from_ts(ts);
-    let today_ord = utc_day_ordinal_from_ts(ts);
+    let date = utc_date_from_ts(ts)?;
+    let week = iso_week_from_ts(ts)?;
+    let today_ord = utc_day_ordinal_from_ts(ts)?;
 
     match spec.event {
         RewardEventKind::Welcome => {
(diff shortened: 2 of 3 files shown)


lib-blockchain/src/execution/reward_claim.rs
```diff
@@ -125,9 +125,9 @@ pub fn apply_reward_claim(
         return reject_claim("signer is not authorized token creator");
     }
 
-    let date = utc_date_from_ts(block_timestamp);
-    let week = iso_week_from_ts(block_timestamp);
-    let today_ord = utc_day_ordinal_from_ts(block_timestamp);
+    let date = utc_date_from_ts(block_timestamp).map_err(TxApplyError::InvalidType)?;
```
umwelt, line 134: These three now return `Err` -> `TxApplyError::InvalidType`, which fails block apply. That's the halt pattern we've hit repeatedly, so worth a comment stating why it's safe here: `block_timestamp` is bounded before it reaches this code — `block_validate.rs:161` (`validate_timestamp`) caps the upper end at `now + max_future_timestamp` and enforces monotonicity, so it's always well inside chrono's ~8.2e12 range and this `Err` can never fire on a validated block.

Two asks: (1) add that one-line comment so the next reader doesn't have to re-derive it; (2) note that the safety is now coupled to `max_future_timestamp` — if that config were ever set absurdly large (> ~8.2e12 - now), a block could pass validation and then halt here. Realistically minutes, so not blocking, but the coupling is real.

umwelt (commented):
Closes #2873. Correctly removes the silent `as i64` wrap that mapped `u64::MAX` -> -1 -> 1969-12-31, replacing it with explicit `i64::try_from` + chrono-range checks. The `settlement.rs` path is the API planner (`ts = claim_unix_ts()`, always current time) so `?` just returns an API error there — fine.

The one thing worth care: this touches the consensus apply path (`apply_reward_claim`) and makes previously-infallible helpers fallible. I checked it — it's safe — but the safety argument should be written down (inline).

zhtp/src/api/handlers/rewards/settlement.rs
```diff
@@ -106,9 +106,9 @@ pub fn plan_reward_claim(
         Err(_) => return Err("owner_did_not_registered".into()),
     }
     let ts = claim_unix_ts();
-    let date = utc_date_from_ts(ts);
-    let week = iso_week_from_ts(ts);
-    let today_ord = utc_day_ordinal_from_ts(ts);
+    let date = utc_date_from_ts(ts)?;
+    let week = iso_week_from_ts(ts)?;
+    let today_ord = utc_day_ordinal_from_ts(ts)?;
```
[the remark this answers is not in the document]
umwelt: Fixed: `plan_reward_claim` maps conversion failures to the stable reason code `invalid_timestamp` (no raw `timestamp N exceeds…` strings to clients / soft_fail matching).

lib-blockchain/src/execution/reward_claim.rs
umwelt: Added an inline comment on the apply path covering both points:

1. Safety: validated blocks already pass `block_validate::validate_timestamp` (`now + max_future_timestamp`, default 2h) so these `Err`s cannot fire on normal consensus input.
2. Coupling: if `max_future_timestamp` were ever set near chrono’s ~8.2e12 ceiling, apply could still reject — noted explicitly as a config coupling, not blocking for realistic values.

umwelt:
## Review responses

| Finding | Fix |
|---------|-----|
| Copilot: leak detailed timestamp errors on API path | Map to stable `invalid_timestamp` |
| umwelt: document apply-path safety + max_future_timestamp coupling | Inline comment on `apply_reward_claim` |

zhtp/src/api/handlers/rewards/settlement.rs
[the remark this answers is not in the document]
umwelt: Fixed: `plan_reward_claim` maps conversion failures to the stable reason code `invalid_timestamp` (no raw `timestamp N exceeds…` strings to clients / soft_fail matching).

lib-blockchain/src/execution/reward_claim.rs
umwelt: Added an inline comment on the apply path covering both points:

1. Safety: validated blocks already pass `block_validate::validate_timestamp` (`now + max_future_timestamp`, default 2h) so these `Err`s cannot fire on normal consensus input.
2. Coupling: if `max_future_timestamp` were ever set near chrono’s ~8.2e12 ceiling, apply could still reject — noted explicitly as a config coupling, not blocking for realistic values.

umwelt:
## Review responses

| Finding | Fix |
|---------|-----|
| Copilot: leak detailed timestamp errors on API path | Map to stable `invalid_timestamp` |
| umwelt: document apply-path safety + max_future_timestamp coupling | Inline comment on `apply_reward_claim` |

PR #2894, opened 2026-07-17
fix(economy): surface excluded_invalid_timestamps in analytics

### LINKED ISSUE #2874: Analytics output is silently self-inconsistent when timestamps are skipped
## Problem

After #2870, transactions with invalid timestamps are counted in `transaction_counts` / `volume_by_category` / `failure_analysis` but excluded from `monthly_trends` and `daily_patterns`.

That's the right behaviour, but it means `sum(monthly_trends[*].total_transactions) != sum(transaction_counts[*])`, and a consumer of `TransactionAnalytics` has no way to know why. The only signal is a server-side log line.

## Ask

Surface the gap in the struct — e.g. `excluded_invalid_timestamps: u64` — so callers can see records were dropped from the time series instead of silently getting numbers that don't reconcile.

Related: #2870

umwelt:
## Summary

Closes #2874.

After invalid timestamps are skipped from time-series analytics, category totals no longer match `sum(monthly_trends)`. Callers had only a server log to explain the gap.

### Changes
- Parse timestamps with `i64::try_from` + `chrono::DateTime::from_timestamp` (no panicking unwrap)
- Count category / volume / fee / failure stats for all records
- Skip invalid timestamps only for `monthly_trends` and `daily_patterns`
- Add `excluded_invalid_timestamps: u64` (`#[serde(default)]`) so
  `sum(monthly[*].total_transactions) + excluded_invalid_timestamps == sum(transaction_counts[*])`
- Regression test covering the reconcile identity

## Test plan
- [x] `cargo test -p lib-economy --lib excluded_invalid`
- [x] `cargo test -p lib-economy --lib test_analytics` (if present)

### DIFF
--- a/lib-economy/src/wallets/transaction_history.rs
+++ b/lib-economy/src/wallets/transaction_history.rs
@@ -16,6 +16,15 @@ use crate::wasm::logging::info;
 // integrations (avoiding blockchain circular dependency)
 use crate::network_types::{get_mesh_status, get_network_statistics};
 
+
+/// Parse a history unix timestamp for chrono without panicking or wrapping.
+///
+/// Returns `None` when the value does not fit `i64` or is outside chrono's range.
+fn parse_history_datetime(ts: u64) -> Option<chrono::DateTime<chrono::Utc>> {
+    let ts_i = i64::try_from(ts).ok()?;
+    chrono::DateTime::from_timestamp(ts_i, 0)
+}
+
 // Local transaction type to avoid blockchain dependency
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BlockchainTransaction {
@@ -158,6 +167,13 @@ pub struct TransactionAnalytics {
     pub daily_patterns: HashMap<u8, DailyActivityData>, // Hour of day -> activity
     /// Performance metrics
     pub performance_metrics: TransactionPerformanceMetrics,
+    /// Records counted in category/volume/failure stats but skipped for
+    /// `monthly_trends` / `daily_patterns` because the timestamp is not
+    /// chrono-representable. Lets callers reconcile
+    /// `sum(monthly_trends[*].total_transactions) + excluded_invalid_timestamps
+    /// == sum(transaction_counts[*])` (#2874).
+    #[serde(default)]
+    pub excluded_invalid_timestamps: u64,
 }
 
 /// Monthly transaction data for trend analysis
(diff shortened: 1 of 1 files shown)


lib-economy/src/wallets/transaction_history.rs
```diff
@@ -737,25 +753,35 @@ impl TransactionHistoryManager {
...
                 .or_insert(0u128) += tx.amount;
             *fees_by_category.entry(tx.category.clone()).or_insert(0u128) += tx.fees;
 
+            // Failure analysis (timestamp-independent)
+            if let TransactionStatus::Failed { reason } = &tx.status {
+                *failure_analysis.entry(reason.clone()).or_insert(0) += 1;
+            }
+
+            // Time-series only when timestamp is chrono-representable (#2870 / #2874)
+            let Some(dt) = parse_history_datetime(tx.timestamp) else {
```
umwelt, line 793: Nice — moving failure-analysis above this `continue` also fixes the bug from the original #2870 review where a failed tx with a skipped timestamp dropped out of `failure_analysis`. The `saturating_add` on the counter and the reconciliation assertion in the test are exactly right.

umwelt (commented):
Closes #2874, and does more than the issue asked — the reconciliation invariant (`monthly_total + excluded_invalid_timestamps == sum(transaction_counts)`) is asserted directly in the new test. `#[serde(default)]` is fine here since this is a JSON response struct, not bincode-persisted. No blocking concerns.

Two cross-PR notes:
1. **This supersedes #2870**, which is still open. #2870 was the original analytics fix (its `u64::MAX` test didn't reproduce the panic); this rewrites the same `compute_analytics` correctly off current `development`. #2870 should be closed in favor of this, or you'll have two competing diffs on the same function.
2. **Textual conflict with #2892** — both insert a helper right after the `network_types` import. Whichever merges second needs a trivial rebase.

lib-economy/src/wallets/transaction_history.rs
umwelt: Thanks — yes, counting failures before the time-series `continue` is intentional so invalid-timestamp failed txs stay in `failure_analysis` (the #2870 gap).

umwelt:
## Review responses

No blocking code changes requested. Acknowledged:

- **umwelt:** failure_analysis before `continue` + reconciliation invariant in the test — intentional.
- **Cross-PR:** this supersedes open [#2870](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2870) on `compute_analytics`. Prefer merging this and closing #2870 (or rebasing #2870 onto this) to avoid competing diffs.
- **#2892:** trivial textual conflict on the helper insertion point; merge either first then rebase the other.

Issue #2900, opened 2026-07-17
[BUBL][MSG-R4] Honest send status (queued vs pushed)


umwelt:
## Parent
Epic #2896

## Problem
`msg/send` always returns `status: \"delivered\"` even when only deposited or only mpsc-pushed.

## Work
- Return `queued` | `pushed` (+ later `acked`) and `message_id`.
- Update lib-client / docs.

## Acceptance
- Client can distinguish offline queue vs live push.

Issue #2899, opened 2026-07-17
[BUBL][MSG-R3] Inbound DID resolution parity with msg/receive


umwelt:
## Parent
Epic #2896

## Problem
`msg/inbound` does not use OPAQUE session-bind path used by `msg/receive`. Mobile ephemeral QUIC keys can open stream under wrong DID → empty inbox while deposits sit under canonical DID.

## Work
- Extract shared DID resolver; use in both receive and inbound.

## Acceptance
- Same OPAQUE mobile session resolves identical DID on receive and inbound.

Issue #2898, opened 2026-07-17
[BUBL][MSG-R2] Live push must not lose without durable backup


umwelt:
## Parent
Epic #2896

## Problem
`try_push` success returns early without depositing. If inbound stream dies with messages still in the mpsc (depth 32), they are lost forever.

## Work
- Persist before/alongside push; requeue to deposit if unacked.
- Enforce full-queue → deposit fallback.

## Acceptance
- Kill stream after push, before client read → message still collectable via receive/deposit.

umwelt:
Delivered via #2909 (restack after #2904 merge).

Issue #2896, opened 2026-07-17
[BUBL] Epic — Messaging node reliability & delivery remediation


umwelt:
## Summary

**BUBL product epic** for fixing node-side reliability of post-quantum E2E messaging (`/api/v1/msg/*`, deposits, mesh `MessageRelay`, inbound push).

Parent feature epic: **#2459** (PQ P2P messaging).  
Related: **#2541** (push / inbound stream), **#2489** (mobile integration guide).

This epic is **not** about new chat UX features. It is about making the **node** behave as a correct store-and-forward relay: no silent loss, honest delivery status, cluster-safe cancellation, and auth that matches BUBL threat model.

---

## Problem statement (current node behavior)

Audit of `zhtp/src/messaging/*` + `messaging_provider` + mesh relay (2026-07-17):

| Area | Today | Risk |
|------|--------|------|
| Deposit removal | `collect_for_recipient` **deletes before** client confirms receipt | Message loss on disconnect mid-response |
| Live push | `try_push` to mpsc only; **no deposit** on success | Stream drop loses message forever |
| API status | Always `"delivered"` | Client/ops cannot distinguish queued vs pushed vs read |
| Mesh fan-out | Copy to all peers; **no cross-node delete** | Stale copies + possible multi-node double-poll |
| Storage | **In-memory only** | Restart drops all undelivered mail |
| Encryption at rest | E2E ciphertext stored as opaque blob; **metadata plaintext**; no node-side re-encrypt | Traffic analysis + harvest ciphertext if node compromised |
| Auth | Deposit/send do not bind `sender_did` to session; no envelope sig check on node | Spam / spoofed deposits |
| Inbound DID resolve | Weaker than `msg/receive` (no OPAQUE session bind) | Empty inbox while mail sits under canonical DID |
| `cancel()` | Implemented, **never called** | Dead API |

**What *does* work:** body is E2E encrypted (node cannot read plaintext); deposit path removes mail **on collect** for that node; 48h TTL + periodic cleanup.

---

## Goals

1. **At-least-once until client ack** (or explicit expiry) — no delete-before-confirm.
2. **Honest delivery states** (`queued` / `pushed` / `acked` / `expired`).
3. **Durable deposits** across process restart (sled or equivalent).
4. **Cluster semantics** for multi-validator mesh (message id + dedupe + optional tombstone).
5. **AuthZ** so only the real sender can deposit as that DID; optional node-side signature verify.
[...]

PR #2904, opened 2026-07-17
fix(msg): MSG-R1 deposit peek/ack (no delete-on-poll)

### LINKED ISSUE #2897: [BUBL][MSG-R1] Deposit drain: no delete-before-confirm
## Parent
Epic #2896

## Problem
`collect_for_recipient` removes envelopes from the node store **before** the client has confirmed receipt. Mid-stream disconnect or HTTP drop → permanent loss.

## Work
- Peek or lease deposits; delete only on ack (or re-deposit on write failure as interim).
- Fix `handle_receive` + inbound initial drain.

## Acceptance
- Kill client mid-drain → remaining messages still available on reconnect.

umwelt:
## Summary
- Deposit store peeks only; envelopes stay until `POST /api/v1/msg/ack`
- Stable `message_id = hex(blake3(envelope))` with idempotent deposits
- `/msg/receive` and `/msg/inbound` no longer delete on read

## Stack
1. **This PR** — MSG-R1 (#2897)
2. MSG-R2 live push durability (#2898)
3. MSG-R3 shared DID resolver (#2899)
4. MSG-R4 honest send status (#2900)

Epic: #2896

## Test plan
- [ ] `cargo test -p zhtp --lib messaging::deposit`
- [ ] Poll `/msg/receive` twice → same messages until ack
- [ ] `POST /msg/ack` with `message_ids` → inbox empty

Closes #2897

### DIFF
--- a/zhtp/src/messaging/handler.rs
+++ b/zhtp/src/messaging/handler.rs
@@ -51,6 +51,12 @@ struct PresenceWatchRequest {
     target_dids: Vec<String>,
 }
 
+#[derive(Deserialize)]
+struct AckRequest {
+    /// Message ids returned by receive / computed as hex(blake3(envelope)).
+    message_ids: Vec<String>,
+}
+
 // ── Handler ────────────────────────────────────────────────────────
 
 fn json_response(data: serde_json::Value) -> Result<ZhtpResponse> {
@@ -168,10 +174,14 @@ impl MessagingHandler {
             }));
         }
 
-        // Always deposit locally — recipient may poll any node, or come online later
-        let _ = self.deposits
-            .deposit(&envelope.sender_did, &recipient_did, vec![envelope_bytes.clone()])
-            .await;
+        // Deposit when no live subscriber (MSG-R2 will deposit before push too).
+        if let Err(e) = self.deposits.deposit_one(
+            &envelope.sender_did,
+            &recipient_did,
+            envelope_bytes.clone(),
+        ) {
+            return Ok(error_resp(ZhtpStatus::BadRequest, &e));
+        }
 
         // Also relay via mesh so all nodes have the message
         if let Ok(mesh_router) = crate::runtime::mesh_router_provider::get_global_mesh_router().await {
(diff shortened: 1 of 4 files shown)


zhtp/src/messaging/handler.rs
```diff
@@ -229,19 +239,43 @@ impl MessagingHandler {
...
+        }
+
+        if request.requester.is_none() {
+            return Ok(error_resp(
+                ZhtpStatus::Unauthorized,
+                "Authenticated DID required",
+            ));
+        }
+
+        let removed = self.deposits.ack(&req.message_ids);
```
umwelt: This acks by raw `message_id` with no check that the authenticated caller is the recipient of those messages. `ack()` in deposit.rs does `by_id.remove(mid)` and deletes regardless of who's asking. Combined with `message_id = blake3(envelope)` (deterministic, disclosed on receive/send), any authenticated user can delete another user's pending messages. Resolve the caller's canonical DID (as `handle_receive` does, and as #2906 factors out) and pass it into `ack` so it only removes entries where `recipient == caller_did`.

umwelt (commented):
MSG-R1 (peek/ack instead of delete-on-poll) — the design is right and the bounds are handled (48h TTL, `MAX_PENDING_PER_PAIR = 100`, GC, idempotent deposits via stable `message_id = blake3(envelope)`). But there's a **critical authorization gap in ack** that defeats the durability this PR exists to provide, plus two secondary issues.

**CRITICAL — `handle_ack` doesn't scope deletion to the caller.** It checks only that *some* DID is authenticated (`requester.is_none()`), then passes `message_ids` straight to `deposits.ack()`, which removes purely by id with no recipient check. `message_id` is `hex(blake3(envelope_bytes))` — deterministic, and handed out on `/msg/receive` (and, once #2907 lands, in the `/msg/send` response to the *sender*). So any authenticated party who has seen an envelope can compute its id and delete it from the recipient's inbox before the recipient ever polls → silent denial of delivery. That is exactly the loss peek/ack is meant to prevent.

The fix is already being built in #2906: resolve the caller's canonical DID (the same `did_resolve` path `handle_receive` uses) and have `ack` only remove entries whose `recipient == caller_did`. Right now `handle_receive` does that careful resolution and `handle_ack` does none of it.

**Secondary:**
- **Sender is spoofable.** `handle_send` reads `envelope.sender_did` from the deserialized envelope and never checks it against the authenticated requester. Since the cap is per-`(sender, recipient)` pair, an attacker cycles fake sender DIDs to store 100×N envelopes against one victim (bounded only by the 48h TTL). Bind `sender_did` to the authenticated identity, or cap per-recipient rather than per-pair.
- **In-memory only** (`RwLock<HashMap>`, "Phase 3 will add sled"). Acknowledged in the code, but worth stating in the PR: a node restart drops all un-acked deposits, so "durable" today means durable-within-process-lifetime.
- Minor: `peek_for_recipient` and `has_pending` scan every `(sender, recipient)` pair filtering by recipient — O(total pairs) per poll.

PR #2905, opened 2026-07-17
fix(msg): MSG-R2 always deposit before live push

### LINKED ISSUE #2898: [BUBL][MSG-R2] Live push must not lose without durable backup

umwelt:
## Summary
- `msg/send` and mesh relay **deposit first**, then best-effort `try_push`
- Live stream drop can no longer lose mail that never hit the store

## Stack
Depends on MSG-R1 (#2897). Part of epic #2896.

## Test plan
- [ ] Send while recipient has open `/msg/inbound` → message still in deposit until ack
- [ ] Kill inbound mid-stream → reconnect/poll still sees message

Closes #2898

### DIFF
--- a/zhtp/src/messaging/handler.rs
+++ b/zhtp/src/messaging/handler.rs
@@ -140,7 +140,7 @@ impl MessagingHandler {
     }
 
     /// POST /api/v1/msg/send
-    /// Relay an encrypted envelope to the recipient via QUIC mesh.
+    /// Deposit first (durable), then best-effort live push + mesh relay.
     async fn handle_send(&self, request: &ZhtpRequest) -> Result<ZhtpResponse> {
         let req: SendRequest = serde_json::from_slice(&request.body)
             .map_err(|e| anyhow::anyhow!("Invalid request: {}", e))?;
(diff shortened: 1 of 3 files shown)


umwelt (commented):
MSG-R2 — correct and it closes a real message-loss window. The old path pushed to a live subscriber and returned `delivered` **without depositing**, so a client that received the push but crashed before processing it had no deposit to re-poll — the message was gone. Always depositing first, with live push demoted to best-effort, fixes that, and the `try_push` doc update ('a successful push is not delivery confirmation — the deposit remains until client ack') states the invariant correctly.

One carry-over: this is sound on its own, but it inherits the ack-authorization gap from #2904 — always-deposit is only durable if ack can't be forged by a third party. Worth landing the ack-scoping fix (see my #2904 review) before this chain is relied on for at-least-once delivery.

PR #2906, opened 2026-07-17
fix(msg): MSG-R3 shared DID resolver for receive and inbound

### LINKED ISSUE #2899: [BUBL][MSG-R3] Inbound DID resolution parity with msg/receive

umwelt:
## Summary
- New `messaging/did_resolve` used by both `/msg/receive` and `/msg/inbound`
- Session bind → sled identity → stub DID (same inbox on poll and stream)

## Stack
Depends on MSG-R2. Part of epic #2896.

## Test plan
- [ ] OPAQUE/mobile session: receive and inbound resolve same canonical DID
- [ ] After restart, sled-backed identity still resolves

Closes #2899

### DIFF
--- a/zhtp/src/messaging/did_resolve.rs
+++ b/zhtp/src/messaging/did_resolve.rs
@@ -1,0 +1,84 @@
+//! Shared recipient DID resolution for messaging endpoints.
+//!
+//! MSG-R3: `/msg/receive` and `/msg/inbound` must use the same resolver so a
+//! client does not see an empty inbox on one path and mail on the other.
+//!
+//! Resolution order (first hit wins):
+//! 1. Session-bound device map (OPAQUE login: quic_key → canonical DID)
+//! 2. Blockchain durable identity lookup by device key id
+//! 3. Fallback stub DID `did:zhtp:{key_id}` (empty inbox, not an error)
+
+use lib_blockchain::Blockchain;
+use lib_protocols::types::ZhtpRequest;
+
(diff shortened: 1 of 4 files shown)


umwelt (commented):
MSG-R3 — good refactor. Extracting `resolve_recipient_did` into a shared `did_resolve` module and using it from both `handle_receive` and the inbound stream removes a real duplication hazard: the two copies could (and the codebase shows a history of them) drift, so a device would resolve to different canonical DIDs on receive vs inbound and see an inconsistent inbox. One resolver is the right call.

Concrete follow-on: this is exactly the tool the #2904 ack bug needs. `handle_ack` currently does no DID resolution at all — wire `resolve_recipient_did_from_request` into it and scope `ack` to the resolved caller DID. Doing that here (or noting it as the immediate next step) would close the loop, since #2906 is what makes the fix a one-liner at the call site.

umwelt:
Review follow-up: ack is already scoped to the authenticated caller DID (R1 #2904), and this PR routes that path through the shared `did_resolve` module via `resolve_caller_did` → `resolve_recipient_did_from_request`.

PR #2907, opened 2026-07-17
fix(msg): MSG-R4 honest send status (queued|pushed)

### LINKED ISSUE #2900: [BUBL][MSG-R4] Honest send status (queued vs pushed)

umwelt:
## Summary
- Send returns `queued` or `pushed` (never dishonest `delivered`)
- Response includes `message_id` for client tracking/ack

## Stack
Depends on MSG-R3. Tip of Phase 1 stack for epic #2896.

## Test plan
- [ ] Send offline recipient → `status: queued` + `message_id`
- [ ] Send with live inbound → `status: pushed` + `message_id`

Closes #2900

### DIFF
--- a/zhtp/src/messaging/handler.rs
+++ b/zhtp/src/messaging/handler.rs
@@ -140,7 +140,8 @@ impl MessagingHandler {
     }
 
     /// POST /api/v1/msg/send
-    /// Deposit first (durable), then best-effort live push + mesh relay.
+    /// Deposit (always) then optionally push to a live inbound subscriber.
+    /// Status is honest: `queued` (stored only) or `pushed` (stored + live).
     async fn handle_send(&self, request: &ZhtpRequest) -> Result<ZhtpResponse> {
         let req: SendRequest = serde_json::from_slice(&request.body)
             .map_err(|e| anyhow::anyhow!("Invalid request: {}", e))?;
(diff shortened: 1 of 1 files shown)


umwelt (commented):
MSG-R4 — correct. `delivered` was a lie (a live push is not delivery confirmation), so `queued` (stored only) vs `pushed` (stored + live) is the honest status, and returning `message_id` lets the client ack. Clean.

One interaction to be aware of, and it sharpens the #2904 finding: returning `message_id` in the `/msg/send` response hands the **sender** the exact token needed to ack — and because #2904's `ack` isn't scoped to the recipient, the sender can then delete their own message from the recipient's inbox before it's read. The status change is right; it just makes fixing the ack-authorization gap (scope `ack` to the resolved caller DID) a prerequisite rather than a nice-to-have. The `message_id` return itself is fine and useful once ack is scoped.

umwelt:
Ack authorization gap noted here was closed in #2904 (recipient-scoped ack). Returning `message_id` on send is safe with that gate.

PR #2908, opened 2026-07-17
fix(reward-claim): UnregisteredSender blocks spend-delegate claims

umwelt:
## Summary
Production rewards claims fail with:
`reward claim submit failed: Failed to submit reward claim: Transaction verification failed`

**Root cause (g1 logs):** `Transaction validation failed: UnregisteredSender`

`RewardClaim` is signed by the BUBL **spend-delegate keystore** (treasury), which is not a registered user identity/wallet. Stateful validation still ran `validate_sender_identity_exists` against in-memory wallet/identity maps (28 wallets on g1; no match). After a store-backed restart those maps no longer mirror sled, so every claim is rejected before the owner_did sled gate.

`TokenTransfer` already skips this check; `RewardClaim` did not.

## Fix
- Exempt `TransactionType::RewardClaim` from `validate_sender_identity_exists` (same class as TokenTransfer). Authorization remains Dilithium sig + `data.from` key_id + on-chain spend delegate at apply; beneficiary `owner_did` still sled-gated.
- Surface `ValidationError` in mempool reject strings (no more opaque client message).
- Regression test: claim accepted when treasury identity is absent and beneficiary is registered.

## Test plan
- [x] `cargo test -p lib-blockchain --test reward_claim_mempool_tests`
- [ ] Deploy and confirm `/api/v1/rewards/claim` no longer returns verification failed
- [ ] g1 journal: no new `UnregisteredSender` on RewardClaim

## Evidence
```
WARN Transaction validation failed: UnregisteredSender
Transaction details: ... type=RewardClaim ...
SECURITY CRITICAL: Transaction from unregistered wallet/identity!
Total wallets to check: 28
```

Not caused by timestamp PRs (#2892/#2893); those only affect day/week bucketing on plan/apply.

### DIFF
--- a/lib-blockchain/src/transaction/validation.rs
+++ b/lib-blockchain/src/transaction/validation.rs
@@ -2549,6 +2549,9 @@ impl<'a> StatefulTransactionValidator<'a> {
 
         // RewardClaim: signer is the spend delegate (system tx), but beneficiary
         // `owner_did` must exist in durable sled — same oracle as executor apply.
+        // Spend-delegate (or legacy token-creator) authorization must also match
+        // apply (`reward_claim.rs`) so a non-delegate claim is rejected at
+        // mempool admission instead of halting block apply (#2859 class).
         if transaction.transaction_type == TransactionType::RewardClaim {
             let data = transaction
                 .reward_claim_data()
@@ -2566,6 +2569,31 @@ impl<'a> StatefulTransactionValidator<'a> {
                 tracing::warn!(
                     "[REWARD_CLAIM] owner_did not registered in sled: {} ({e})",
                     &data.owner_did[..data.owner_did.len().min(32)]
+                );
+                return Err(ValidationError::InvalidTransaction);
+            }
+            if let Some(rewards_state) = blockchain.get_rewards_module_state(&data.token_id) {
+                if data.from != rewards_state.spend_delegate_key_id {
+                    tracing::warn!(
+                        "[REWARD_CLAIM] signer is not on-chain spend delegate: from={} expected={}",
+                        hex::encode(data.from),
+                        hex::encode(rewards_state.spend_delegate_key_id)
+                    );
+                    return Err(ValidationError::Unauthorized);
+                }
+            } else if let Some(contract) = blockchain.get_token_contract(&data.token_id) {
+                if contract.creator.key_id != data.from {
+                    tracing::warn!(
+                        "[REWARD_CLAIM] signer is not token creator (legacy path): from={} creator={}",
+                        hex::encode(data.from),
+                        hex::encode(contract.creator.key_id)
+                    );
+                    return Err(ValidationError::Unauthorized);
+                }
+            } else {
+                tracing::warn!(
+                    "[REWARD_CLAIM] token contract not found for claim token_id={}",
+                    hex::encode(data.token_id)
                 );
                 return Err(ValidationError::InvalidTransaction);
             }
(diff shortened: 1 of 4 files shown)


lib-blockchain/src/transaction/validation.rs
```diff
@@ -2491,6 +2491,14 @@ impl<'a> StatefulTransactionValidator<'a> {
...
             && transaction.transaction_type != TransactionType::IdentityRegistration
             && transaction.transaction_type != TransactionType::TokenTransfer
+            // RewardClaim is signed by the asset spend-delegate keystore (treasury),
+            // not a user identity. Authorization is Dilithium sig + on-chain
+            // RewardsModuleState.spend_delegate_key_id at apply; beneficiary
+            // `owner_did` is gated separately (sled) below. Requiring the
+            // delegate key in identity_registry/wallets rejects all claims
+            // after a store-backed restart (in-memory maps empty / delegate
+            // never registered as a user). Same exemption class as TokenTransfer.
+            && transaction.transaction_type != TransactionType::RewardClaim
```
umwelt, line 2501: Exempting RewardClaim here is correct — but it means mempool no longer applies *any* sender-side gate to RewardClaim, while apply still hard-rejects a non-delegate signer (`reward_claim.rs:118` -> `Err` -> `?` at `executor.rs:3544` -> block halt). So this line, as written, converts 'non-delegate claim' from 'rejected at mempool (because the key wasn't a registered identity)' into 'admitted at mempool, halts the chain at apply.' Add the spend-delegate check (`data.from == RewardsModuleState.spend_delegate_key_id`, with the legacy `contract.creator.key_id` fallback) to this stateful validation path so the mempool decision matches apply, the same way #2859 made the owner_did gate symmetric. Without that, removing `validate_sender_identity_exists` leaves apply as the only gate, and apply's only failure mode is a halt.

umwelt (commented):
The immediate fix is correct and the happy-path authorization is fully sound — I traced the whole chain and it holds. But removing the mempool sender check **widens a pre-existing halt vector**, and this PR is the right place to close it. Detail inline; summary here.

**Why the legit-claim authorization is safe (verified):** the envelope signature is verified (validation.rs:1257), `data.from == signature.public_key.key_id` is enforced (validation.rs:822, rejects on mismatch), and apply requires `data.from == rewards_state.spend_delegate_key_id` (reward_claim.rs:110). So a valid claim must be signed by the on-chain spend delegate; `validate_sender_identity_exists` (which only checks whether the key is a *registered user*) was both wrong for a treasury keystore and redundant. Exempting RewardClaim — same class as TokenTransfer — is the right call, and the #2859 owner_did mempool gate is preserved. Good regression test on the positive case.

**The problem: the spend-delegate check exists only at apply, and apply-failure halts.** `apply_reward_claim` returns `Err` for a non-delegate signer (reward_claim.rs:118), which propagates via `?` at executor.rs:3544 → block apply fails → consensus halt. That's the exact #2859 mechanism (mempool admits what apply rejects). Before this PR, `validate_sender_identity_exists` incidentally required the signer to be a registered identity, which raised the bar. After it, **any** key can craft a RewardClaim with `from == signer` and a registered `owner_did`, pass mempool, land in a block, and halt every validator at apply. The vector pre-exists for registered users; this PR removes the registration barrier and makes it trivial.

**Fix:** mirror what #2859 did for `owner_did` — add the spend-delegate check to the mempool / pre-vote path so a non-delegate claim is rejected at admission and never reaches a block. (Alternatively, make a non-delegate claim a graceful per-tx drop at apply instead of a fatal `Err`, but the mempool-gate approach is what #2859 established and is the safer of the two.)

**Test gap:** the PR proves the delegate claim is accepted; it needs the negative — a claim where `from == signer` but `from != spend_delegate_key_id` must be rejected **at mempool**, not reach apply. That test would fail today and is exactly what demonstrates the missing gate.

The reject-string surfacing is a good incidental improvement.

umwelt:
Addressed review: mempool now mirrors apply for spend-delegate / legacy token-creator authorization (`Unauthorized` if mismatch). Added `non_delegate_signer_rejected_from_mempool`. Display formatting for validation errors was already fixed in d8e373e4.

PR #2912, opened 2026-07-17
fix(lib-client): MSG-NODE-001 session open errors + UHP handshake parity

### LINKED ISSUE #2911: fix(lib-client): make zhtp_quic_session_open UHP handshake reliable for mobile /msg/inbound
## CR: MSG-NODE-001

Make `zhtp_quic_session_open` (lib-client) reliable for mobile persistent session + `GET /msg/inbound` push.

### Problem
Mobile path B (`zhtp_quic_session_open`) often fails with opaque `openFailed` / length-prefix connection lost, forcing poll fallback.

### Resolution (this PR)
- Structured last-error + stage on session open/rpc/inbound
- Canonical initiator already used; public `quic_uhp_capabilities()` for golden parity
- Stage-tagged connect errors in ZhtpClient
- Live verified: open + health RPC against g1

### Mobile follow-up
Rebuild natives; call `zhtp_quic_session_last_error` / `_stage` on null open; prefer push + ack.

Related: MSG-CLIENT-001 (peek/ack — orthogonal)

umwelt:
## Summary
Implements **MSG-NODE-001** for mobile persistent QUIC session reliability.

- **Structured FFI errors**: `zhtp_quic_session_last_error`, `zhtp_quic_session_last_error_stage`, `zhtp_quic_session_last_error_free` — format `stage=<name>: <chain>` so mobile never only sees `openFailed`
- **Handshake parity**: public `quic_uhp_capabilities()` is the single capabilities constructor for initiator/responder/wire golden test (canonical path already used by `zhtp_quic_session_open` → `ZhtpClient`)
- **Connect stages**: resolve / quic / uhp_handshake / trust / identity / alpn tagged in error chains
- **Tests**: unit (alpn, identity, stage parse) + ignored live open; **live open + health RPC passed against g1**

Closes #2911

## Mobile integration

```text
handle = zhtp_quic_session_open(...)
if handle == null {
  err = zhtp_quic_session_last_error()   // free with zhtp_quic_session_last_error_free
  stage = zhtp_quic_session_last_error_stage()  // static C string, do not free
}
```

## Test plan
- [x] `cargo test -p lib-client --lib quic_session`
- [x] `cargo test -p lib-network --test handshake_wire_bytes_test`
- [x] Live: `ZHTP_SESSION_TEST_HOST=77.42.37.161 cargo test -p lib-client --lib live_session_open -- --ignored`
- [ ] Mobile rebuild natives + soak `/msg/inbound`

## Notes
Quinn-FFI one-shot is unchanged. If Quinn-FFI ClientHello diverges from `quic_uhp_capabilities`, mobile should align **to** the canonical initiator (or share this module), not the reverse.

### DIFF
--- a/lib-client/src/quic_session.rs
+++ b/lib-client/src/quic_session.rs
@@ -8,16 +8,23 @@
 //! RPC, and inbound work on the shared `CLIENT_ENDPOINT_RUNTIME` from
 //! lib-network. FFI callers may call from any thread; commands route through
 //! a tokio mpsc queue.
+//!
+//! # MSG-NODE-001 — errors + handshake parity
+//!
+//! `zhtp_quic_session_open` uses the **canonical** UHP initiator
+//! (`quic_handshake::handshake_as_initiator` / `quic_uhp_capabilities`).
+//! On failure it returns null and stores a structured last-error string
+//! (`stage=…: message`) retrievable via `zhtp_quic_session_last_error`.
 
 #![cfg(not(target_arch = "wasm32"))]
 
-use std::ffi::{c_char, CStr};
+use std::cell::RefCell;
+use std::ffi::{c_char, CStr, CString};
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
 
 use anyhow::{anyhow, Context, Result};
-use tokio::io::AsyncReadExt;
 use tokio::sync::oneshot;
 
 use lib_identity::types::IdentityType;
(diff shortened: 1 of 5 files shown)


umwelt (commented):
Solid, careful FFI PR — the memory safety is right and the stage-tagged errors are a real improvement over opaque `openFailed`. I traced the FFI boundary specifically:
- `zhtp_quic_session_last_error` returns null on no-error and on interior-NUL, `into_raw` paired with `from_raw` in `_free`, and `_free` is null-checked — no leak, no double-free, no null-deref.
- `_last_error_stage` returns static NUL-terminated byte strings (correctly 'do not free').
- The connect runs on the shared `CLIENT_ENDPOINT_RUNTIME` worker thread but propagates each stage's error back through the `ready` oneshot, and `set_last_error` runs on the **caller** thread in `open_session` — so the error is stored where the caller can read it.
- The golden wire test is a real byte-capture `assert_eq!`, so the `quic_uhp_capabilities` unification is guarded against silently changing the ClientHello. Good.

One contract issue worth addressing, plus two minors — inline.

lib-client/src/quic_session.rs
```diff
@@ -29,6 +36,79 @@ use lib_protocols::wire::read_response;
 
 use crate::identity::Identity;
 
+// =============================================================================
```
umwelt, line 47: **Thread-local error storage is an undocumented FFI footgun.** The error is stored per-thread and set on the caller thread (via the `ready` channel result in `open_session`), so it's only readable if mobile calls `zhtp_quic_session_last_error()` / `_stage()` on the **same OS thread** that called `zhtp_quic_session_open()`. The architecture makes that correct for a synchronous same-thread sequence, and thread-local even helps by keeping concurrent opens on different threads from clobbering each other.

But the mobile integration snippet in the PR shows the call sequence without stating the same-thread requirement, and Dart FFI / JNI can marshal calls onto different threads (helper isolates, thread pools). If `open()` runs on one thread and `last_error()` on another, the reader gets null — mobile is back to 'only openFailed', the exact failure this PR removes.

Either (a) document the same-thread requirement prominently at the FFI boundary, or (b) switch to a global `Mutex<Option<(&'static str, String)>>`. For a boundary where you don't control the consumer's threading, the global is the more robust contract — 'last error' is already last-writer-wins, so the only thing you trade is that concurrent opens on different threads can overwrite each other's error, which is usually acceptable and far less surprising than a silently-empty error.

lib-client/src/quic_session.rs
```diff
@@ -334,70 +509,98 @@ pub extern "C" fn zhtp_quic_session_open(
...
-            return std::ptr::null_mut();
-        }
+/// Stage token only (`uhp_handshake`, `quic`, `resolve`, …). Static C string
+/// — do not free. Returns `"none"` when no error is stored.
+#[no_mangle]
+pub extern "C" fn zhtp_quic_session_last_error_stage() -> *const c_char {
+    LAST_STAGE.with(|s| stage_as_c_str(*s.borrow()))
+}
+
+fn stage_as_c_str(stage: &str) -> *const c_char {
```
umwelt: Two parallel stage lists that must stay in sync: `STAGES` (line ~70, used by `classify_stage`) and this `stage_as_c_str` match. If they ever drift, `_last_error_stage()` returns `"none"` while the message string still carries the real `stage=…`. Consider deriving one from the other, or a unit test asserting every `STAGES` entry has a matching arm here (and vice-versa).

lib-client/src/quic_session.rs
```diff
@@ -315,6 +413,83 @@ fn to_zhtp_identity(id: &Identity) -> Result<ZhtpIdentity> {
...
+        Err(e) => {
+            set_last_error_from_anyhow("identity", &e);
+            return Err(e);
+        }
+    };
+
+    let (tx, rx) = tokio::sync::mpsc::channel::<SessionCmd>(64);
+    let (ready_tx, ready_rx) = oneshot::channel::<Result<()>>();
+    let worker = spawn_session_worker(addr, server_name, alpn, zhtp_identity, rx, ready_tx);
+
```
umwelt, line 521: `open_session` blocks the caller thread on `ready_rx.blocking_recv()` until the QUIC connect + UHP handshake completes or fails. That's fine and probably intended, but it means `zhtp_quic_session_open` is a blocking call — worth a one-line doc note so mobile calls it off the UI thread (a blocking network connect on the UI thread would jank). Not a code issue.

umwelt:
Addressed review:

| Item | Fix |
|------|-----|
| Thread-local last-error footgun (umwelt) | Process-global `Mutex` last-writer-wins; cross-thread unit test |
| Parallel STAGES / stage_as_c_str | Single `STAGES` table + `stage_table_c_str_covers_every_token` |
| Blocking open doc | Module + `zhtp_quic_session_open` docs: call off UI thread |
| eprintln noise (Copilot) | `tracing::debug!` only |
| `open_session` pub raw ptr (Copilot) | Now private (FFI + tests only) |
| inbound_read stale error (Copilot) | `clear_last_error()` on entry; success/timeout stay clear |

PR #2914, opened 2026-07-17
Fix/add updated domain prefix

### LINKED ISSUE #2746: crypto: add ZHTP-domain-update-v1 domain prefix to domain signing
## Context

Domain updates currently sign the wire message:
`domain|expected_previous_manifest_cid|new_manifest_cid|timestamp`

Identity attestations use domain separation (`ZHTP-identity-sig-v1\0`). Domain updates should eventually use `ZHTP-domain-update-v1\0` prefix for cross-protocol isolation.

## Task

Coordinate a breaking change across:
- `lib-network/src/web4/domain_signing.rs`
- `zhtp-cli` deploy/domain commands
- `lib-client/src/token_tx.rs::build_domain_update_request`
- API handlers in `zhtp/src/api/handlers/web4/domains.rs`

## Acceptance criteria

- [ ] Canonical prefix documented
- [ ] All signers and verifiers updated in same release
- [ ] Integration test with prefixed message

PeterGRutherford:
## Summary

Added `ZHTP-domain-update-v1\0` domain separation prefix to all domain update signing and verification paths. Domain update messages are now signed as `ZHTP-domain-update-v1\0domain|expected_previous_manifest_cid|new_manifest_cid|timestamp` instead of the raw pipe-delimited format. This prevents cross-protocol signature collisions, mirroring the existing `ZHTP-identity-sig-v1\0` pattern used in identity attestations.

## Linked issues

Closes #2746

## Architecture compliance

n/a

## Test plan

- [x] Unit tests added/updated (`lib-network/src/web4/domain_signing.rs`):
  - `domain_update_message_contains_prefix` — asserts message starts with `ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN`
  - `domain_update_signature_fails_without_prefix` — asserts old unprefixed signatures are rejected by the canonical verifier
  - `domain_update_roundtrip_sign_and_verify` — end-to-end sign + verify with prefix
- [x] `cargo check --workspace` passes (exit code 0)

## Risk and reversibility

**Breaking change:** old signatures (without the prefix) will fail verification. All nodes (signers in `lib-network`, `lib-client`, `zhtp-cli`; verifier in `zhtp` API handler) must be deployed together in the same release — no coordinated network restart needed, but rolling deployments will have a brief window where old signatures are rejected.

### DIFF
--- a/lib-network/src/web4/domain_signing.rs
+++ b/lib-network/src/web4/domain_signing.rs
@@ -1,13 +1,17 @@
 //! Canonical domain-update signing and Dilithium5 verification.
 //!
-//! Wire format matches zhtp-cli and lib-client:
-//!   `domain|expected_previous_manifest_cid|new_manifest_cid|timestamp`
+//! Wire format (with domain separation prefix):
+//!   `ZHTP-domain-update-v1\0domain|expected_previous_manifest_cid|new_manifest_cid|timestamp`
 
 use anyhow::{anyhow, Result};
 use lib_crypto::verify_signature;
 
 /// Hex-encoded Dilithium5 detached signatures are 4595 bytes → 9190 hex chars.
 pub const DILITHIUM5_HEX_SIGNATURE_LEN: usize = 9190;
+
+/// Domain separation prefix for domain update signatures.
+/// Mirrors the `ZHTP-identity-sig-v1\0` pattern used in identity attestations.
+pub const ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN: &[u8] = b"ZHTP-domain-update-v1\0";
 
 /// Returns true when a stored owner public key is present (full Dilithium5 size, non-zero).
 pub fn has_owner_signing_key(owner_dilithium_pk: &[u8]) -> bool {
@@ -23,18 +27,24 @@ fn owner_pk_array(owner_dilithium_pk: &[u8]) -> Result<[u8; 2592]> {
     Ok(arr)
 }
 
-/// Build the canonical domain-update signing message.
+/// Build the canonical domain-update signing message with domain separation prefix.
+///
+/// Format: `ZHTP-domain-update-v1\0domain|expected_previous_manifest_cid|new_manifest_cid|timestamp`
 pub fn domain_update_signing_message(
     domain: &str,
     expected_previous_manifest_cid: &str,
     new_manifest_cid: &str,
     timestamp: u64,
 ) -> Vec<u8> {
-    format!(
-        "{}|{}|{}|{}",
-        domain, expected_previous_manifest_cid, new_manifest_cid, timestamp
-    )
-    .into_bytes()
+    let mut message = ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN.to_vec();
+    message.extend_from_slice(
+        format!(
+            "{}|{}|{}|{}",
+            domain, expected_previous_manifest_cid, new_manifest_cid, timestamp
+        )
+        .as_bytes(),
+    );
+    message
 }
 
 /// Reject missing or malformed owner signatures before mutating domain state.
@@ -249,5 +259,67 @@ mod tests {
         )
         .expect("verify runs");
         assert!(!bad);
+    }
+
+    #[test]
+    fn domain_update_message_contains_prefix() {
+        let domain = "app.zhtp";
+        let prev = "bafkabc123";
+        let new_cid = "bafkdef456";
+        let timestamp = 1_700_000_000u64;
+
+        let message = domain_update_signing_message(domain, prev, new_cid, timestamp);
+
+        // Verify the prefix is present at the start of the message
+        assert!(
+            message.starts_with(ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN),
+            "Message must start with domain separation prefix"
+        );
+
+        // Verify the payload follows the prefix correctly
+        let payload = &message[ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN.len()..];
+        let expected_payload = format!(
+            "{}|{}|{}|{}",
+            domain, prev, new_cid, timestamp
+        );
+        assert_eq!(
+            payload,
+            expected_payload.as_bytes(),
+            "Payload after prefix must match pipe-delimited format"
+        );
+    }
+
+    #[test]
+    fn domain_update_signature_fails_without_prefix() {
+        // A signature created over the raw (unprefixed) message must fail verification
+        // against the canonical helper that expects the prefix.
+        let keypair = generate_keypair().expect("keypair");
+        let domain = "app.zhtp";
+        let prev = "bafkabc123";
+        let new_cid = "bafkdef456";
+        let timestamp = 1_700_000_000u64;
+
+        // Sign the OLD raw format (without prefix)
+        let raw_message = format!(
+            "{}|{}|{}|{}",
+            domain, prev, new_cid, timestamp
+        );
+        let sig = sign_message(&keypair, raw_message.as_bytes()).expect("sign");
+        let sig_hex = hex::encode(&sig.signature);
+
+        // Verify using the canonical helper (which expects the prefix) should FAIL
+        let ok = verify_domain_update_signature(
+            keypair.public_key.dilithium_pk.as_slice(),
+            domain,
+            prev,
+            new_cid,
+            timestamp,
+            &sig_hex,
+        )
+        .expect("verify runs");
+        assert!(
+            !ok,
+            "Signature over raw unprefixed message must fail verification with canonical helper"
+        );
     }
 }
--- a/lib-network/src/web4/domain_signing.rs
+++ b/lib-network/src/web4/domain_signing.rs
@@ -1,7 +1,13 @@
 //! Canonical domain-update signing and Dilithium5 verification.
 //!
-//! Wire format (with domain separation prefix):
-//!   `ZHTP-domain-update-v1\0domain|expected_previous_manifest_cid|new_manifest_cid|timestamp`
+//! ## Current wire format (signers must produce this)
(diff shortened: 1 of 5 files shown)


lib-network/src/web4/domain_signing.rs
```diff
@@ -1,14 +1,18 @@
...
 
 use anyhow::{anyhow, Result};
 use lib_crypto::verify_signature;
 
 /// Hex-encoded Dilithium5 detached signatures are 4595 bytes → 9190 hex chars.
 pub const DILITHIUM5_HEX_SIGNATURE_LEN: usize = 9190;
 
+/// Domain separation prefix for domain update signatures.
+/// Mirrors the `ZHTP-identity-sig-v1\0` pattern used in identity attestations.
+pub const ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN: &[u8] = b"ZHTP-domain-update-v1\0";
```
umwelt, line 20: This is the canonical `pub const` (and `domain_update_signing_message()` right below it) — good. The problem is nothing else imports it. `zhtp-cli/src/commands/deploy.rs`, `lib-client/src/token_tx.rs`, and the verifier in `zhtp/src/api/handlers/web4/domains.rs` each redefine `ZHTP-domain-update-v1\0` locally (or as a raw literal), and reconstruct the `domain|prev|new|ts` message inline instead of calling this builder. That's a signing/verification contract copy-pasted 4×: if any one drifts (a typo, a `v1`→`v2` bump in only some), signer and verifier silently disagree and **every** domain update fails verification with no compile error. Make this the single source of truth — export it, and have the CLI, client, and verifier import both the const and `domain_update_signing_message()`. If a crate-dependency direction blocks the import, hoist the const into `lib-crypto` next to the other `ZHTP-*-v1` domain tags.

zhtp/src/api/handlers/web4/domains.rs
```diff
@@ -1784,13 +1784,21 @@ impl Web4Handler {
...
-        let signed_message = format!(
-            "{}|{}|{}|{}",
-            update_request.domain,
-            update_request.expected_previous_manifest_cid,
-            update_request.new_manifest_cid,
-            update_request.timestamp
-        );
+        // Reconstruct the signed message with domain separation prefix and verify
+        let signed_message = {
+            let mut msg = b"ZHTP-domain-update-v1\0".to_vec();
```
umwelt: The verifier builds the signed message from a **raw byte literal** `b"ZHTP-domain-update-v1\0"` rather than the canonical const. This is the single most important place to not have a private copy — if the signer's prefix and this literal ever differ by one byte, every legitimate update is rejected here and there's nothing to catch it at build time. Import `ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN` / `domain_update_signing_message()` from `lib_network::web4::domain_signing` and use it on both sides.

umwelt: Transition: this verifier does a hard cutover — after deploy it accepts only prefixed signatures. Because the signature is checked at admission (not apply), the break is client↔gateway: a mobile or CLI client still on the old build signs unprefixed and gets rejected by an upgraded gateway, and mobile clients update on the app-store's schedule, not yours. Per the project's API-backward-compat rule, consider accepting **either** prefixed or unprefixed here for one release (try the prefixed verification, fall back to the legacy format), log when the legacy path is hit, then drop the fallback in the next release once clients have rolled. That turns a user-facing outage window into a clean two-step migration.

.cargo/config.toml
```diff
@@ -4,3 +4,15 @@
...
+# Windows: the main thread created by #[tokio::main] inherits the OS-default
+# 1 MB stack, which is too small for the deeply-nested startup futures in
+# RuntimeOrchestrator::start_node().  Raising the linker stack reserve to 16 MB
+# prevents "thread 'main' has overflowed its stack" crashes on stock Windows.
+# This is a linker-level setting and does NOT affect runtime memory usage
+# (only the reserved address space, not committed pages).
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "link-arg=/STACK:16777216"]
+
+[target.x86_64-pc-windows-gnu]
```
umwelt: Example of the scope issue from the summary: a Windows linker stack-size setting has nothing to do with domain-update signing. It may well be a correct fix (deep startup futures overflowing the 1 MB main-thread stack), but it belongs in the Windows-support PR, not here.

umwelt (changes requested):
The core change is legitimate — a domain-separation prefix for domain-update signatures prevents cross-protocol signature reuse, and mirroring the existing `ZHTP-identity-sig-v1\0` pattern is the right precedent. Tests are reasonable (prefix present, old sig rejected, roundtrip).

Good news on risk: I traced the apply path and **this is not a consensus/fork change**. `contracts.rs` processes `DomainUpdate` with a CAS on `expected_previous_manifest_cid` and then updates the record — it does **not** verify the signature. So the signature is an admission-layer gate only; replaying a historical `DomainUpdate` with the new binary can't diverge. That removes the scariest failure mode.

Three things to fix before merge:

1. **Split the unrelated Windows changes into their own PR.** This is titled/scoped as a domain-prefix fix but also silently changes IPC transport on Windows (`ipc/client.rs`/`server.rs` Unix-socket→TCP-loopback), adds a linker stack setting (`.cargo/config.toml`), and touches `bluetooth/discovery.rs` / `wifi_direct.rs`. Landing platform-transport changes under a 'domain prefix' title is a review hazard — someone approving the signature change is not signing up to review Windows IPC. Separate PR.

2. **Dedupe the security-critical prefix.** It's defined 4 different ways across the tree, and the verifier uses a raw byte literal. Details inline.

3. **Consider dual-accept for the transition** rather than a hard cutover — inline on the verifier.

umwelt:
## @PeterGRutherford — what we did + what you need next

Thanks for the original domain-prefix work. A formal review requested changes; we applied them on a **clean domain-only branch** and force-updated this PR head (`fix/add-updated-domain-prefix` → commits based on current `development`).

### What we did

| Step | Result |
|------|--------|
| New branch from `development` | `fix/domain-update-prefix-v1` |
| Cherry-pick **only** your domain commit | Dropped Windows IPC, stack, BT/wifi from this PR |
| Single signing helper | CLI + lib-client call `domain_update_signing_message()` — no local prefix copies |
| Dual-accept verify | Prefixed first, legacy unprefixed fallback (+ log) for one release |
| Tests | `cargo test -p lib-network --lib domain_signing` — 6 passed |

### What you should do next

1. **Pull / reset** your local branch to the updated PR head (history was rewritten — force-push already done):
   ```bash
   git fetch origin fix/add-updated-domain-prefix
   git checkout fix/add-updated-domain-prefix
   git reset --hard origin/fix/add-updated-domain-prefix
   ```
2. **Re-request review** once CI is green (address any remaining reviewer nits).
3. **Do not re-add** the Windows/IPC/stack commits onto this PR — use the split issues below.
4. After merge + one client release: implement **#2917** (drop legacy accept).

### Open issues for the rest of the work (from the old PR surface)

| Issue | Work |
|-------|------|
| **#2915** | Windows IPC: Unix socket → TCP loopback |
| **#2916** | Windows linker stack reserve (`.cargo/config.toml`) |
| **#2918** | BLE_MESH / Mutex import fixes (bluetooth, wifi_direct) |
| **#2917** | Drop dual-accept legacy domain-update signatures (follow-up) |

Those can be separate small PRs when you’re ready — they are real work, just not domain-signing.

### Your original commits

Preserved in git history on the old tip before force-push (`0c4ecddb`, `8400413f`, `424303f6`). Domain logic is rebased + fixed; Windows pieces should be re-landed via #2915/#2916/#2918.

cc reviewers: please re-check against the three original asks (scope / single const / dual-accept) — all three should now be satisfied.

umwelt:
### @PeterGRutherford — Windows follow-ups assigned to you

The non-domain work that was stripped from this PR is tracked and **assigned to you**:

| Issue | Work |
|-------|------|
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2915 | Windows IPC (Unix socket → TCP loopback) |
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2916 | Windows linker stack reserve (`.cargo/config.toml`) |
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2918 | BLE / wifi_direct import fixes |

Please pick those up as **separate PRs** (not on #2914). Domain-only path stays here for #2746 / dual-accept #2917.

umwelt:
### @PeterGRutherford — full follow-up list (assigned to you)

Including BLE/wifi and the dual-accept cleanup:

| Issue | Work |
|-------|------|
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2915 | Windows IPC (Unix socket → TCP loopback) |
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2916 | Windows linker stack reserve |
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2918 | **BLE / wifi_direct import fixes** |
| https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2917 | **After one release: drop legacy dual-accept** for domain-update signatures |

All four are assigned to you. Keep #2915 / #2916 / #2918 as separate Windows PRs; do **#2917** only after clients ship the prefixed signer (post–#2914 merge + one release).

umwelt (approved):
Re-review after cleanup commit `ecfd3a76`:

1. Windows IPC/stack/BLE noise removed (tracked in #2915–#2918).
2. Single canonical `domain_update_signing_message()` + `ZHTP_DOMAIN_UPDATE_SIGN_DOMAIN`; no private prefix copies.
3. Dual-accept verify (prefixed first, then legacy) for one release; drop tracked in #2917.

LGTM — approving so #2746 can close on merge.

PR #2920, opened 2026-07-18
fix(identity): close registration hardening issues #2768–#2777

### LINKED ISSUE #2768: identity/register returns success when chain enqueue fails

### LINKED ISSUE #2769: identity/register retry after partial failure creates new HD seed and wallet IDs
## Problem

Duplicate detection in `handle_register_client_identity` only checks `wallets_for_owner` (non-empty on-chain wallets). It does not check `query_identity_exists` / identity registry for an identity already committed without wallets.

A retry after partial failure runs `register_external_citizen_identity` again with a **new random HD master seed**, producing different wallet IDs, duplicate welcome-bonus mint attempts, and client/server wallet-ID drift.

## Location

`zhtp/src/api/handlers/identity/mod.rs` (~2052–2084)

## Suggestion

Before creating in-memory state, check on-chain identity existence. If identity exists without wallets, reuse deterministic derivation or explicit recovery path instead of full re-onboarding.

## Related

#2768, #2727

### LINKED ISSUE #2770: identity/register: reject future timestamps on registration proof
## Problem

Registration proof freshness only rejects stale timestamps (`now - timestamp > 300`). Future timestamps pass because `saturating_sub` yields 0.

## Location

`zhtp/src/api/handlers/identity/mod.rs` (~1922–1935)

## Suggestion

Reject if `req.timestamp` is more than ~60s ahead of `now`, matching bidirectional skew pattern used in Kyber key update handler.

## Related

#2768

### LINKED ISSUE #2771: OPAQUE genesis test: assert negative path when [opaque] absent
## Problem

`test_build_block0_loads_opaque_setup_when_present` early-returns when `config.opaque.is_none()`, so CI/configs without `[opaque]` get no coverage.

## Location

`lib-blockchain/src/genesis/mod.rs` (~1095)

## Suggestion

Add fixture genesis fragment with known minimal valid `server_setup_b64`, or assert negative path: opaque absent => `bc.opaque_server_setup.is_none()`.

## Related

OPAQUE lobby auth fix (build_block0 loads opaque setup)

### LINKED ISSUE #2772: identity/register: validate display_name before enqueueing system tx
## Problem

`display_name` in client identity registration is not validated with `validate_username` and is not checked via `identity_display_name_taken` (unlike `handle_claim_username`).

Colliding or policy-invalid names can be written into shadow cache and queued on-chain without a caller-friendly conflict response.

## Location

`zhtp/src/api/handlers/identity/mod.rs` (~2121–2125)

## Suggestion

Reuse `validate_username` on trimmed display name; reject with `409 Conflict` when `identity_display_name_taken` is true before `add_system_transaction`.

## Related

#2768

### LINKED ISSUE #2774: wallet/provision endpoint lacks auth gate (ops-only surface)
## Problem

`POST /api/v1/wallet/provision` (`handle_provision_wallet`) is exposed without visible auth/principal checks. Comment says ops-only / not end-user onboarding, but any reachable caller could register wallets and queue welcome-bonus TokenMint for arbitrary `owner_identity_id`.

## Location

`zhtp/src/api/handlers/wallet/mod.rs` (~1543–1753)

## Suggestion

Gate on `Role::Council` / node operator principal, or remove from public handler routing until GENESIS-3 ops auth is wired.

## Related

#2733, #2727

### LINKED ISSUE #2775: Add tests for ClientIdentityRegistration system-tx registration path
## Problem

No unit/integration tests cover the `add_system_transaction` + `ClientIdentityRegistration` path introduced in the identity registration fix.

Missing coverage for:
- Valid proof enqueues system tx with correct originator
- Invalid proof rejected
- Duplicate-with-wallets returns conflict
- Empty display name becomes `user-{prefix}` fallback
- Shadow insert tracks pending height

## Location

`zhtp/src/api/handlers/identity/mod.rs` (test module)

## Related

ClientIdentityRegistration fix, #2768

### LINKED ISSUE #2777: validate_system_injection: add ClientIdentityRegistration structural checks
## Problem

`validate_system_injection` has rules for TokenMint and funded WalletRegistration, but no originator-specific checks for `ClientIdentityRegistration`.

System injectors bypass mempool signature verification. API-layer proof is the trust boundary, but a compromised/buggy handler could inject arbitrary IdentityRegistration payloads (wrong public_key, empty ownership_proof, DID/key_id mismatch).

## Suggestion

For `SystemOriginator::ClientIdentityRegistration`, require:
- `IdentityRegistration` tx type
- Non-empty `ownership_proof`
- `tx.signature.public_key` Dilithium bytes match `identity_data.public_key`
- `identity_data.did == did:zhtp:{hex(key_id)}`

## Location

`lib-blockchain/src/blockchain.rs` `validate_system_injection`

## Related

ClientIdentityRegistration originator (identity registration fix)

umwelt:
## Summary

Validates and finishes the client-identity registration hardening set (#2768–#2777). Core fixes already shipped in #2779; this PR lands residual gaps and closes the open issues.

**Not stacked** with Windows PRs #2796 / #2797 — independent, bases on `development`.

| Issue | Status |
|-------|--------|
| #2768 false success on enqueue fail | Already hard-failed identity enqueue; **now** also hard-fails wallet queue failures + missing `blockchain_tx` |
| #2769 retry after partial failure | On-chain `identity_exists` gate (from #2779) |
| #2770 future timestamps | Bidirectional skew max +60s (from #2779) |
| #2771 OPAQUE genesis test | Negative path already present; positive path no longer silent early-return |
| #2772 display_name validation | `validate_username` + uniqueness (from #2779) |
| #2774 wallet/provision auth | Council/System gate (from #2779) |
| #2775 registration tests | Existing + empty-name fallback test |
| #2777 system injection checks | Structural validation + mempool tests (from #2779) |

Closes #2768
Closes #2769
Closes #2770
Closes #2771
Closes #2772
Closes #2774
Closes #2775
Closes #2777

## Test plan

- [x] `cargo test -p zhtp --lib test_register_identity`
- [x] `cargo test -p lib-blockchain --lib system_injection_`
- [x] `cargo test -p lib-blockchain --lib genesis::tests::test_build_block0_opaque_absent_means_none`

### DIFF
--- a/lib-blockchain/src/blockchain/identity.rs
+++ b/lib-blockchain/src/blockchain/identity.rs
@@ -6,6 +6,63 @@ impl Blockchain {
         self.identity_registry.insert(did, data);
     }
 
+    /// Roll back a partially applied client identity registration that has not
+    /// yet been included in a block.
+    ///
+    /// Removes mempool txs (identity / wallet registrations / welcome mint),
+    /// identity + wallet shadows, and height indexes so a later retry can
+    /// re-onboard cleanly (#2768 / #2769). Does not touch committed blocks.
+    ///
+    /// After removing explicitly tracked `queued_txs`, also sweeps any remaining
+    /// pending txs that still match the DID or wallet ids (covers tracking gaps
+    /// if a mempool lookup missed a just-enqueued entry).
+    pub fn abort_pending_client_registration(
+        &mut self,
+        queued_txs: &[Transaction],
+        did: &str,
+        wallet_id_hexes: &[&str],
+    ) {
+        self.remove_pending_transactions(queued_txs);
+
+        let related: Vec<Transaction> = self
+            .pending_transactions
+            .iter()
+            .filter(|tx| {
+                if tx
+                    .identity_data()
+                    .is_some_and(|data| data.did == did)
+                {
+                    return true;
+                }
+                if let Some(wallet) = tx.wallet_data() {
+                    let wid = hex::encode(wallet.wallet_id.as_bytes());
+                    if wallet_id_hexes.iter().any(|h| *h == wid) {
+                        return true;
+                    }
+                }
+                // Welcome-bonus TokenMint targets the primary wallet_id in `to`.
+                if let Some(mint) = tx.token_mint_data() {
+                    let to_hex = hex::encode(mint.to);
+                    if wallet_id_hexes.iter().any(|h| *h == to_hex) {
+                        return true;
+                    }
+                }
+                false
+            })
+            .cloned()
+            .collect();
+        if !related.is_empty() {
+            self.remove_pending_transactions(&related);
+        }
+
+        self.identity_registry.remove(did);
+        self.identity_blocks.remove(did);
+        for wallet_id in wallet_id_hexes {
+            self.wallet_registry.remove(*wallet_id);
+            self.wallet_blocks.remove(*wallet_id);
+        }
+    }
+
     /// In-memory shadow only — test premises that must not use sled-first facades.
     pub fn identity_shadow_contains_key(&self, did: &str) -> bool {
         self.identity_registry.contains_key(did)
--- a/lib-blockchain/src/blockchain/identity.rs
+++ b/lib-blockchain/src/blockchain/identity.rs
@@ -16,6 +16,10 @@ impl Blockchain {
     /// After removing explicitly tracked `queued_txs`, also sweeps any remaining
     /// pending txs that still match the DID or wallet ids (covers tracking gaps
     /// if a mempool lookup missed a just-enqueued entry).
+    ///
+    /// **Lock ordering (callers):** acquire `blockchain` write lock first, call
+    /// this, drop that lock, then acquire `identity_manager` for local remove.
+    /// Never hold `identity_manager` while waiting for `blockchain` write.
     pub fn abort_pending_client_registration(
         &mut self,
         queued_txs: &[Transaction],
--- a/zhtp/src/api/handlers/identity/mod.rs
+++ b/zhtp/src/api/handlers/identity/mod.rs
@@ -80,6 +80,43 @@ fn pending_tx_by_hash(
         .cloned()
 }
 
+/// Register one citizen wallet, track its pending tx, or abort the whole commit.
+fn register_wallet_or_abort(
+    blockchain: &mut lib_blockchain::Blockchain,
+    wallet_data: lib_blockchain::transaction::WalletTransactionData,
+    label: &str,
+    queued_txs: &mut Vec<Transaction>,
+    did: &str,
+    wallet_id_hexes: &[&str],
+) -> Result<(), (ZhtpStatus, String)> {
+    match blockchain.register_wallet(wallet_data) {
+        Ok(hash) => match pending_tx_by_hash(blockchain, &hash) {
+            Some(tx) => {
+                queued_txs.push(tx);
+                Ok(())
+            }
+            None => {
+                blockchain.abort_pending_client_registration(queued_txs, did, wallet_id_hexes);
+                Err((
+                    ZhtpStatus::InternalServerError,
+                    format!(
+                        "Failed to track {} wallet for registration rollback",
+                        label
+                    ),
+                ))
+            }
+        },
+        Err(e) => {
+            tracing::error!("Failed to register {} wallet: {}", label, e);
+            blockchain.abort_pending_client_registration(queued_txs, did, wallet_id_hexes);
+            Err((
+                ZhtpStatus::InternalServerError,
+                format!("Failed to queue {} wallet: {}", label, e),
+            ))
+        }
+    }
+}
+
 /// Clean identity handler implementation
 pub struct IdentityHandler {
     identity_manager: Arc<RwLock<IdentityManager>>,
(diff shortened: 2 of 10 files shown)


umwelt:
**Stack note:** Independent of the Windows stack (#2796 → #2797). Base remains `development`. Will not merge until your code review.

zhtp/src/api/handlers/identity/mod.rs
```diff
@@ -2276,7 +2284,16 @@ impl IdentityHandler {
                 initial_balance: 0,
             };
             if let Err(e) = blockchain.register_wallet(ubi_wallet_data) {
-                tracing::warn!("Failed to register UBI wallet: {}", e);
+                tracing::error!("Failed to register UBI wallet: {}", e);
+                let mut identity_manager = self.identity_manager.write().await;
```
umwelt: Representative of the incomplete-rollback bug: reaching here means the identity tx (~2223) and the primary wallet tx + `wallet_registry` entry (~2249) are already queued via `add_system_transaction`, and the welcome-bonus mint (~2261) is queued too. `remove_identity` undoes none of that — it only clears the in-memory identity manager. Returning 500 here leaves identity + primary wallet + 50 SOV bound for the next block while telling the client it failed. Either reorder to validate-then-commit, or extend every rollback path to `blockchain.remove_pending_transactions(...)` the already-queued txs and undo the wallet-registry inserts.

zhtp/src/api/handlers/identity/mod.rs
```diff
@@ -2247,19 +2247,27 @@ impl IdentityHandler {
...
+                identity_manager.remove_identity(&identity_id);
+                return Ok(ZhtpResponse::error(
+                    ZhtpStatus::InternalServerError,
+                    format!(
+                        "Failed to queue primary wallet (identity may be pending on chain): {}",
+                        e
+                    ),
+                ));
             }
+            queue_welcome_bonus_token_mint(
```
umwelt: The welcome-bonus mint is queued right after the primary wallet succeeds — so a UBI or savings failure below leaves an orphaned 50 SOV TokenMint to a primary wallet whose identity the handler just 'removed' (in-memory only). This mint isn't reversed by any of the rollback paths. Part of the same finding: nothing queued to the blockchain before the failure point is actually undone.

zhtp/src/api/handlers/identity/mod.rs
```diff
@@ -2296,28 +2313,45 @@ impl IdentityHandler {
...
-                    primary_wallet_id: primary_wallet_id.clone(),
-                    ubi_wallet_id: ubi_wallet_id.clone(),
-                    savings_wallet_id: savings_wallet_id.clone(),
-                    registered_at: now,
-                },
-            );
-        }
+        // Never report success without a chain identity tx hash (false-success path).
+        let Some(ref tx_hash) = blockchain_tx_hash else {
+            let mut identity_manager = self.identity_manager.write().await;
```
umwelt: The no-tx-hash guard is correct in intent (never report success without a chain tx), but note that by the time control reaches here the three wallet registrations have already run and queued their txs + registry entries. If `blockchain_tx_hash` is `None` at this point, you're removing the identity from the manager while three wallet txs remain queued for an identity that will never be projected. Same root cause — the rollback needs to cover the wallets too, or the wallets must not be registered until the identity tx hash is confirmed.

umwelt (commented):
The direction is right — turning silent `warn`-and-continue into hard failures is the correct fix for #2768, and the genesis test (both opaque branches) and empty-name fallback test are good additions. But the **rollback is incomplete in a way that recreates #2769** (the exact issue this PR claims to close), and I'd block on it.

**The problem:** the identity tx is queued at line ~2223 (`blockchain_tx_hash = Some(...)`) *before* the three wallet registrations, and `register_wallet` (wallets.rs:750) both calls `add_system_transaction` (queues a real tx for the next block) *and* inserts into `wallet_registry`. So by the time any later step fails, real on-chain-bound state already exists. Every failure path then rolls back with only `identity_manager.remove_identity(&identity_id)` — which touches the in-memory identity manager and **nothing on the blockchain**.

**Concrete failure:** UBI-wallet registration fails. The handler returns 500. But already queued for the next block and NOT rolled back: (1) the identity registration tx, (2) the primary wallet tx + its `wallet_registry` entry, (3) the 50 SOV welcome-bonus mint. The next block includes all three. The user's identity + primary wallet + bonus are on-chain, but the API told them registration failed — and when they retry, the on-chain `identity_exists` gate (the #2769 fix in this same PR) rejects the retry as a duplicate. The user is now stuck: reported-failed, can't retry, half-committed on-chain, with no UBI/savings wallet. That's arguably worse than the false-success it replaces. The error strings ('identity may be pending on chain', 'identity/primary may be pending on chain') show the author already senses this.

**The fix is available and not used.** `Blockchain::remove_pending_transactions(&[Transaction])` exists (blockchain.rs:2832). Two options:
- **Validate-then-commit (preferred):** run every precondition up front — `wallet_exists` for all three wallet ids, name validation, the identity_exists gate — and only call the identity enqueue + `register_wallet` calls once all are known-good. Then a failure aborts before any on-chain state and `remove_identity` is a complete rollback.
- **Complete the rollback:** on failure, also `remove_pending_transactions` for the identity tx + each already-registered wallet tx + the welcome mint, and undo the `wallet_registry` / `wallet_blocks` inserts. Messier (there's no `unregister_wallet`, so you'd need to add one), and racy against block production — which is why the first option is cleaner.

Inline notes mark the specific spots.

umwelt:
## Review feedback addressed (`1e1ecbf0`)

### Blocker: incomplete rollback recreating #2769 (umwelt + Copilot)
- **Validate-then-commit** under the blockchain write lock: display name free, identity absent, all three wallets absent — before any enqueue.
- Track every enqueued tx (identity, wallets, welcome mint). On any mid-commit failure call `Blockchain::abort_pending_client_registration`, which:
  - `remove_pending_transactions` for tracked txs
  - sweeps remaining mempool txs matching the DID / wallet ids / welcome TokenMint `to`
  - clears identity + wallet shadows and height indexes
- Then remove the in-memory `IdentityManager` entry so retries are clean.
- Hard-fail if a wallet hash cannot be resolved for rollback tracking (avoids silent tracking gaps).

### Copilot: parallel test races on global blockchain
- Added `global_blockchain_test_lock()` mutex around tests that call `set_global_blockchain`.

### Copilot: OPAQUE genesis test comment
- Reworded to assert correct behavior for whichever embedded genesis is present (not “both branches in one run”); forced-absent path remains in a separate test.

Tests run:
- `cargo test -p lib-blockchain --lib abort_pending_client_registration`
- `cargo test -p zhtp --lib test_register_identity`

Not merging until your code review.

umwelt (commented):
Re-review of `1e1ecbf0` — **the critical rollback finding is resolved, correctly and completely.** I traced the new path end to end:

- **Race-free.** A single `blockchain.write()` guard is acquired once and held across the whole sequence — identity enqueue, all three `register_wallet` calls, and every `abort_pending_client_registration`. Block production needs that same write lock, so no block can be produced in a gap and lock in the txs before rollback. This is the property that makes the abort meaningful.
- **Complete teardown.** `abort_pending_client_registration` removes the tracked txs, then defensively sweeps `pending_transactions` for anything matching the DID, the wallet ids, or the welcome-mint (`mint.to == primary`), and clears all four in-memory maps (`identity_registry`, `identity_blocks`, `wallet_registry`, `wallet_blocks`).
- **Real removal, not just a list edit.** `remove_pending_transactions` drains `pending_transactions` (which block production iterates at blockchain.rs:2324), updates `mempool_state` admit accounting, and deletes the durable `store` entry. So a rolled-back registration leaves nothing queued, admitted, or persisted.
- **Tested.** `mempool_admit_tests` asserts `pending_transactions.is_empty()` after abort and checks admit-state consistency.

The #2769-in-a-new-form scenario from the last review (500 returned while identity + primary wallet + 50 SOV land on-chain, retry blocked by the dedup gate) can no longer happen. Nice.

Two minor, non-blocking notes inline — both about the fact that correctness here now rests on holding the blockchain write lock across the full critical section.

zhtp/src/api/handlers/identity/mod.rs
```diff
@@ -2182,6 +2196,10 @@ impl IdentityHandler {
         {
             let mut blockchain = blockchain_arc.write().await;
```
umwelt, line 2252: This single `blockchain.write()` guard is what makes the rollback race-free — good — but it's now held across every `.await` in the registration critical section (`register_wallet`, `queue_welcome_bonus_token_mint`, the aborts). That serializes *all* blockchain writes, including block production, for the full duration of a registration. It's the right trade for atomicity, but worth confirming nothing inside the section awaits on something slow or network-bound; a stall here stalls block production cluster-wide. A one-line comment noting 'lock held deliberately across the commit for atomic rollback' would flag it for the next reader.

lib-blockchain/src/blockchain/identity.rs
```diff
@@ -6,6 +6,63 @@ impl Blockchain {
...
+    /// yet been included in a block.
+    ///
+    /// Removes mempool txs (identity / wallet registrations / welcome mint),
+    /// identity + wallet shadows, and height indexes so a later retry can
+    /// re-onboard cleanly (#2768 / #2769). Does not touch committed blocks.
+    ///
+    /// After removing explicitly tracked `queued_txs`, also sweeps any remaining
+    /// pending txs that still match the DID or wallet ids (covers tracking gaps
+    /// if a mempool lookup missed a just-enqueued entry).
+    pub fn abort_pending_client_registration(
```
umwelt, line 23: The teardown is thorough. One consistency check worth doing (not in this diff): the handler holds the `blockchain` write lock and then acquires the `identity_manager` write lock inside it (for `remove_identity`). That ordering — blockchain-before-identity_manager — must be the same everywhere, or two paths taking them in opposite orders will deadlock. Given how many handlers touch both, a quick grep to confirm the ordering is uniform would be worth it. Not something this PR introduced, but this PR adds another site that nests them.

umwelt:
## Follow-up review notes addressed

### Already resolved (prior `1e1ecbf0`, confirmed by re-review)
- Full mempool + shadow rollback via `abort_pending_client_registration`
- Validate-then-commit under a single blockchain write
- Global blockchain test mutex + OPAQUE genesis comment reword

### New in this push
1. **Lock held across commit** — explicit comment that the write lock is deliberate for atomic rollback (block production shares it).
2. **No slow await under the lock** — welcome-bonus TokenMint is built (keystore load + sign) *before* `blockchain.write()`; only `add_system_transaction` runs under the lock.
3. **Lock ordering** — commit failures abort on-chain state under the blockchain lock, then drop it, then take `identity_manager` for local remove. Documented on `abort_pending_client_registration`. Avoids nesting that could deadlock with reverse-order paths.
4. **no-tx-hash path** — success path now yields `tx_hash` only from a successful commit `Result` (no partial wallets without a hash).

Tests: `abort_pending_client_registration_*` + all `test_register_identity*`.

umwelt:
## Deployed to testnet (2026-07-19)

**Commit:** `508d992a` (`development`) — identity registration hardening (#2920) + domain compile fix (#2921)

| Node | Binary MD5 | Status |
|------|------------|--------|
| zhtp-g1 | `b57b288015c7389a96b5c3fd46dcd9b2` | active, consensus epoch 6, height ~89751+ |
| zhtp-g2 | same | active |
| zhtp-g3 | same | active |
| zhtp-gateway | same | active |
| zhtp-gateway-2 | — | **not deployed** — SSH publickey denied from deploy host |

Consensus was halted at height 89741/89742 for the validator rollout, then restarted; chain is producing again.

PR #2919, opened 2026-07-18
feat(lib-client): AssetLaunch C FFI + domain 100 SOV / asset_id V3 (mobile M1/M4)

umwelt:
## Summary

Mobile M1/M4 surface for DAO launch:

- **`zhtp_client_build_asset_launch`** C FFI wraps `build_asset_launch_tx` (hex tx for `POST /api/v1/assets/launch`).
- Domain registration fee client constant **10 → 100 SOV** (matches `DEFAULT_DOMAIN_REGISTRATION_FEE_ATOMS` / live `TxFeeConfig`).
- Optional **`asset_id`** threads client → system-tx memo → server so V3 (`DOMREG3`) + `domain_tx_signature_hex` stay byte-identical.

### Review follow-ups

- Manifest pointers fail closed (both null → derive; both set → copy; mixed → NULL).
- AssetLaunch supply uses **`initial_supply_atoms_lo` / `_hi`** (Swift/Kotlin-safe; not bare `u128`).
- Shared `parse_optional_asset_id_hex` on server; fee-display paths use live whole-SOV fee / `DOMAIN_REGISTRATION_FEE`.
- Tests: FFI-default launch round-trip, client V3 memo decode, server V3 signature match/mismatch.
- Documented **C ABI break** on domain-register FFI trailing `asset_id_hex` and **client fee coupling** to governance-tunable `TxFeeConfig`.

### Mobile / release notes

- **Breaking C ABI (M4):** `zhtp_client_build_domain_register_request_with_fee_payment` gained trailing `asset_id_hex` — regenerate headers; pass `NULL` for non-DAO domains.
- **New AssetLaunch ABI:** supply is two `u64` halves (`hi << 64 | lo`).
- Fee signing still uses whole-SOV `DOMAIN_REGISTRATION_FEE` (100); if governance changes the fee, rebuild clients or query the node first.

## Test plan

- [x] `cargo test -p lib-client --lib domain_system_tx_tests`
- [x] `cargo test -p lib-client --lib ffi_defaults_local_manifest`
- [x] `cargo test -p zhtp --lib v3_domain_tx_signature`
- [ ] CI green on PR

PR #2921, opened 2026-07-19
fix(domain): restore zhtp compile after #2919 (no zhtp_client in node)

umwelt:
## Summary
- Development fails to compile: `domains.rs` uses `zhtp_client::…` but `zhtp_client` is only a **dev-dependency** of `zhtp` (E0433 from #2919).
- Canonicalize `DEFAULT_DOMAIN_REGISTRATION_FEE_WHOLE` + `parse_optional_asset_id_hex` in **lib-blockchain**.
- Node uses lib-blockchain; lib-client re-exports the same symbols for FFI stability.

## Test plan
- [x] `cargo check -p zhtp --lib --features full-blockchain`
- [x] `cargo check -p lib-client`
- [x] `cargo test -p lib-blockchain --lib parse_optional_asset_id`

## Notes
Scoped compile fix only — does not include identity registration rollback (#2920).

### DIFF
--- a/lib-blockchain/src/transaction/domain.rs
+++ b/lib-blockchain/src/transaction/domain.rs
@@ -25,6 +25,32 @@ pub const DOMAIN_REGISTRATION_PREFIX_V2: &[u8] = b"DOMREG2:";
 pub const DOMAIN_REGISTRATION_PREFIX_V3: &[u8] = b"DOMREG3:";
 pub const DOMAIN_UPDATE_PREFIX: &[u8] = b"DOMUPD1:";
 
+/// Parse optional 64-char (optionally `0x`-prefixed) hex into a 32-byte asset id.
+///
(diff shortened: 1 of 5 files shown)


umwelt:
## Deployed to testnet (2026-07-19)

**Commit:** `508d992a` (`development`) — identity registration hardening (#2920) + domain compile fix (#2921)

| Node | Binary MD5 | Status |
|------|------------|--------|
| zhtp-g1 | `b57b288015c7389a96b5c3fd46dcd9b2` | active, consensus epoch 6, height ~89751+ |
| zhtp-g2 | same | active |
| zhtp-g3 | same | active |
| zhtp-gateway | same | active |
| zhtp-gateway-2 | — | **not deployed** — SSH publickey denied from deploy host |

Consensus was halted at height 89741/89742 for the validator rollout, then restarted; chain is producing again.

(Also covers this compile fix; shipped in the same binary as #2920.)

Issue #2002, opened 2026-03-26
[ZHTP-AUTH][Phase 2D] Add restart/replay equivalence suites for projection rebuilds


umwelt:
## Summary
Add restart/replay equivalence tests for identity, wallet, and token projection rebuild behavior.

## Parent
- #1985 `[ZHTP-AUTH][Phase 2] Harden replay and rebuild identity projection from canonical history`

## Primary Agent Ownership
- Primary: Agent 10 — QA and Release Readiness Agent
- Secondary reviewer: Agent 1 — Blockchain Core State Agent
- Secondary reviewer: Agent 3 — Token Consensus Agent
- Security signoff: Agent 8 — Security and Replay Assurance Agent

## Problem
Phase 2 needs executable proof that replay rebuilds local state and projections without changing canonical outcomes.

## Scope
- add restart/replay equivalence tests for:
  - identities
  - wallets
  - token balances
  - total token supply
- validate projection rebuild behavior after committed-state restart
- ensure tests fail on ghost-state creation or projection drift

## Acceptance Criteria
- Automated restart/replay suites cover identity, wallet, balances, and supply equivalence.
- Tests fail on projection drift or supply mismatch.
- QA gate references these suites explicitly.

Issue #2000, opened 2026-03-26
[ZHTP-AUTH][Phase 2B] Instrument legacy replay/startup fixups and define removal gates


umwelt:
## Summary
Instrument legacy fixup paths in replay/startup and define evidence-based removal gates.

## Parent
- #1985 `[ZHTP-AUTH][Phase 2] Harden replay and rebuild identity projection from canonical history`

## Primary Agent Ownership
- Primary: Agent 10 — QA and Release Readiness Agent
- Secondary reviewer: Agent 1 — Blockchain Core State Agent
- Secondary reviewer: Agent 2 — Storage and Atomicity Agent

## Problem
Legacy fixups may still be altering startup state, but without telemetry they cannot be safely evaluated or removed.

## Scope
- instrument these paths at minimum:
  - `repair_backfill_inflation`
  - `migrate_sov_key_balances_to_wallets`
  - `backfill_token_balances_from_contract`
- emit metrics/logging on firing frequency and affected entities
- document removal gate: zero firings for two weeks across target environments

## Acceptance Criteria
- Telemetry exists for all targeted legacy fixup paths.
- Operational gate for removal is documented and measurable.
- Rollout dashboards/log review procedure is defined or referenced.

umwelt:
Progress update:

Implemented startup/replay telemetry for the targeted legacy fixup paths in `lib-blockchain/src/blockchain.rs`.

What changed locally:
- added structured per-fixup reporting (`affected_entities`, `affected_raw_units`) for:
  - `migrate_sov_key_balances_to_wallets`
  - `repair_backfill_inflation`
  - `backfill_token_balances_from_contract` call-site in `load_from_store()`
- added a single startup summary log for each path:
  - `LEGACY_FIXUP_SUMMARY path=load_from_store ...`
  - `LEGACY_FIXUP_SUMMARY path=load_from_file ...`
- summary includes the explicit removal gate text:
  - `zero firings for 14 days across target environments`
- summary severity is `warn!` if any fixup fired and `info!` if all stayed at zero.

Focused verification passing:
- `cargo test -p lib-blockchain test_legacy_fixup_summary_includes_gate_and_counts -- --nocapture`
- `cargo test -p lib-blockchain test_load_from_store_rebuilds_missing_identity_projection -- --nocapture`
- `cargo test -p lib-blockchain test_load_from_store_replay_removes_stale_revoked_identity_projection -- --nocapture`
[...]

Issue #1985, opened 2026-03-26
[ZHTP-AUTH][Phase 2] Harden replay and rebuild identity projection from canonical history


umwelt:
## Summary
Harden replay in `load_from_store` so helper persistence is rebuilt from canonical chain history and legacy fixups are observable in production.

## Parent Epic
- #1982 `[EPIC] ZHTP Blockchain-Authoritative State: committed blocks as the only source of truth`

## Related Blocker
- [ ] #1993 Eliminate startup SOV backfill minting from non-block state

## Child Issues
- [ ] #1999 Phase 2A: rebuild identity sled projection during canonical replay
- [ ] #2000 Phase 2B: instrument legacy replay/startup fixups and define removal gates
- [ ] #2001 Phase 2C: remove replay/load token reconstruction from non-block snapshot state
- [ ] #2002 Phase 2D: add restart/replay equivalence suites for projection rebuilds

## Required Changes
- after replay populates the in-memory identity registry from blocks, also rebuild the sled-backed identity projection via `persist_identity_registration()` or equivalent canonical projection path
- ensure replay rebuilds local identity trees deterministically instead of relying on stale preexisting sled content
- identify and remove any replay/load path that materializes token balances or supply from snapshots rather than committed block transactions
- instrument legacy fixups such as:
  - `repair_backfill_inflation`
  - `migrate_sov_key_balances_to_wallets`
  - `backfill_token_balances_from_contract`
- emit telemetry/metrics/logging sufficient to detect when these code paths still fire in production
- define removal gate: zero firings for two weeks across target environments

## Invariants
- Replay remains deterministic and canonical.
- Projection rebuilds do not create an alternate mutation path.
- Replay/load does not mint or recreate token balances from non-block state.
- Legacy repair logic is observable and removable on evidence.

## Acceptance Criteria
- Replayed identity data repopulates local sled projection consistently.
- Restart/replay tests prove equivalence between pre-restart committed state and post-restart rebuilt state.
- No startup backfill path can create ghost balances or excess token supply.
- Telemetry exists for all legacy fixup paths targeted for later removal.
- Removal gate for Phase 5 is documented.

umwelt:
[...]

Issue #1982, opened 2026-03-26
[EPIC] ZHTP Blockchain-Authoritative State: committed blocks as the only source of truth


umwelt:
## Summary
ZHTP currently behaves too much like an application with an eager local database and not enough like a blockchain system with block-authoritative state. Identity, wallet, and token-related state can be mutated before a transaction is included in a committed block, and local stores such as DHT/sled/in-memory maps can effectively become source-of-truth candidates.

This epic makes committed blocks the only authoritative source of ZHTP state and moves all API/runtime/local persistence behavior behind canonical block execution and deterministic replay.

## Problem
The current model risks consensus and recovery drift:
- API-time writes can report success before block inclusion.
- Restart can lose or diverge state that was only present in memory or non-canonical stores.
- Replay and cross-node recovery can disagree with pre-restart in-memory state.
- DHT and helper persistence layers are too close to authority instead of being projections/caches.
- Confirmed audit divergence shows non-block wallet insertion and non-block SOV mint/backfill behavior still exist.

## Goal
Establish this invariant across ZHTP and blockchain integration:
- Authoritative state is whatever can be deterministically derived from committed blocks.
- Pending transactions are availability concerns, not authority.
- DHT, sled helper trees, blockchain.dat snapshots, and in-memory registries are projections or caches and must be rebuildable from canonical chain data.

## Invariants
- No canonical identity, wallet, or token state is mutated outside block processing / commit flow.
- Restarted node state equals pre-restart committed state.
- Cross-node replay converges to the same state.
- API responses for registration flows are queue/inclusion aware, not prematurely final.
- DHT is never authoritative for consensus-relevant state.
- Total token supply is derivable from canonical history only.

## Scope
This epic covers:
- audit of replay-vs-runtime divergence
- removal of pre-block writes in wallet/identity/peer paths
- elimination of startup non-block mint/backfill behavior
- replay hardening and projection rebuilds
- wallet sled tree introduction
- DHT demotion to cache/projection-only role
- legacy workaround telemetry and eventual removal
- required restart/replay/crash-safety test coverage

## Out of Scope
- changing transaction types
- changing block format or serialization
- consensus/BFT redesign
- rewriting BlockExecutor logic
- introducing new token execution models

## Risk
Highest immediate risks are:
- queue-only cutover without durable pending transactions
- startup/replay code manufacturing balances or supply from non-block state

## Mitigation
- persist and reload pending registration transactions from sled/mempool
[...]

PR #2924, opened 2026-07-21
feat(zhtp-auth): Phase 2 identity projection + reward-claim admit safety

umwelt:
## Stack (bottom — review first)
```
development
  └── #2924 (this) Phase 2 identity projection rebuild
        └── #2925 DHT cache-only + chain-first PoUW
```

## Summary
- Rebuild sled identity projections from block-scan registry on `load_from_store` (#1985/#1999)
- Legacy fixup telemetry `target=legacy_fixup` for Phase 5 removal gate (#2000)
- Identity projection restart equivalence tests (#2002)
- RewardClaim mempool/API admit requires token contract (consensus halt safety)
- Phase 1 queue-only registration / pending-tx durability pieces in tree

## Related
- Epic #1982 · `docs/ops/legacy-fixup-removal-gate.md`
- Stacked child: #2925

## Test plan (local after your review)
- `cargo test -p lib-blockchain --test identity_projection_restart_tests`
- focused mempool admit tests
- Then `cargo clean`

### DIFF
--- a/lib-blockchain/src/transaction/validation.rs
+++ b/lib-blockchain/src/transaction/validation.rs
@@ -2554,18 +2554,35 @@ impl<'a> StatefulTransactionValidator<'a> {
             tracing::debug!("[BREADCRUMB] validate_zk_proofs OK");
         }
 
-        // RewardClaim: must match apply (`reward_claim::apply_reward_claim`) so a
-        // claim that cannot execute never enters the mempool or a BFT proposal.
-        // Order is intentional and identical to apply:
-        //   1) owner_did registered in sled
-        //   2) token_contract MUST exist (rewards-module alone is not enough —
-        //      AssetLaunch can write rewards state without a token_contracts row;
-        //      admitting such claims caused the GENESIS-3 H=1920 halt)
-        //   3) spend-delegate (if rewards module) else legacy token-creator
+        // RewardClaim mempool filter (best-effort early reject).
+        // Stateless / tip-stable checks mirror apply. Stateful eligibility
+        // (welcome/daily/partner already claimed) can race tip vs apply height
+        // and is soft-dropped at apply (`TxApplyError::RewardClaimDropped`)
+        // so unmirrored rejects cannot halt consensus after BFT finality.
+        //
+        // Order for tip-stable checks (matches apply):
+        //   0) amount > 0
+        //   1) recipient_key_id == key_id_from_did(owner_did)
+        //   2) owner_did registered in sled
+        //   3) token_contract exists (GENESIS-3 H=1920: rewards-module alone
+        //      is not enough — pure AssetLaunch has no token_contracts row)
+        //   4) spend-delegate (if rewards module) else legacy token-creator
         if transaction.transaction_type == TransactionType::RewardClaim {
             let data = transaction
                 .reward_claim_data()
                 .ok_or(ValidationError::MissingRequiredData)?;
+            if data.amount == 0 {
+                tracing::warn!("[REWARD_CLAIM] amount must be greater than 0");
+                return Err(ValidationError::InvalidTransaction);
+            }
+            if crate::transaction::reward_claim::key_id_from_did(&data.owner_did)
+                != Some(data.recipient_key_id)
+            {
+                tracing::warn!(
+                    "[REWARD_CLAIM] recipient_key_id does not match owner_did"
+                );
+                return Err(ValidationError::InvalidTransaction);
+            }
             let blockchain = self
                 .blockchain
                 .ok_or(ValidationError::InvalidTransaction)?;
@@ -2582,10 +2599,6 @@ impl<'a> StatefulTransactionValidator<'a> {
                 );
                 return Err(ValidationError::InvalidTransaction);
             }
-            // Apply hard-requires token_contract first. Checking rewards_module
-            // before contract previously admitted pure-AssetLaunch BUBL claims
-            // that then failed apply with "token contract not found" and halted
-            // consensus after BFT finality.
             let Some(contract) = blockchain.get_token_contract(&data.token_id) else {
                 tracing::warn!(
                     "[REWARD_CLAIM] token contract not found for claim token_id={}",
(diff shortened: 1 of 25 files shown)


lib-blockchain/src/transaction/validation.rs
```diff
@@ -2547,11 +2554,14 @@ impl<'a> StatefulTransactionValidator<'a> {
...
         }
 
-        // RewardClaim: signer is the spend delegate (system tx), but beneficiary
-        // `owner_did` must exist in durable sled — same oracle as executor apply.
-        // Spend-delegate (or legacy token-creator) authorization must also match
-        // apply (`reward_claim.rs`) so a non-delegate claim is rejected at
-        // mempool admission instead of halting block apply (#2859 class).
+        // RewardClaim: must match apply (`reward_claim::apply_reward_claim`) so a
+        // claim that cannot execute never enters the mempool or a BFT proposal.
+        // Order is intentional and identical to apply:
```
umwelt: "identical to apply" is the goal, but it currently mirrors 5 of apply's ~13 reject conditions. `apply_reward_claim` also rejects on: `recipient_key_id != key_id_from_did(owner_did)` (reward_claim.rs:83 — **not checked anywhere in mempool**, I grepped), `no enabled policy trigger` (:162), the reward-amount/policy bound (:170), and the stateful eligibility checks — welcome already claimed (:186), daily already claimed (:196), invalid peer_did (:207/:215), partner already counted / weekly cap (:223/:232). Each is a claim that can pass admission and then fail apply → block halt, same mechanism as the one you just fixed. The `recipient_key_id` binding is the easy one to add here; the eligibility ones are the deeper problem (next comment).

lib-blockchain/src/transaction/validation.rs
```diff
@@ -2572,6 +2582,17 @@ impl<'a> StatefulTransactionValidator<'a> {
                 );
                 return Err(ValidationError::InvalidTransaction);
             }
+            // Apply hard-requires token_contract first. Checking rewards_module
+            // before contract previously admitted pure-AssetLaunch BUBL claims
+            // that then failed apply with "token contract not found" and halted
+            // consensus after BFT finality.
+            let Some(contract) = blockchain.get_token_contract(&data.token_id) else {
```
umwelt, line 2602: This token_contract check is correct and is the direct fix for GENESIS-3 — good. But note *why* mirroring can't be the whole answer for the eligibility rejects: they're **stateful and height-dependent**. A welcome/daily/partner claim can be eligible at mempool admission (not yet claimed at the current tip) and *ineligible* at apply because an intervening block claimed it. No amount of mirroring closes that race — admission and apply observe state at different heights — and the loser still halts the chain via apply's `Err` → `?`.

The durable fix is the one I raised on #2908 (option b): make a RewardClaim apply-failure a **graceful per-tx drop** (skip the tx, keep the block valid) instead of a fatal `Err` that halts. That removes the entire halt class in one place — token_contract, recipient binding, and every eligibility check — and makes the mempool mirroring a best-effort admission filter (reject early when you can) rather than a correctness-critical invariant that must stay byte-identical to apply forever. This PR is a solid stopgap for the specific incident; I'd want a follow-up issue for the graceful-drop change so we stop playing whack-a-mole with apply checks. Happy to be wrong if there's a reason apply-failure must stay fatal for RewardClaim specifically.

umwelt (commented):
Reviewed the whole PR (didn't split). Two parts stand out.

**Identity projection rebuild (good):** rebuilding sled projections from a block-scan on `load_from_store` is the right approach, and — crucially — it ships with `identity_projection_restart_tests` that assert the rebuilt state matches expected values (display_name, kyber key, `registered_at_height`, `identity_blocks`). That restart-equivalence test is exactly what guards against a restarted node diverging from a live one. No concerns there.

**RewardClaim admit safety:** this correctly fixes the GENESIS-3 H=1920 halt — requiring `token_contract` to exist at mempool admission, ordered before the rewards-module check, mirrors apply and stops pure-AssetLaunch BUBL claims from being admitted-then-halting. This is the token_contract half of exactly what I flagged on #2908, and it's done right.

But it doubles down on the fragile strategy, and I don't think it fully closes the halt class. Details inline. Short version: apply has ~13 reject conditions; this mirrors 5. Every unmirrored one is still a halt vector because RewardClaim apply-failure still halts the block (`apply_reward_claim` returns `Err` → `?` at `executor.rs:3544`; executor.rs isn't touched here). And the stateful ones can't be closed by mirroring at all.

lib-blockchain/src/transaction/validation.rs
umwelt: Addressed in 0425c87f:

1. **Mempool**: now also checks `recipient_key_id == key_id_from_did(owner_did)` and `amount > 0`. Comment updated so mempool is explicitly **best-effort** (not “identical to apply”).
2. **Durable fix (option b / #2908)**: apply rejections for RewardClaim now surface as `TxApplyError::RewardClaimDropped` and are **soft-dropped** in the executor (same path as `ReplayDropped`) — block stays valid; claim is skipped. Covers eligibility races, policy/timestamp/nonce-gap/liquidity, and the halt class you described.

Happy to take another pass if you want eligibility pre-checks kept as pure admission filters only with zero claim of parity.

lib-blockchain/src/transaction/validation.rs
umwelt: Agreed — height-dependent eligibility cannot be closed by mempool mirroring alone.

Landed the durable fix in this PR (0425c87f) rather than a follow-up: RewardClaim apply failures are soft-dropped (`RewardClaimDropped`) so intervening claims / race losers skip the tx and keep the block valid. Mempool token_contract + recipient binding remain early filters for the GENESIS-3 class of junk; they are no longer correctness-critical for consensus liveness.

No special reason apply-failure needed to stay fatal for RewardClaim — soft-drop matches option b from #2908.

umwelt:
## Review response (umwelt + Copilot)

Pushed **0425c87f** on this branch:

| Feedback | Change |
|---|---|
| Mempool only mirrors subset of apply checks | Added `recipient_key_id` binding + `amount > 0`; comment now says **best-effort**, not parity |
| Stateful eligibility → soft-drop (option b / #2908) | `RewardClaimDropped` + executor soft-drop on apply (not only pre-validate). Halt class closed for claim apply failures |
| Copilot: empty system sig on WalletRegistration | Empty-sig path binds dilithium `public_key` to `signature.public_key` and `wallet_id` to `blake3(public_key)` |

**Tests:** `reward_claim_mempool_tests` 16/16 (incl. soft-drop + recipient mismatch).

Ready for re-review. Stacked #2925 may need a rebase onto this tip.

Issue #2935, opened 2026-07-24
[EPIC] Authorization model: bootstrap → decentralized (principals, subject kinds, enforcement)


umwelt:
## Motif

The network is **decentralized by design** and **operator-bootstrapped today** (council, operator-run validators). Authorization work fragmented across incomplete ACL epics, an older DID permission track, and ungated wallet reads.

**Goal:** one model, one epic, enforced endpoints — execute end-to-end without reopening superseded issues.

**ADR (source of truth):** [`docs/arch/authorization-bootstrap-adr.md`](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/blob/development/docs/arch/authorization-bootstrap-adr.md) · PR [#2936](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2936)

---

## Is this epic enough to execute?

**Yes — for the authorization / API-principal work described here.**  
Execution package = **this issue (#2935) + ADR (+ PR #2936 when merged)**. You do **not** need the closed ACL/UAC issues open to implement.

| In scope (execute under #2935) | Out of scope (separate issues; link only) |
|--------------------------------|-------------------------------------------|
| Principal extraction & roles | AccessContract / SID ZK eligibility (#239, #293, …) |
| Handler enforcement (wallet privacy, admin gates) | Selective disclosure epic (#2303) |
| ScopedGrant, views, graph traverse | Chain contract caller auth (#1527) |
| Ops InfraAdmin clearance | Specialized wallet mint auth (#1556) |
| Tests/matrix/perf for **this** policy engine | Citizenship upgrade product path (#845) |
| | Non-Human identity types (#359) |
| | DAO launch readiness (#2934) |

When Phase 1–3 touch economy/chain call sites, open **child PRs** and optionally **link** #1527 / #1556 — do not reopen UnifiedAccessControl.

---

## Three axes (do not collapse)

| Axis | Question | Lives in |
|------|----------|----------|
| **A. Subject kind** | What is this DID? | `IdentityType` (Human, Organization, Device, Agent, Contract) |
| **B. Principal role** | What may this session do? | `lib-access-control::Role` + capabilities / ScopedGrant |
| **C. Economic authority** | What may this key sign on-chain? | Consensus (creator, governance, spend delegate, council txs, asset `dao_class` FP/NP) |

- **Citizen** ≈ Human DID + `Role::Citizen` (today’s default).
[...]

PR #2936, opened 2026-07-24
fix(rewards)+docs: BUBL AssetLaunch claims, wallet hydrate, auth ADR (#2933/#2935)

umwelt:
## Summary

**Not docs-only.** This PR lands:

1. **Runtime (consensus + API)** — GENESIS-3 BUBL / AssetLaunch `RewardClaim` when rewards module exists without a legacy `token_contracts` row; wallet identity hydrate from chain on multi-validator nodes.
2. **Architecture** — ADR for authorization bootstrap → decentralized model ([#2935](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2935)).

**Supersedes [#2933](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/pull/2933)** — same code path plus review fixes + ADR. Do not merge both (2933 is closed).

### Runtime changes
- Apply / mempool / settlement share **`authorize_reward_claim`** (module → spend delegate; else contract creator).
- Mempool admits module-only claims; apply-side test credits beneficiary balance.
- Wallet hydrate: fail closed on bad keys, Kyber when present, SOV balances from chain, O(1) DID map lookup.
- Status API: `claims_settleable` requires positive policy amount.

### ADR
- `docs/arch/authorization-bootstrap-adr.md` — axes A/B/C, phases, non-goals.
- Status: **Proposed (bootstrap)** until Phase 1 wallet-read enforcement lands.

## Test plan
- [x] `cargo test -p lib-blockchain --test reward_claim_mempool_tests` (17)
- [x] `cargo test -p lib-blockchain --lib transaction::reward_claim::tests`
- [x] `cargo check -p zhtp --lib`
- [ ] Deploy / claim smoke on GENESIS-3 (manual)

## Merge notes
- #2933 closed as superseded.
- Auth Phase 1–3 stack (#2937–#2939) still stacks on this branch for principal/wallet privacy work.

### DIFF
--- a/docs/arch/authorization-bootstrap-adr.md
+++ b/docs/arch/authorization-bootstrap-adr.md
@@ -1,0 +1,130 @@
+# ADR: Authorization model — bootstrap → decentralized
+
+**Status:** Proposed (bootstrap) — becomes Accepted when Phase 1 wallet-read enforcement lands under [#2935](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2935)  
+**Date:** 2026-07-24  
+**Epic:** [#2935](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2935)  
+**Supersedes:** Identity ACL epic #2272 and children; DID UnifiedAccessControl track #227–#230; network PoQ research #2111 (as prerequisite for API privacy)
+
+---
+
+## Context
+
+The network is **decentralized by design** but **operator-bootstrapped today**: council-run validators, a small on-chain council list, and one practical end-user principal (`Role::Citizen`). Access control work split across:
+
+- `lib-access-control` (principal × relation × domain × op) — incomplete enforcement
+- Older DID “UnifiedAccessControl” issues (#228–#230) — parallel, unused
+- Wallet reads intentionally ungated for mobile compat (#2299)
+- Roles `InfraAdmin` / `PolicyAdmin` / `Emergency` defined but never assigned
+
+**Urgent:** any authenticated DID can read any wallet balance/list/stats/history.
+
+---
+
+## Decision
+
+### 1. Three orthogonal axes (never collapse into one enum)
+
+| Axis | Question | Where it lives |
+|------|----------|----------------|
+| **A. Subject kind** | What is this DID? | `IdentityType` (Human, Organization, Device, Agent, Contract) |
+| **B. Principal role** | What may this session do at the API / mesh boundary? | `lib-access-control::Role` + capabilities / grants |
+| **C. Economic authority** | What may this key sign on-chain? | Consensus (creator, governance proof, spend delegate, council txs, `dao_class` on assets) |
+
+**DAO for-profit vs non-profit** is axis C (`dao_class` on `SovereignAsset`), not a DID kind and not an API role.
+
+**Citizenship / AccessLevel** (Visitor, FullCitizen, …) remains product/eligibility state; it does not replace axis B for request authorization.
+
+### 2. Bootstrap vs mature network
+
+| Concern | Bootstrap (now) | Mature |
+|---------|-----------------|--------|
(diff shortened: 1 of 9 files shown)


umwelt (commented):
Reviewed the whole PR (ADR + code). Good news up front, then two things to resolve.

## This resolves my #2933 blocker — verified

On #2933 I flagged that the one test touching the rewards-module-without-contract path was vacuous (it asserted *rejection*, and passed only because the claim died at a generic `Invalid transaction structure` check before the auth logic). Here that test is renamed `rewards_module_without_token_contract_admitted_to_mempool` and now asserts **admission** (`get_pending_transactions().len() == 1`). I checked out this branch and ran it: it passes, and the **full mempool suite is 16/16**. There's also an apply-side soft-drop test (`executor_soft_drops_duplicate_welcome_claim_after_first_applied`) citing the #2924/#2908 reviews. So the BUBL fix now has a real positive test proving it works end-to-end at both mempool and apply. This is exactly what I asked for — thanks for closing it properly.

## Correctness carries over (no halt regression)

The reward-claim widening (mempool / apply / settlement all branch `rewards-module → spend-delegate; else contract + creator`) is the same symmetric change I verified on #2933: admission and apply agree, and `reject_claim` is soft-drop, so it does not reopen the GENESIS-3 halt. Wallet hydrate is the same sound read-only, add-only, chain-authoritative reconstruction (minor: O(n) registry scan per foreign-DID read on g2/g3 — a key_id index later). The ADR's "reads must not 403 → return empty/filtered" rule (§4) and the InfraAdmin ops model (§5 Phase 2) are consistent with the code here, and with the maintenance-banner work.

## Two things to resolve

**1. #2936 and #2933 are parallel branches editing identical code — they cannot both merge.** I checked: #2936 does not contain #2933; both modify `reward_claim.rs`, `validation.rs`, `wallet/mod.rs`, `settlement.rs`, `rewards/mod.rs`, and the same tests, independently off `development`. Whichever merges second conflicts. #2936 is strictly the better of the two (it has the working admit test that #2933 lacks, plus the ADR), so the clean move is to **close #2933 in favor of this** — but that decision should be explicit, not left to a merge-time conflict. Don't merge both.

**2. Scope/title.** This is titled `docs(arch)` and the test plan says "Doc review only (no runtime change)," but it carries ~190 lines of **consensus-path** change (reward-claim authorization in apply + mempool + settlement, wallet balance hydration). That's the opposite of no-runtime-change, and it's the kind of thing a reviewer skimming a "docs" PR would wave through. Retitle to reflect that it's the reward-claim/wallet-hydrate fix *plus* the ADR, and fix the test plan — the code here needs the same scrutiny #2933 got (it got it here, but only because I recognized the files).

Minor: the ADR is `Status: Accepted` while landing alongside its own first partial implementation; `Proposed`/`Accepted (bootstrap)` would be more honest until Phase 1 enforcement lands. Non-blocking.

umwelt:
## Addressing review (Copilot + @umwelt)

### Parallel with #2933
- **Resolved by consolidating into this PR.** Branch reset to include full #2933 tip (review fixes + `authorize_reward_claim`) **plus** the ADR.
- Will **close #2933 as superseded** so only one merge path remains.

### Scope / title
- Retitled and rewritten PR body: **runtime + ADR**, not “docs only.”
- Test plan lists consensus/API verification (not doc-only).

### ADR status
- Changed to **Proposed (bootstrap)** until Phase 1 enforcement lands (non-blocking note).

### Copilot hydrate items
Already fixed on the #2933 line included here:
- Fail closed on bad Dilithium length + Kyber when present
- Byte owner compare + SOV `token_balance` + `calculate_total_balance`
- **New:** O(1) DID key lookup on identity registry snapshot

Sonar duplication % on this PR may still flag overlapping new code; the authorize helper is the intentional dedup of the three ladders.

PR #2937, opened 2026-07-24
feat(auth): Phase 1 wallet read privacy + principal hardening (#2935)

umwelt:
## Summary

Implements **#2935 Phase 1** (stacked on ADR PR #2936).

- **Wallet read soft-privacy**: list / balance / statistics / transactions return 200 + empty/zero for non-self, non-Council/InfraAdmin callers (no balance leak).
- **Principal hardening**: bare Bearer no longer elevates to `Role::Citizen`; client `x-node-type` no longer yields `Role::Node`.
- **Helpers**: `may_read_wallet_subject`, `wallet_read_privacy_enforced` (`ZHTP_ENFORCE_WALLET_READ_PRIVACY`, default on), early `ZHTP_INFRA_ADMIN_DIDS` role assignment.
- Mesh identity principal path aligned (no spoof elevation).

## Stack

1. #2936 ADR (base)
2. **This PR** — Phase 1
3. Phase 2 → Phase 3 (follow-on)

## Test plan

- [x] `cargo test -p zhtp --lib principal_extraction_tests`
- [x] `cargo test -p zhtp --lib api::principal::tests`
- [x] `cargo check -p zhtp --lib`
- [ ] Reviewer: self-DID still sees own wallets; cross-identity empty; Council full; send still owner-gated 403

## Do not merge until

Morning review of the full epic stack. **Do not merge this PR alone without ADR base.**

Implements #2935 Phase 1.

### DIFF
--- a/zhtp/src/api/handlers/identity/mod.rs
+++ b/zhtp/src/api/handlers/identity/mod.rs
@@ -2591,28 +2591,28 @@ mod principal_extraction_tests {
     }
 
     #[test]
-    fn test_extract_principal_bearer_returns_citizen() {
+    fn test_extract_principal_bearer_without_did_is_public() {
+        // #2935 Phase 1: bare Bearer must not elevate to Citizen.
         let request = request_with_header("authorization", "Bearer token123");
         let principal = extract_principal_from_request(&request);
-        assert_eq!(principal.role, Role::Citizen);
-        assert_eq!(principal.did, "did:zhtp:session");
+        assert_eq!(principal.role, Role::Public);
+        assert_eq!(principal.did, "did:zhtp:public");
     }
 
     #[test]
-    fn test_extract_principal_node_type_returns_node() {
+    fn test_extract_principal_node_type_header_not_trusted() {
+        // #2935 Phase 1: client-declared x-node-type must not yield Role::Node.
         let request = request_with_header("x-node-type", "validator");
         let principal = extract_principal_from_request(&request);
-        assert_eq!(principal.role, Role::Node);
-        assert_eq!(principal.did, "did:zhtp:node");
-        assert_eq!(principal.node_type, NodeType::Validator);
+        assert_eq!(principal.role, Role::Public);
+        assert_eq!(principal.did, "did:zhtp:public");
     }
 
     #[test]
-    fn test_extract_principal_relay_node_type() {
+    fn test_extract_principal_relay_node_type_header_not_trusted() {
         let request = request_with_header("x-node-type", "relay");
         let principal = extract_principal_from_request(&request);
-        assert_eq!(principal.role, Role::Node);
-        assert_eq!(principal.did, "did:zhtp:node");
-        assert_eq!(principal.node_type, NodeType::Relay);
+        assert_eq!(principal.role, Role::Public);
+        assert_eq!(principal.did, "did:zhtp:public");
     }
 }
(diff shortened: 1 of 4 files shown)


umwelt (commented):
Security review — I traced the whole authorization path end to end, and this is solid. It correctly implements the ADR Phase 1 without the bypasses this kind of change usually ships with. Verified below, then four minor notes (none blocking).

## Verified secure

1. **Principal is un-spoofable.** `principal.did` and `role` derive **only** from `request.requester`, which is set server-side from the authenticated session (`quic_handler.rs:754` and `router.rs:356`, both from `session.peer_did()`), never from a client header. Unauthenticated → `Public`.
2. **Both elevation paths are killed — and in both principal derivations.** `x-node-type` → `Node` and bare-Bearer → `Citizen` are removed in `principal.rs`, **and** the separate mesh path in `server/mesh/identity_api.rs` (which had its own `default to Node` + `x-node-type` elevation) is aligned to Public-until-authenticated. That second one matters — a mesh-routed request was the obvious bypass, and it's closed.
3. **Role assignment fails closed.** `resolve_role_for_dids` grants `Council` only on an explicit `is_council_member_blocking(did) == Some(true)`; `None`/error/cold-cache → falls through to `Citizen`. No fail-open to a privileged role.
4. **Self-check is exact.** `identity_id_matches_caller` compares hex for equality after stripping `did:zhtp:` — no substring/prefix laxity.
5. **Coverage is complete and consistent.** All four read routes (list/balance/statistics/transactions) gate on `may_read_wallet_subject` (soft-privacy: 200 + empty/zero); all four write routes (send/cross-transfer/stake/unstake) hard-**403** via `identity_id_matches_caller || Council`. Every authz decision in `wallet/mod.rs` routes through the two canonical helpers — no hand-rolled copies to drift. Default is fail-closed (Public → deny).

Tests pass (5/5), and the env rollback (`ZHTP_ENFORCE_WALLET_READ_PRIVACY`) is a sensible kill-switch.

## Minor (non-blocking)

**a. Case-sensitivity in the self-check.** `identity_id_matches_caller` does `caller_hex == request_hex` on raw strings. If the URL path carries upper-case hex and the authenticated DID is lower-case (or vice-versa), a caller is denied *their own* wallet. It's fail-closed (a UX bug, not a leak), but normalize both sides with `eq_ignore_ascii_case` to avoid it.

**b. `ZHTP_INFRA_ADMIN_DIDS` is env-based privilege.** Per the ADR this is the intended Phase-1 bootstrap (Phase 2 moves it on-chain), so fine — but an InfraAdmin gets *full wallet-read across every user*. A typo or leaked env value silently grants cross-user balance visibility. Worth calling out operationally as a secret-grade config, and logging at startup which DIDs were granted InfraAdmin so a mistake is visible.

**c. Cold-cache council flap.** `is_council_member_blocking` returning `None` before the council cache is warm means a real council member is treated as `Citizen` right after restart — denied their council-level reads until warm. Availability, not security. Confirm the cache is warmed before the API starts serving, or council reads will flap on every restart.

**d. Dead bindings.** `let _ignored_spoofable_node_type` / `_ignored_unbound_bearer` (principal.rs) are reads whose values are discarded. They document intent nicely but will trip `unused` lints; a short comment + dropping the binding, or `#[allow]`, reads cleaner.

One thing to confirm (not a defect): the soft-privacy empty body for a denied cross-identity read is, by design, indistinguishable from a legit empty wallet. That's the ADR's chosen behavior (§4 'reads must not 403 → empty/filtered'), so it's correct — just make sure the mobile client doesn't treat an empty balance as 'account error' when viewing another user.

PR #2938, opened 2026-07-24
feat(auth): Phase 2 InfraAdmin ops, audit logs, CI matrix (#2935)

umwelt:
## Summary

Implements **#2935 Phase 2** (stacked on Phase 1).

- **`is_ops_elevated`**: Council | InfraAdmin (System removed in Phase 3 stack tip).
- **Ops gates**: halt-consensus, chain export/import, wallet provision.
- **Audit**: `log_access_decision` denies at INFO (`target: access_control`).
- **CI matrix + perf**: `lib-access-control/src/matrix_tests.rs` — role/domain/op expectations + `check_access` &lt;1ms average budget.

## Stack

Phase 1 → **this** → Phase 3

## Test plan

- [x] `cargo test -p lib-access-control --lib` (18+ tests including matrix + perf)
- [ ] Reviewer: set `ZHTP_INFRA_ADMIN_DIDS` and confirm ops endpoints; Citizen still 403 on halt/export

Implements #2935 Phase 2.

### DIFF
--- a/docs/arch/authorization-bootstrap-adr.md
+++ b/docs/arch/authorization-bootstrap-adr.md
@@ -83,6 +83,15 @@ Device QUIC key → canonical DID binding remains required for owner-gated endpo
 3. Structured access-decision logging (denies INFO).
 4. Subset of E2E matrix in CI.
 
+**Ops note — `ZHTP_INFRA_ADMIN_DIDS` is consensus-critical.** Comma-separated
+DIDs that receive `Role::InfraAdmin`. Blast radius after Phase 2: halt consensus
+(network DoS), full chain export (exfiltration), chain import, wallet provision,
+and cross-user wallet read (Phase 1). Treat the variable as secret-grade
+allowlist config; log the granted DIDs at node startup (`access_control` target).
+Bootstrap keeps **export under Council | InfraAdmin** so ops can recover nodes
+without Council keys; Phase 3 may tighten export to Council-only or ScopedGrant
+if product policy requires a higher bar.
+
 #### Phase 3 — Mature access
 
 1. **ScopedGrant**: council-issued, scoped, expiring; remove `Role::System` bypass.
--- a/lib-access-control/src/lib.rs
+++ b/lib-access-control/src/lib.rs
@@ -15,6 +15,9 @@ pub mod policy;
 pub mod principal;
 pub mod types;
 
+#[cfg(test)]
+mod matrix_tests;
+
 pub use decision::{AccessDecision, ReasonCode};
 pub use policy::AccessPolicy;
 pub use principal::SecurityPrincipal;
--- a/lib-access-control/src/matrix_tests.rs
+++ b/lib-access-control/src/matrix_tests.rs
@@ -1,0 +1,335 @@
+//! CI policy matrix: Role × SubjectRelation × AccessDomain × AccessOperation.
+//!
+//! Snapshot expectations so policy changes are intentional (#2935 Phase 2).
+//! Default DENY for unmatched combinations is covered by the matrix rows that
+//! assert denial.
+
+#![cfg(test)]
+
+use crate::policy::AccessPolicy;
+use crate::principal::SecurityPrincipal;
+use crate::types::{
+    AccessDomain, AccessOperation, Capability, Role, SubjectRelation,
+};
+use lib_types::NodeType;
+
+fn p(role: Role) -> SecurityPrincipal {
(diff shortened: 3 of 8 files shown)


umwelt (commented):
Security review — Phase 2 ops gating. Clean and correct; verified below. One operational point I'd want acknowledged before merge (not a code defect), plus small notes.

## Verified secure

1. **`is_ops_elevated` fails closed** — `matches!(role, Council | InfraAdmin | System)`; every other role (Citizen/Public/Device/…) → false. Since Phase 1's role assignment is un-spoofable and defaults to Citizen, nothing reaches these gates without a real Council/InfraAdmin DID.
2. **Gates are cleanly replaced, not double-gated.** halt-consensus went from `if role != Council` → `if !is_ops_elevated`; the export/import path went from `Council || System` → `is_ops_elevated`. All four ops surfaces (halt / export / import / provision) route through the one helper — no leftover inline check that would either shadow the new gate or leave a gap.
3. **`System` in `is_ops_elevated` is inert for these gates.** `extract_principal_from_request` only ever yields Council/InfraAdmin/Citizen (via `resolve_role_for_dids`) or Public — never System. So a network request can't reach the System arm; it's dead-but-harmless until Phase 3 removes it. Worth a one-line comment saying so, so nobody thinks a request can arrive as System.
4. **Audit log leaks nothing sensitive** — `log_access_decision` records DID (public), role, subject, domain, op, reason. No keys, balances, or bodies. Denies at INFO / allows at DEBUG is a sensible split.
5. **Perf test isn't flaky** — 1ms *average* over many iterations is a huge margin for a `check_access` that should run in microseconds; it's a pathological-regression guard, not a tight benchmark. Fine.

## Important — blast radius of `ZHTP_INFRA_ADMIN_DIDS` just escalated (operational, not a bug)

On Phase 1 I flagged that env-configured InfraAdmin grants full cross-user wallet read, so treat the var as secret-grade. Phase 2 widens InfraAdmin to **halt consensus, export the entire chain, import chain state, and provision wallets**. So a typo or leaked `ZHTP_INFRA_ADMIN_DIDS` now means network-wide DoS (halt) and full-chain exfiltration (export) — not just balance visibility. This is the intended ADR widening, so no code change needed, but two asks: (a) **log the granted InfraAdmin DIDs at startup** (I recommended this on Phase 1; it's now important, not nice-to-have — an operator must be able to see who holds halt/export rights); (b) note in the ops runbook that this env var is consensus-critical. Chain export in particular deserves a second thought — is `Council | InfraAdmin` the right bar for dumping the whole chain, or should export stay Council-only even in bootstrap?

## Minor

- My own maintenance-banner PR (#2929) hand-rolls this exact gate (`matches!(role, Council | InfraAdmin)` in a local `require_operator`). Once this lands, #2929 should adopt `is_ops_elevated` for consistency — flagging so the two don't drift. (Note the semantic difference: my helper excludes System; `is_ops_elevated` includes it — inert per point 3, but worth being deliberate about.)
- The stack ordering note (`Phase 1 → this → Phase 3`, base `auth/2935-phase1-wallet-privacy`) means this can't merge before #2937, which can't merge before the #2936 ADR — and #2936 still has the parallel-with-#2933 / mis-titled-as-docs issue open. The whole chain is gated on sorting that out.

umwelt:
## Addressing review (@umwelt)

### Verified secure
Acknowledged — no code change needed for the five verified points. Added the one-line note on `is_ops_elevated`: `Role::System` is **inert for HTTP** (`extract_principal_from_request` never yields System).

### Blast radius of `ZHTP_INFRA_ADMIN_DIDS`
1. **Startup log** — `log_infra_admin_config_at_startup()` runs from `main` after tracing init (`target: access_control`). Empty → INFO that no env InfraAdmin is configured; non-empty → count + DID list (public identifiers).
2. **Ops/ADR note** — Phase 2 section of `docs/arch/authorization-bootstrap-adr.md` now states the var is **consensus-critical** (halt/export/import/provision + wallet read) and must be treated as secret-grade allowlist config.
3. **Export bar** — kept **Council | InfraAdmin** for bootstrap (ops recovery without Council keys). Documented that Phase 3 may tighten export to Council-only or ScopedGrant if product policy requires it.

### Minor
- **#2929** adopting `is_ops_elevated`: noted; out of scope for this PR (will align when that lands on the same stack tip).
- Stack / #2936 parallel: already consolidated (#2933 closed; #2936 is BUBL+hydrate+ADR).

Pushed on `auth/2935-phase2-ops-matrix` (`fix(auth): address #2938 review — log InfraAdmin at startup`). Phase 3 restacked on this tip.

PR #2939, opened 2026-07-24
feat(auth): Phase 3 ScopedGrant, no System god-mode, product caps (#2935)

umwelt:
## Summary

Implements **#2935 Phase 3** (stack tip).

- **`ScopedGrant`**: time/use bounded domain+op elevation on `SecurityPrincipal.grants`.
- **`Role::System` no longer god-mode**: self-only + grants; external deny by default.
- **Product capabilities**: `DaoParticipate`, `DexRead`, `DexTrade`, `Web4Publish`, `EconomicAdmin` under existing policy (not a second UAC engine).
- **`check_graph_edge`**: per-edge Traverse helper.
- Ops elevation: Council | InfraAdmin | NodeGraph Traverse grant (no System).

## Stack

ADR #2936 → Phase 1 → Phase 2 → **this**

## Test plan

- [x] `cargo test -p lib-access-control --lib` (22 tests)
- [x] `cargo check -p lib-identity -p lib-governance -p zhtp --lib`
- [ ] Reviewer: grant elevation path; confirm System cannot export chain without InfraAdmin/Council

## Note

Full grant **issuance API** (council-only POST) and wiring grants into live sessions is intentionally thin here — types + policy evaluation land first for review. Follow-up can attach grants at login from chain/config.

Implements #2935 Phase 3.

### DIFF
--- a/lib-access-control/src/grant.rs
+++ b/lib-access-control/src/grant.rs
@@ -2,14 +2,37 @@
 //!
 //! Replaces ad-hoc `Role::System` god-mode and overuse of Council for ops.
 //! Issuance is council-only at the API layer; this module is pure evaluation.
+//!
+//! # What is enforceable without a grant store
+//!
+//! Grants are attached to a per-request `SecurityPrincipal` and are **not**
+//! mutated across requests. Evaluation therefore only enforces:
+//! - domain / operation coverage
+//! - absolute time expiry (`expires_at_unix`)
+//! - sticky revoke (`consumed` already true when loaded)
+//!
+//! **Use limits (`max_uses`) are intentionally absent.** A counter on a
+//! throwaway principal cannot enforce "use once" across requests. When a
+//! server-side grant store exists (keyed by grant id, atomically decremented
+//! on successful use), reintroduce use limits there — not on this type.
+//!
+//! # Hot-path bound
+//!
+//! `MAX_GRANTS_PER_PRINCIPAL` caps grants attached to a principal so
+//! `grants_allow` stays O(1)-bounded on the authz path. Future council
+//! issuance must refuse to exceed this.
 
 use crate::types::{AccessDomain, AccessOperation, Did};
 use serde::{Deserialize, Serialize};
 
+/// Upper bound on grants attached to a single principal (authz hot path).
+/// Council-only issuance APIs must reject or truncate beyond this.
+pub const MAX_GRANTS_PER_PRINCIPAL: usize = 32;
+
 /// A council-issued (or future-governance-issued) scoped grant held by a principal.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ScopedGrant {
-    /// Stable grant id (hex or uuid string).
+    /// Stable grant id (hex or uuid string). Used as key when a store lands.
     pub id: String,
     /// DID that may exercise this grant.
     pub grantee_did: Did,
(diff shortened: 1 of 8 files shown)


lib-access-control/src/grant.rs
```diff
@@ -0,0 +1,169 @@
...
+    domain: AccessDomain,
+    op: AccessOperation,
+    now_unix: u64,
+) -> bool {
+    grants.iter().any(|g| {
+        g.grantee_did == grantee_did
+            && g.is_active(now_unix)
+            && g.covers(domain, op)
+    })
+}
```
umwelt, line 117: `record_use` is the only thing that advances `uses`, and it's called from nowhere except the test module below. Combined with `is_active`'s `uses >= max` check, that means `max_uses` never triggers — `uses` is always 0. And because grants sit on a per-request `SecurityPrincipal` that isn't persisted, calling `record_use` from the enforcement path wouldn't help either (the increment dies with the request). A use-limited grant therefore needs a persistent, atomically-decremented server-side store keyed by grant id; a counter on the principal can't enforce it. Until that exists, `max_uses` is a no-op — either build the store or remove the field so it isn't mistaken for an enforced limit.

umwelt (commented):
Security review — Phase 3 (ScopedGrant + de-god-moding System). The design is mostly right and it is **not exploitable today**, but there's a latent design bug in `max_uses` that will silently fail to enforce once grants are wired in — and grants feed the ops gate, so I'd fix the model before building the wiring on top of it.

## Not exploitable today (verified)

`principal.grants` is never populated anywhere in `zhtp` (I grepped) — `extract_principal_from_request` builds principals with no grants. So both grant consumers are currently dead paths: `is_ops_elevated`'s `grant_allows(NodeGraph, Traverse, now)` and the policy engine's early grant check. Consistent with the PR note that wiring is intentionally thin. The clock (`now`) is `SystemTime::now()` server-side — not spoofable. Good.

## System de-god-moding — correct

`policy.rs` restricts `System` to the `Self_` relation + grants (no blanket allow), `is_ops_elevated` no longer accepts `System`, and sensitive/private reads are denied for System (`policy.rs:426`). This closes the god-mode path cleanly. Product capabilities go through the existing decision engine rather than a second UAC — also the right call.

## Latent bug — `max_uses` is unenforceable as designed (fix before wiring)

`is_active` checks `self.uses >= max`, but `self.uses` is only ever incremented by `record_use`, and **`record_use` is called from nothing but a unit test**. So `uses` stays `0` forever and the use-limit never trips. Worse, it can't be fixed just by calling `record_use` from the enforcement path: grants live on `SecurityPrincipal.grants`, which is rebuilt per request and never persisted, so a `record_use` mutation lands on a throwaway object and is gone by the next request.

Once grant-wiring lands (attach grants at login, per your follow-up), the effect is: **any `max_uses: N` grant is usable an unlimited number of times until `expires_at_unix`** (or forever if no expiry). And this isn't confined to a low-value scope — grants authorize (a) ops elevation via `is_ops_elevated` → halt-consensus / chain export / import / provision, and (b) cross-identity access in the policy engine (`policy.rs:53` "grants can authorize cross-identity ops"). So a grant issued as "export the chain **once**" or "read Alice's wallet **once**" silently becomes unlimited-until-expiry.

Fix the model now, while it's cheap: use-limited grants need a **server-side persistent grant store keyed by grant id, decremented atomically on each successful use** — not a `uses` counter on a per-request principal. If that store isn't in scope for this PR, drop `max_uses` from the `ScopedGrant` type until it exists, so no one issues a use-limited grant believing it's enforced. As-is, the field is a security control that looks real and does nothing.

## Minor

- `grants_allow` linear-scans all grants per check with no cap. Harmless until issuance exists, but the (future) council-only issuance API should bound grants-per-principal so this can't become an O(n) DoS on the authz hot path.
- Consider a test that asserts a `max_uses(1)` grant is rejected on the *second* `grants_allow` after a real use — it would fail today, which is precisely the point.

umwelt:
## Addressing review (@umwelt)

### Latent bug — `max_uses` unenforceable
Agreed. Fixed by **removing use-count fields** from `ScopedGrant` rather than shipping a fake control:

- Dropped `max_uses`, `uses`, `with_max_uses`, `record_use`.
- Evaluation remains: domain/op coverage + `expires_at_unix` + sticky `consumed` when already set **before** attach (store/issuer revoke).
- Module docs + ADR Phase 3 spell out: use limits return only with a **server-side grant store** (grant id key, atomic decrement on success). Not inventable on a throwaway principal.

### Minor — grants linear scan
- Added `MAX_GRANTS_PER_PRINCIPAL` (32); `SecurityPrincipal::with_grants` truncates.
- Future council issuance API must refuse beyond this bound.

### Not exploitable / System de-god-moding
Acknowledged — no further change.

Pushed on `auth/2935-phase3-scoped-grant` (`fix(auth): drop unenforceable ScopedGrant max_uses`). `cargo test -p lib-access-control --lib` — 23 passed.

PR #2940, opened 2026-07-25
fix(ipc): replace Unix domain sockets with platform-agnostic transport

### LINKED ISSUE #2915: Windows: IPC transport (Unix socket → TCP loopback)
## Context

Split out of PR #2914 (domain-update signature prefix). That PR originally included Windows IPC changes that do not belong with crypto/domain-signing work.

## Summary

- `lib-blockchain/src/ipc/client.rs` / `server.rs`
- On Unix: keep Unix domain sockets
- On Windows: `TcpListener` on `127.0.0.1:0`, write assigned port to path file; client reads port and connects via `TcpStream`
- Public API should remain unchanged

## Source

Original commit on the old #2914 branch: `0c4ecddb` (also on historical `fix/Windows-Stack-Overflow` lineage).

## Acceptance

- [ ] Compiles and works on Windows targets
- [ ] Unix path unchanged
- [ ] Separate PR from domain-signing / consensus work

PeterGRutherford:
Summary
Replaced Unix domain sockets (UnixListener/UnixStream) with platform-conditional type aliases so the IPC layer compiles on Windows. On Windows, the server binds a TcpListener to 127.0.0.1:0 and writes the assigned port to the path file; the client reads the port from the file and connects via TcpStream. Unix behavior is unchanged.

Linked issues
Refs #2914 (split out Windows IPC changes from domain-update signature prefix PR).

Architecture compliance
n/a — standalone platform-compatibility fix, not part of an epic-tracked design doc.

Test plan
 cargo check -p lib-blockchain passes on Windows
 Manual verification: start blockchain engine on Windows, verify IPC client connects and queries succeed
Risk and reversibility
Low risk. Unix codepath is unchanged (same #[cfg(unix)] code as before). Windows codepath is new but isolated behind #[cfg(not(unix))]. No coordinated network restart needed

### DIFF
--- a/lib-blockchain/src/ipc/client.rs
+++ b/lib-blockchain/src/ipc/client.rs
@@ -1,14 +1,18 @@
-//! IPC client — connects to the blockchain engine via Unix domain socket.
+//! IPC client — connects to the blockchain engine via platform-native transport.
+//!
+//! - **Unix**: Unix domain socket
+//! - **Windows**: TCP loopback (port discovered from path file)
 //!
 //! Provides owned-type query methods equivalent to `BlockchainQuery`.
 //! Can be used as a drop-in replacement for `Arc<RwLock<Blockchain>>` reads.
 
+use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::UnixStream;
 use tokio::sync::Mutex;
 use tracing::warn;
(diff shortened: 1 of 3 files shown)


umwelt (approved):
Verified: addresses the scope-split we requested on #2914 (focused, platform-additive). Unix path unchanged (cfg-gated); combined build of all four Windows PRs is green on Linux with no new errors/warnings. No regression.

PR #2942, opened 2026-07-25
fix(lib-network): add missing BLE_MESH_SERVICE_UUID and Mutex imports for windows

### LINKED ISSUE #2918: Windows: fix BLE_MESH_SERVICE_UUID / Mutex imports (bluetooth, wifi_direct)
## Context

Split out of PR #2914. Small Windows compile fixes that rode along with domain-update signing:

- `lib-network/src/protocols/bluetooth/discovery.rs` — import `BLE_MESH_SERVICE_UUID`
- `lib-network/src/protocols/wifi_direct.rs` — import `std::sync::Mutex` for Windows device-watcher blocks

## Acceptance

- [ ] Windows `cargo check` for those modules
- [ ] Separate tiny PR (or fold into broader Windows support PR)

PeterGRutherford:
Add missing `BLE_MESH_SERVICE_UUID` and `std::sync::Mutex` imports in `lib-network` that caused compile failures on Windows. These code paths are `#[cfg(target_os = "windows")]`-gated, so the errors were masked on Linux/macOS.

## Linked issues

Refs #2918. Split from #2914.

## Architecture compliance

n/a — import-only fix, no logic change.

## Test plan

- [x] `cargo check -p lib-network` on Windows now passes for the two affected modules (Bluetooth discovery and WiFi Direct device-watcher blocks). The remaining `libc::kill` error in `zhtp/src/api/handlers/network/mod.rs` is unrelated and will be handled separately.

## Risk and reversibility

Zero. Two `use` statements added; no runtime behaviour change. Can be reverted by dropping the commit.

### DIFF
--- a/lib-network/src/protocols/bluetooth/discovery.rs
+++ b/lib-network/src/protocols/bluetooth/discovery.rs
@@ -5,6 +5,8 @@ use std::sync::Arc;
 use tokio::sync::RwLock;
 use tracing::{info, warn};
 
+use crate::constants::BLE_MESH_SERVICE_UUID;
+
 use super::device;
 use super::device::{BleDevice, MeshPeer};
 use super::BluetoothMeshProtocol;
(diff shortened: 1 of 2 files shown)


umwelt (approved):
Verified: addresses the scope-split we requested on #2914 (focused, platform-additive). Unix path unchanged (cfg-gated); combined build of all four Windows PRs is green on Linux with no new errors/warnings. No regression.

PR #2946, opened 2026-07-25
docs(arch): dual-auth DID+grant model and implementation plan

umwelt:
## Summary
- ADR for one-DID dual-auth elevation (DID ∧ grant proof)
- Implementation plan (phases A–E), two issuance rails (council + protocol), claim lifecycle
- Pointer from bootstrap authorization ADR

## Stack
This is **PR 1 of N** (base of dual-auth stack). Do not merge until the stack is reviewed.

## Test plan
- [ ] Docs-only review of ADR + plan vs product decisions (one DID, grants as meta-credentials, role purity)
- [ ] Confirm bootstrap ADR pointer is correct

### DIFF
--- a/docs/arch/authorization-bootstrap-adr.md
+++ b/docs/arch/authorization-bootstrap-adr.md
@@ -108,6 +108,12 @@ per-request principal and are not mutated across requests, so **use-count limits
 grants-per-principal (`MAX_GRANTS_PER_PRINCIPAL`) so `grants_allow` cannot become
 an O(n) DoS on the authz hot path.
 
+**Concept target (not fully implemented):** one subject DID + **dual-auth grants**
+as meta-credentials — elevated power requires grant key proof separate from
+personal DID unlock; wallet stores DID keys only; grant keys are external /
+on-demand (optionally cold). See
+[did-grant-dual-auth-adr.md](./did-grant-dual-auth-adr.md).
+
 ### 6. Non-goals
 
 - Replacing `lib-access-control` with a second UnifiedAccessControl Permission mega-enum.
@@ -144,4 +150,5 @@ an O(n) DoS on the authz hot path.
 - `lib-access-control` (`Role`, `AccessDomain`, `AccessPolicy`)
 - `zhtp/src/api/principal.rs` (`extract_principal_from_request`)
 - Wallet ungated reads: historical #2299
+- [did-grant-dual-auth-adr.md](./did-grant-dual-auth-adr.md) — one DID, grant meta-credentials, separate unlock / cold grant keys
 - Sovereign assets / dao_class: `docs/arch/sovereign-asset.md`, `docs/arch/dao-launch-decision-register.md`
--- a/docs/arch/did-grant-dual-auth-adr.md
+++ b/docs/arch/did-grant-dual-auth-adr.md
@@ -1,0 +1,221 @@
+# ADR: One DID, dual-auth grants (identity vs elevation)
+
+**Status:** Proposed  
+**Date:** 2026-07-24  
+**Epic:** [#2935](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2935)  
+**Related:** [authorization-bootstrap-adr.md](./authorization-bootstrap-adr.md)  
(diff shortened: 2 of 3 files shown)


umwelt (commented):
Docs review of the dual-auth ADR + plan. **The design is strong — I'd accept the model** — and it correctly internalizes the enforcement gaps from the earlier auth stack. But the drift inventory documents a live, exploitable hole that should not wait for the phased rollout. Details below.

## The model is right

- **§3 Dual-auth (eligibility ≠ exercise)** is exactly the fix for the god-DID / blanket-role problems I flagged on #2937/#2938: being on the Council list no longer authorizes halt/export/cross-wallet — you must present an *active, authenticated grant* for that domain/op. No more fat roles.
- **§10 carries forward my #2939 finding verbatim** — "Do not reintroduce `max_uses` without a store" — and Phase B defers use-count until the persistent grant store exists. That's the correct resolution of the exact bug (per-request principal can't enforce use-limits).
- **§4 key-custody separation** (grant secret keys never in `user_private_key.json`, separate from DID keys) is sound, and the risk table honestly flags "no cold-grant mode: High."

## Act now, don't wait for Phase A: `node/shutdown` is unauthenticated today

The drift inventory lists "Node shutdown | Today allow: **Open**," and Phase A item 1 is "gate node shutdown with `is_ops_elevated` (or remove)." I verified on `development`: `handle_node_shutdown` has no principal/role check at all — `POST /api/v1/node/shutdown` is a **remote unauthenticated DoS** right now. This PR says "do not merge until the stack is reviewed," and Phase A is a future PR behind it — so a known catastrophic open endpoint is queued behind a multi-phase ADR. **Pull the shutdown gate out as an immediate standalone one-liner** (and while there, confirm chain export/import are gated — #2938 gated them to Council|InfraAdmin, which is at least not open). Don't let the documentation of a live hole substitute for closing it.

## Two plan risks worth resolving in the ADR before building on it

1. **Grant store: sled vs chain is left as "product choice" (§ Phase B) — but for ops/audit grants that choice is consensus-affecting.** A sled-local grant store means a grant claimed or **revoked** on g1 is invisible to g2/g3, so ops-grant *enforcement and revocation would be inconsistent across validators* — exactly the wrong property for cluster-wide powers like halt/export. Recommend the ADR commit to **chain-backed (or gossip-replicated) for ops-class grants**, reserving sled-local only for node-scoped protocol grants. Leave it open and someone will ship per-node state for a network-wide capability.

2. **The dual-path window (A→D) is a window of known-exploitable roles.** Phase B's exit is explicitly "fat roles still work (dual path)" until the Phase D cutover. That's fine *only if* Phase A seals the highest-risk drift, not just shutdown. Make Phase A's "seal holes" explicitly harden export/import and the Council-send-as-other write bypass (drift inventory: "Council send/stake as other | Role bypass"), so the migration window isn't months of live over-broad ops.

**Minor:** the lifecycle (§5.1) doesn't state revocation propagation latency — for a revoked *ops* grant, how fast every node stops honoring it should be specified; it ties directly to the store-consistency point above.

Net: accept the design; fix `node/shutdown` immediately out-of-band; commit the ops-grant store to chain-backed before Phase B.

umwelt:
Added hard ordering gate: **production Dilithium `GrantSignatureVerifier` (Phase B3) must land before any Phase D fat-role cut-over** — otherwise ops become DID-only or unreachable. See updated plan Phase D + §10–11 + PR DAG B3 node.

umwelt:
Addressed review feedback in latest commits:

**umwelt**
- Shutdown gate already out-of-band as #2950 (surgical, not blocked on docs).
- ADR + plan now **commit ops/audit/vote grants to chain-backed or gossip-replicated store**; sled/process-local only for node-scoped `NodeOperate`. Revocation latency called out.
- Phase A expanded: confirm export/import stay gated; track Council write-as-other in dual-path window; operator release-note for InfraAdmin audit loss.

**Copilot**
- Phase C keystore wording aligned with ADR §4.1 (user wallet DID-only; node/mesh separate).
- Linked #2935 in plan header.
- B3 hard-gate wording made implementation-agnostic (no false implication that types only exist in docs).

Please re-review when convenient.

PR #2947, opened 2026-07-25
fix(auth): Phase A dual-auth prep — gate shutdown, ops-pure InfraAdmin

umwelt:
## Summary
Phase A of dual-auth plan: seal holes before grant plumbing.

- Gate `POST .../node/shutdown` with `is_ops_elevated`
- InfraAdmin no longer cross-wallet role audit; self or ScopedGrant only
- Unfiltered wallet enum no longer treats InfraAdmin as fat audit role

## Stack
Depends on docs PR (`docs/dual-auth-model` → `development`). Base is docs branch.

## Test plan
- [ ] Unit/handler tests for shutdown + wallet subject read
- [ ] Confirm Council still has cross-wallet soft audit path (eligibility not yet cut)
- [ ] InfraAdmin cannot read other wallets without grant

### DIFF
--- a/zhtp/src/api/handlers/blockchain/mod.rs
+++ b/zhtp/src/api/handlers/blockchain/mod.rs
@@ -3205,16 +3205,24 @@ impl BlockchainHandler {
             None
         };
 
-        // Access control: unfiltered enumeration is restricted to elevated roles.
+        // Access control: unfiltered enumeration is audit-class (not ops-only InfraAdmin).
+        // Dual-auth plan Phase A: InfraAdmin = ops purity, no blanket wallet enum.
         if owner_filter.is_none() {
             let allowed = matches!(
                 principal.role,
                 lib_access_control::Role::System
                     | lib_access_control::Role::Node
-                    | lib_access_control::Role::InfraAdmin
                     | lib_access_control::Role::Council
                     | lib_access_control::Role::Emergency
-            );
+            )
+                || principal.grant_allows(
+                    AccessDomain::WalletGraph,
+                    AccessOperation::Enumerate,
+                    std::time::SystemTime::now()
+                        .duration_since(std::time::UNIX_EPOCH)
+                        .map(|d| d.as_secs())
+                        .unwrap_or(0),
+                );
             if !allowed {
                 return Ok(ZhtpResponse::error(
                     ZhtpStatus::Forbidden,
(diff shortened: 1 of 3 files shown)


umwelt (commented):
Phase A review — clean and correct. This is the "seal holes" PR from the dual-auth plan, and it does exactly that.

## Verified

1. **The live unauthenticated `node/shutdown` DoS is closed.** `handle_node_shutdown` went from *no auth at all* to `if !is_ops_elevated → 403` (with a logged denial). This was a remote unauthenticated node-kill on `development`; good to see it gated.
2. **InfraAdmin ops-purity is applied consistently.** InfraAdmin is removed from the cross-wallet full-read arm in **both** paths — `may_read_wallet_subject` (balance/list/statistics/transactions) and the unfiltered-enum path in `blockchain/mod.rs` — and both now fall through to `self OR grant_allows(WalletGraph, Read|Enumerate)`. Council retains the bootstrap audit path (correct per plan — eligibility isn't cut until later). The doc comments and the `ZHTP_INFRA_ADMIN_DIDS` startup-log line are updated to match. The grant branches are present but inert until Phase B wires `principal.grants`, which is the right Phase-A state.
3. **Regression guard is there** — the test asserts `!may_read_wallet_subject(&infra, hex)`, locking in the narrowing.

## Two notes

1. **This is a deliberate breaking change for operators, and it should be called out in release notes.** Any ops tooling that used an InfraAdmin DID to read or enumerate *other* users' wallets (support, debugging) now gets denied — by design (ops ≠ audit), but it will silently stop working until an audit grant exists (Phase B). The startup-log wording helps; a NEWS/runbook line would help more.
2. **The shutdown gate is the one piece I'd still lift out of the stack.** It's stacked on the docs branch (`auth/dual-auth-phase-a → docs/dual-auth-model → development`), so it can't land until the ADR PR (#2946, marked "do not merge until stack reviewed") merges. That's a live unauthenticated DoS waiting on doc sign-off. Cherry-pick just the `node/shutdown` gate straight to `development` now; keep the rest of Phase A in the stack. A catastrophic open endpoint shouldn't be blocked on ADR review.

Lineage note for anyone following along: the #2935 phase stack (#2937/#2938/#2939) is **merged**, so this builds on it sequentially — not a parallel/conflicting auth redesign. Coherent.

umwelt:
Review notes addressed:

1. **Release notes** — called out in plan Phase A item 8 (InfraAdmin loses cross-wallet audit until audit grants).
2. **Shutdown cherry-pick** — landed as standalone #2950 → `development`.
3. Shutdown deny message aligned with `is_ops_elevated` (grants included) to match #2950 review.

PR #2948, opened 2026-07-25
feat(access-control): dual-auth grant types and in-memory registry (Phase B1)

umwelt:
## Summary
Authentication half of grants + lifecycle registry (library only; fat roles still exercise).

- `GrantRecord` / offer → claim → active / revoke / lapse
- Pluggable `GrantSignatureVerifier` + `DevAccept` for tests
- Protocol class guards (NodeOperate purity)
- In-memory `GrantRegistry`
- `SecurityPrincipal::with_authenticated_grants`

## Stack
Depends on Phase A PR. Base is `auth/dual-auth-phase-a`.

## Test plan
- [x] `cargo test -p lib-access-control --lib` (32 tests green locally)
- [ ] Review types vs dual-auth ADR §4–5
- [ ] Confirm no grant private keys in DID keystore paths

### DIFF
--- a/lib-access-control/src/grant_auth.rs
+++ b/lib-access-control/src/grant_auth.rs
@@ -1,0 +1,533 @@
+//! Grant authentication half — dual-auth elevation (DID ∧ grant proof).
+//!
+//! Authorization (domain/op/expiry) lives in [`crate::grant`]. This module is
+//! structural verify + types for offer/claim/exercise. Production Dilithium
+//! (or other) signature checks plug in via [`GrantSignatureVerifier`]; the
+//! library stays free of crypto crates.
+
+use crate::grant::ScopedGrant;
+use crate::types::{AccessDomain, AccessOperation, Did};
+use serde::{Deserialize, Serialize};
+
+/// Who created the grant offer.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum IssuerKind {
+    /// Council / governance vote registered the offer.
+    Council,
+    /// Protocol rule (e.g. node admission) registered the offer.
+    Protocol,
+}
+
+/// High-level grant purpose (class catalog; domains still authoritative).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GrantClass {
+    /// Protocol: may operate a node (scope carries node_id). Not halt/export/audit.
+    NodeOperate,
+    /// Catastrophic / maintenance ops (halt, export, import, provision).
+    Ops,
+    /// Cross-identity wallet / graph audit.
+    AuditRead,
+    /// Governance vote / council admin exercise.
+    VoteGovernance,
+    /// Explicit act-as / spend delegation (optional).
+    DelegateSpend,
+    /// Unclassified / custom domain set only.
+    Custom,
+}
+
+/// Lifecycle of a registry row.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GrantStatus {
+    /// Protocol rail pre-offer (request submitted, not yet accepted).
+    Requested,
+    /// Registered vs grantee DID; not exercisable until claimed.
+    Offered,
+    /// Claimed; grant pubkey bound; dual-auth exercise allowed.
+    Active,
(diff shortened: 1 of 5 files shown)


lib-access-control/src/grant_auth.rs
```diff
@@ -0,0 +1,533 @@
...
+    }
+    let auth = record.auth.as_ref().ok_or(GrantAuthError::MissingAuth)?;
+    match auth.scheme {
+        GrantAuthScheme::DevAccept => {
+            if proof.signature != b"DEV-OK" {
+                return Err(GrantAuthError::BadSignature);
+            }
+        }
+        GrantAuthScheme::Signature => {
+            if auth.public_key.is_empty() {
```
umwelt, line 365: This arm compiles into production and accepts a fixed constant `b"DEV-OK"` as a valid grant proof — no cryptography. Since the scheme is read from the (caller-claimable) grant record, any `DevAccept` record is exercisable by anyone who can build the (non-secret) proof envelope, bypassing the grant half of dual-auth entirely. Gate both this arm and the `DevAccept` variant behind `#[cfg(any(test, feature = "dev-grants"))]` so a production build cannot honor `DEV-OK`, and reject `DevAccept` at `register_offer`/`claim` unless that feature is on. Harmless today (grants unwired) — a total bypass the moment Phase C attaches grants to sessions.

umwelt (commented):
Phase B1 review (grant types + registry + verification). The Signature path is well-built and the replay/lifecycle design is solid — but there's one latent auth bypass I'd fix before any of this gets wired into live authz (Phase C). It's not exploitable today (library-only, grants unwired), which is exactly why now is the time to close it.

## Solid

- **Replay/binding is right.** `verify_grant_proof` checks grant_id, grantee, **session_binding** (blocks cross-session replay), and a **clock-skew bound** — the Signature path is a proper second factor.
- **Lifecycle gating is correct.** `is_exercisable` denies revoked / non-Active / expired.
- **No decorative `max_uses`.** `to_scoped_grant` sets expiry + consumed-on-revoked but does *not* carry a use-count, and `is_exercisable` doesn't pretend to enforce one. That's the correct resolution of the #2939 trap (ADR §10 "no `max_uses` without a store") — deferred honestly rather than shipped non-functional.
- `RejectAllVerifier` as the safe archetype is the right instinct.

## Fix before Phase C wiring — `DevAccept` is a production-compiled bypass

`GrantAuthScheme::DevAccept` and its `DEV-OK` arm in `verify_grant_proof` are **not** `#[cfg(test)]`/feature-gated — only their *test usages* are. So the production binary contains a code path where a grant record with `scheme: DevAccept` is exercised by anyone who sends `signature = b"DEV-OK"`. grant_id / grantee_did / session_binding are all non-secret, so the **grant half of "dual-auth" — the entire second factor — is bypassable** for any DevAccept record. And `GrantRegistry::claim` takes a caller-supplied `GrantAuthDescriptor`, so the scheme isn't forced to `Signature` at claim time. The only thing preventing exploitation is "no one ever mints/claims a DevAccept grant in prod," enforced by nothing in the type system or build.

This is the dual-auth version of shipping a test verifier to production, and it undermines the exact property the ADR is built on. Not live yet (nothing wires grants into authz, and no prod code constructs DevAccept) — but it becomes a total elevation bypass the moment Phase C attaches grants. Fix now:
- `#[cfg(any(test, feature = "dev-grants"))]` on the `DevAccept` variant **and** the match arm in `verify_grant_proof`, so a production build physically cannot honor `DEV-OK`.
- Belt-and-suspenders: `register_offer` / `claim` reject `scheme == DevAccept` unless that feature is enabled.

## Note — in-memory registry must not become the ops-grant store

`GrantRegistry` is `Arc<Mutex<HashMap>>`, i.e. per-process. Fine as a B1 library scaffold, but it's the store-consistency point from the ADR review (#2946): for ops/audit grants, a per-node registry means a claim or **revocation** on g1 is invisible to g2/g3 — inconsistent enforcement of cluster-wide powers. Before Phase C wires ops grants, the sled/chain-backed (or gossip-replicated) store needs to land, and this in-memory one should be labeled test/dev scaffold so it isn't mistaken for the production store.

umwelt:
**Blocking DevAccept issue fixed** (`fix(access-control): gate DevAccept behind dev-grants feature`):

- `GrantAuthScheme::DevAccept` + verify arm only under `#[cfg(any(test, feature = "dev-grants"))]`
- Feature documented; production builds do not compile the constant second factor
- In-memory `GrantRegistry` labeled **test/dev scaffold** (not ops-grant store)

Please re-check the cfg gate before stack merge.

PR #2949, opened 2026-07-25
feat(auth): dual-auth claim/elevate API and principal grant attach (Phase B2)

umwelt:
## Summary
Phase B2 skeleton: offer → claim → dual-auth elevate → principal attach.

- Global in-memory `GrantRegistry` + elevated session store (15m TTL)
- API: `GET /me`, `POST /claim`, `POST /elevate`, `POST /offers/register`
- `extract_principal_from_request` attaches elevated `ScopedGrant`s
- Bootstrap offer registration: Council or ops-elevated only
- Protocol class purity enforced at register (NodeOperate only for protocol)

## Stack
Depends on Phase B1 (#2948). Base: `auth/dual-auth-phase-b1`.

Full stack (review before merge, do **not** merge yet):
1. #2946 docs → development
2. #2947 Phase A → docs
3. #2948 Phase B1 → Phase A
4. **this PR** Phase B2 → B1

## Out of scope (next)
- Persistent grant store (sled/chain)
- Dilithium grant signature verifier
- Client custody / grant key vault (Phase C)
- Fat-role cut-over (Phase D) — needs claimed vaulted grants first

## Test plan
- [x] `cargo test -p lib-access-control --lib` (33 tests)
- [x] `cargo test -p zhtp --lib grants::` (offer/claim/elevate flow)
- [x] `cargo test -p zhtp --lib elevated_session`
- [ ] Manual: register offer as Council → claim as grantee (DevAccept) → elevate → soft-read path with grant


umwelt (commented):
Phase B2 review — this is the wiring that makes grants *live*, so it's the security-critical PR of the stack. The plumbing is well-built (session-keyed elevation, offer gating, replay binding), and it does **not** regress anything. But it activates the DevAccept bypass from #2948 on a path that now reaches ops, and it creates a hard ordering dependency for Phase D. Both are fixable and worth nailing before merge.

## No regression today (verified)

- **Additive, not a cutover.** `is_ops_elevated` still returns true for `Role::Council | InfraAdmin` directly; the grant path is an *additional* route (`grant_allows(NodeGraph, Traverse)`). Merging removes no existing control.
- **Offer registration is gated** — `handle_register_offer` returns 403 unless `Role::Council || is_ops_elevated` (grants/mod.rs:325). So an external caller can't self-register an ops offer; the root of trust stays the fat roles.
- **Elevated-session store is sound.** Keyed on `(did, session_binding)` where `did` is the authenticated `request.requester` — you cannot attach grants to a DID you don't control. 15-minute TTL bounds the window. Good.

## Fix before merge — DevAccept is now a live, constant-proof second factor on the ops path

B2 wires elevated grants into both `is_ops_elevated` (halt/export/import/provision) and `may_read_wallet_subject`. And the **only functional scheme is DevAccept**: the elevate path passes `RejectAllVerifier` for `Signature` (grants/mod.rs — Dilithium is out of scope), so Signature grants always fail, and DevAccept is the sole working path. DevAccept's "proof" is the public constant `b"DEV-OK"` (verifier isn't even consulted for that arm), and it's not `cfg`-gated (#2948), so a **production build honors `DEV-OK`**.

Net: every grant that actually works today has a no-op second factor. "Dual-auth" collapses to DID-only for all live grants — now including ops. It's not an *external* bypass (the DID session + the ops-gated offer still stand), but it voids the ADR's core promise that a compromised DID *alone* cannot elevate: with DevAccept, controlling the grantee's DID session + sending `DEV-OK` is sufficient. There's also an amplification wrinkle — an ops-elevated grant can itself register further offers (grants/mod.rs:325 accepts `is_ops_elevated`), so one constant-proof ops grant begets more.

Gate `DevAccept` behind `#[cfg(any(test, feature = "dev-grants"))]` (variant + `verify_grant_proof` arm + `parse_auth_scheme`) so production cannot honor `DEV-OK`. This was a "fix before Phase C" note on #2948; B2 makes it *fix before this merges*.

## Hard ordering gate — Dilithium verifier before Phase D

Signature = RejectAll today. So if Phase D removes the `Role::Council | InfraAdmin` arm of `is_ops_elevated` while the only working scheme is DevAccept, ops becomes either DID-only (DevAccept, if left in) or unreachable (RejectAll, if DevAccept is gated out). Make "real Dilithium `GrantSignatureVerifier` wired into `/elevate`" an explicit *blocking predecessor* of Phase D, not just an out-of-scope bullet.

## Recurring — in-memory per-node store

Registry + elevated sessions are `OnceLock` in-memory (grant_registry_provider). Elevation on g1 is invisible to g2/g3 and revocation doesn't propagate — the cluster-consistency issue from #2946/#2948. Acceptable for a skeleton; must be chain/gossip-backed before ops grants are relied on cross-validator.

---
*(Anchoring note: the `grant_allows(NodeGraph, Traverse)` line that routes grants into `is_ops_elevated` lives at `zhtp/src/api/principal.rs:270` — added in the merged #2939, so it's not in this diff to inline-comment, but it's the exact line this PR activates.)*

umwelt:
Hard gate reminder: elevate still uses `RejectAllVerifier` for Signature scheme. **Phase B3 (real Dilithium verifier) is required before Phase D.** Do not merge cut-over without it.

umwelt:
**Blocking before-merge items from review:**

1. **DevAccept production bypass** — fixed on B1 and rebased here. Production elevate cannot honor `DEV-OK`; zhtp tests enable `dev-grants` only via `[dev-dependencies]`. Claim default scheme is `signature`.
2. **Dilithium before Phase D** — documented in plan + code comments; B3 remains hard predecessor.
3. **In-memory store** — scaffold only; ops/audit need chain/gossip before reliance (ADR decision).

Ready for re-review.

Issue #1011, opened 2026-01-29
[BFT-I][R1] Define key rotation + validator identity binding


umwelt:
## Title
[R1] Define key rotation + validator identity binding

## Category
- R1 Validator Lifecycle

## Problem
Key generation/derivation supports multiple schemes but validator identity binding and rotation rules are not explicit.

Include:
- File(s)
- Function(s)
- Code path(s)

Files/Functions/Paths:
- lib-crypto/src/keypair/derivation.rs (key derivation)
- lib-crypto/src/keypair/generation.rs (key generation)
- lib-blockchain/src/integration/consensus_integration.rs (get_validator_keypair)

## Current Behavior
Validator keys are derived/generated but rotation semantics are not enforced or versioned in consensus.

## Required Behavior
Define whether key rotation is supported; if supported, specify how identity persists across rotations and how consensus validates it.

## Safety Impact
Unspecified rotation can enable validator identity ambiguity and equivocation across keys.

## Proposed Fix
Implement explicit rotation rules and bind validator identity to a canonical key history.

## Acceptance Criteria
- [ ] Unsafe code path removed or disabled
- [ ] Invariant enforced in code
- [ ] No regression in consensus correctness
- [ ] Tests updated or added

## Estimate
- 2 days

## Planned dates
- Start: 2026-04-28
- End: 2026-04-29

PR #2967, opened 2026-08-05
feat: integrate production distributed sharding with Reed-Solomon and…

Seth-Ramsay:
This PR matures the Sovereign Network storage layer by moving from single-blob storage to a fully distributed, fault-tolerant sharding system based on Reed-Solomon erasure coding, while maintaining end-to-end Post-Quantum encryption. It also stabilizes the node runtime for Windows environments by patching a critical stack overflow during heavy cryptographic identity generation.
Linked issues
•
Refs #Phase2-Storage: Distributed Sharding Implementation.
•
Refs #1011: Windows Stack Overflow during Dilithium5 Keygen.
•
Refs #Gap-Verification: Multi-node fault tolerance proving.
Architecture compliance
Implements Distributed Storage Sharding as outlined in Phase 2 objectives. Complies with the cryptographic standards for CRYSTALS-Kyber1024 and CRYSTALS-Dilithium5. The ContentManager was also refactored to be generic over storage backends to support upcoming Sled and SQLite persistence layers.
Test plan
•
[x] Fault Tolerance Verified: Successfully reconstructed original documents after simulating a 60% data loss (lost 3 out of 5 shards) in multi_node_sharding.rs.
•
[x] PQ-Security Verified: Confirmed successful decryption of reconstructed shards using Post-Quantum private keys in production_sharding_verify.rs.
•
[x] Windows Stability: Verified that the node no longer crashes during identity generation on Windows after implementing spawn_blocking wrappers for Dilithium5 operations.
•
[x] Manual verification: Run cargo run --example production_sharding_verify -p lib-storage to see the full sharding/reconstruction lifecycle in action.
Risk and reversibility
Medium Impact: This changes the fundamental storage format for new content. While it significantly increases data durability, content stored before this PR (single-blob format) will not be compatible with the new sharded retrieval logic. A coordinated network data-wipe or migration script is recommended for testnet participants.

I documented it all in docs/sharding_verification.md file contains the full technical breakdown and mathematical proof of the recovery logic.


PeterGRutherford:
## What is correct
1. The Reed-Solomon encode/decode wiring is mathematically sound. Shard indices map correctly to the combined `data_shards + parity_shards` layout, the "need at least `data_shards` shards" threshold is correct, and `original_size = processed_size` is consistent with what was encoded.
2. Shard keys are deterministic and unique: `blake3("shard:" + content_hash + ":" + index)`.
3. Persisting content metadata for the Sled backend (instead of in-memory only) is a reasonable direction.

## What is wrong / incomplete
1. **Not actually distributed across nodes.** All shards are written to a *single* `dht_storage` instance under different keys. There is no routing/replication to separate nodes. The "distributed across Kademlia nodes" claim in (docs/sharding_verification.md) is only demonstrated by (lib-storage/examples/multi_node_sharding.rs), which is a hand-written simulation that manually places shards in two separate `UnifiedStorageSystem` instances — it does **not** exercise the production `upload_content`/`download_content` path.

2. **Erasure config is hardcoded and ignores configuration.** `ContentManager::new` hardcodes `ErasureCoding::new(4, 2)` and ignores `UnifiedStorageConfig.erasure_config`. The examples set `erasure_config` to `2+3`/`4+2`, but that value has no effect on content upload. This is also redundant with the already-existing config-driven `store_with_erasure_coding` path in (lib-storage/src/lib.rs).

3. **Backward-compatibility break.** `ContentMetadata` is serialized with `bincode`, and four new fields are appended with no `#[serde(default)]`/versioning. Old persisted metadata will fail to deserialize. More importantly, `download_content` now reads only `metadata.shard_hashes` and errors if empty, so content uploaded under the old single-blob path (stored under `content_hash`) is no longer retrievable — there is no fallback to the legacy read path.

4. **Web4 content serving is broken end-to-end.** In (zhtp/src/api/handlers/web4/domains.rs) the branch shards each item into `self.storage`, then passes an **empty** `initial_content` to `register_domain_with_content`. In (lib-network/src/web4/domain_registry.rs), `register_domain` only builds `content_mappings` by iterating `request.initial_content`, so the domain record ends up with no content mappings. Serving ((zhtp/src/api/handlers/web4/mod.rs) → `Web4ContentService.serve`) reads from the domain registry's own content store, not from `lib_storage`. Net effect: Web4 content uploaded this way is sharded but not servable.

5. **Unrelated/regressive changes bundled in the same branch:**
   - Committed IDE noise: `.idea/` (including a ~2.2k-line `deviceStreaming.xml`).
   - Windows `spawn_blocking` keygen patches are legitimate but unrelated to sharding.

6. **Minor regressions:** download no longer updates `last_accessed`/`access_count`, and the Web4/domain loops hold a write lock across async `upload_content`, creating contention.

## Recommendation
Do not merge as-is. The Reed-Solomon core can be salvaged, but "distributed sharding" needs: (a) real multi-node distribution/replication in the production `upload_content`/`download_content` path, (b) config-driven shard counts instead of hardcoded `4+2`, (c) metadata versioning plus a legacy single-blob fallback, and (d) wiring every read path (especially Web4 `serve_content`) to reconstruct through `download_content`. The `wallets_for_owner` revert and unrelated `.idea`/test deletions should be removed from the branch.

Issue #2970, opened 2026-08-13
wallets_for_owner() missed during #2639 sled migration — in-memory-only, needs sled-first + overlay pattern


PeterGRutherford:
## Problem

(lib-blockchain/src/blockchain/wallets.rs:741) reads exclusively from the in-memory `wallet_registry` HashMap. It was missed during the [#2639](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2639) sled migration.

Current code at line 741-750:
```rust
pub fn wallets_for_owner(&self, owner_identity_id: &Hash)
    -> Vec<(String, WalletTransactionData)>
{
    self.wallet_registry
        .iter()
        .filter(|(_, wallet)| wallet.owner_identity_id.as_ref() == Some(owner_identity_id))
        .map(|(wallet_id, wallet)| (wallet_id.clone(), wallet.clone()))
        .collect()
}
```

The doc comment explicitly says: *"in the in-memory shadow only (O(N) over `wallet_registry`, not sled)"*

## Impact

- Called by (zhtp/src/api/handlers/identity/backup_recovery.rs:632) to find wallets owned by an identity
- After restart, the in-memory `wallet_registry` is empty (only repopulated by the legacy `process_wallet_transactions` bridge)
- If [GENESIS-7 (#2735)](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2735) removes the bridge, this method silently returns empty

## Fix

Apply the same sled-first + in-memory-overlay pattern as (lib-blockchain/src/blockchain/wallets.rs:930):

```rust
pub fn wallets_for_owner(&self, owner_identity_id: &Hash)
    -> Vec<(String, WalletTransactionData)>
{
    let mut out = HashMap::new();
    // 1. sled first
    if let Some(store) = self.get_store() {
        if let Ok(iter) = store.iter_wallet_projections() {
            for (wallet_id, record) in iter {
                if record.wallet_data.owner_identity_id.as_ref() == Some(owner_identity_id) {
                    out.insert(hex::encode(wallet_id), record.wallet_data);
                }
            }
        }
    }
    // 2. in-memory overlay (wins on conflict)
    for (wallet_id_hex, data) in &self.wallet_registry {
        if data.owner_identity_id.as_ref() == Some(owner_identity_id) {
            out.insert(wallet_id_hex.clone(), data.clone());
        }
    }
    out.into_iter().collect()
}
```

## Dependencies

[...]

Issue #2735, opened 2026-06-27
[GENESIS-7] Delete legacy process_*_transactions write paths


umwelt:
# Context

Part of epic #2727. **Supersedes #2641.**

Depends on GENESIS-1 (#2729), GENESIS-2 (#2730), GENESIS-6 (#2734). Do **not** land until genesis sled projection is proven — deleting seed-sled without projection makes replay worse.

# Problem

Legacy `process_*_transactions` in `process_and_commit_block` re-applies every block to in-memory HashMaps. Executor already wrote sled. Dual-write + seed-sled reconciliation is the root fragility.

# Scope

Delete legacy replay writers per #2641 / #2640 completion:

1. Remove `process_token_transactions` and siblings from hot path
2. Remove seed-sled / back-sync dance in `process_and_commit_block`
3. Replay rebuilds state from sled via executor + lazy cache refresh only
4. Update `replay_from_store` / `load_from_store` paths

# Acceptance criteria

- [ ] No `process_*_transactions` calls in block commit hot path
- [ ] No seed-sled / back-sync reconciliation in `process_and_commit_block`
- [ ] GENESIS-2 replay test passes after deletion
- [ ] `cargo test -p lib-blockchain` and g4 soak pass

# Dependencies

- GENESIS-1 (#2729)
- GENESIS-2 (#2730)
- GENESIS-6 (#2734)
- #2640 (closed)

# Supersedes

- #2641

PR #2974, opened 2026-08-14
fix(#2971): get_wallets_for_owner now reads sled-first with in-memory overlay

### LINKED ISSUE #2971: get_wallets_for_owner() missed during #2639 sled migration — in-memory-only, breaks voting power after restart
## Problem

(lib-blockchain/src/blockchain/wallets.rs:990) reads exclusively from the in-memory `wallet_registry` HashMap. It was missed during the [#2639](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2639) sled migration.

Current code at line 990-998:
```rust
pub fn get_wallets_for_owner(&self, owner_identity_id: &Hash)
    -> Vec<&WalletTransactionData>
{
    self.wallet_registry
        .values()
        .filter(|wallet| wallet.owner_identity_id.as_ref() == Some(owner_identity_id))
        .collect()
}
```

## Impact

Called by (lib-blockchain/src/blockchain.rs:4322) to sum SOV balances across all wallets owned by an identity for DAO governance voting power. After restart, the in-memory `wallet_registry` is empty, so voting power would be calculated as 0.

There is already a test proving the intent: [`calculate_user_voting_power_reads_sled_not_in_memory`](lib-blockchain/src/blockchain/tests/facade_tests.rs:461).

## Fix

Apply the same sled-first + in-memory-overlay pattern as (lib-blockchain/src/blockchain/wallets.rs:930). Note: the return type is `Vec<&WalletTransactionData>` — since sled data isn't borrowed from self, this may need to change to return owned `Vec<WalletTransactionData>` instead.

## Dependencies

- **Blocks** [GENESIS-7 (#2735)](https://github.com/SOVEREIGN-NET/The-Sovereign-Network/issues/2735)
- **Related** to #2970 (sibling method `wallets_for_owner()` has the same gap)

PeterGRutherford:
## Summary

`get_wallets_for_owner()` was reading exclusively from the in-memory `wallet_registry` HashMap, which is empty after restart. Changed to sled-first + in-memory-overlay pattern so DAO governance voting power correctly reads from durable sled storage.

## Linked issues

Closes #2971. Refs #2639, #2735, #2970.

## Architecture compliance

Implements the sled-first + in-memory-overlay pattern established in [`wallet_registry_snapshot()`](lib-blockchain/src/blockchain/wallets.rs:930) and [`wallet_transaction_data()`](lib-blockchain/src/blockchain/wallets.rs:912), completing the migration missed during #2639.

## Test plan

- [x] Existing test `calculate_user_voting_power_reads_sled_not_in_memory` passes — proves voting power reads from sled, not the empty in-memory map after restart.
- [x] `cargo check` — compiles cleanly.

## Risk and reversibility

Low risk. Return type changed from `Vec<&WalletTransactionData>` to `Vec<WalletTransactionData>` (owned), but both callers in [`blockchain.rs`](lib-blockchain/src/blockchain.rs:4322) use `.iter()` which works identically on either. No coordinated network restart required.

### DIFF
--- a/lib-blockchain/src/blockchain/wallets.rs
+++ b/lib-blockchain/src/blockchain/wallets.rs
@@ -987,14 +987,43 @@ impl Blockchain {
         })
     }
 
+    /// Wallets owned by an identity, sled-first with in-memory overlay (#2639, #2971).
+    ///
+    /// Returns owned `WalletTransactionData` from durable wallet projections when
+    /// the store is attached, then overlays the in-memory shadow (pending / same-block
+    /// registrations win). After restart the in-memory shadow is empty, so this
+    /// correctly reads from sled — fixing the gap where voting power was 0 after
+    /// restart (#2971).
     pub fn get_wallets_for_owner(
         &self,
         owner_identity_id: &Hash,
-    ) -> Vec<&crate::transaction::WalletTransactionData> {
-        self.wallet_registry
-            .values()
-            .filter(|wallet| wallet.owner_identity_id.as_ref() == Some(owner_identity_id))
-            .collect()
+    ) -> Vec<crate::transaction::WalletTransactionData> {
+        let mut out = HashMap::new();
+        if let Some(store) = self.get_store() {
+            match store.iter_wallet_projections() {
+                Ok(iter) => {
+                    for (wallet_id, record) in iter {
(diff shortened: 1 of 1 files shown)


PeterGRutherford:
sled-first work on get_wallets_for_owner() would be redundant and creates duplication. I'll likely depreciate the function instead.

